PR #1, opened 2018-11-14
Initial Prototype w/ PassthroughController

DwayneJengSage:
Initial prototype w/ PassthroughController and an EchoController for basic health testing and as an example controller.

Set up FindBugs and Jacoco test coverage.

Ran a subset of Integ Tests against my local env. Things seem to work except for HttpTest which fails on the CORS stuff.

Also created an Elastic Beanstalk Env and did some basic testing: http://bridgeserver2-dwaynejeng.us-east-1.elasticbeanstalk.com/


src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
```diff
@@ -0,0 +1,144 @@
...
+                pfRequest = Request.Post(fullUrl);
+                break;
+            case "DELETE":
+                pfRequest = Request.Delete(fullUrl);
+                break;
+            default:
+                String errorMessage = "Method " + request.getMethod() + " not supported";
+                LOG.warn(errorMessage);
+                return ResponseEntity.badRequest().build();
+        }
```
alxdarksage, line 81: We need to pass through the OPTIONS HTTP method, I believe. If you were to access this server through the Bridge study manager, I believe you'd see errors around the failure to process OPTIONS requests. I should try and get this running locally and work on the CORS stuff, I'll do that after you merge.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
DwayneJengSage: Spring MVC has special handling of the OPTIONS method, so this handler never sees that at all.

PR #6, opened 2018-12-20
HTTPS Redirection

DwayneJengSage:
Use an interceptor to redirect from HTTP to HTTPS. This uses the same logic as our BridgePF HTTP handler.

Testing done:
* mvn verify (unit tests, FindBugs, Jacoco test coverage)
* Manually built and uploaded to dev. Verified that an http call gets a 301 Moved Permanently with the correct URL.
* Manually tested locally, to make sure it works on local server.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/config/BridgeWebMvcConfig.java
+++ b/src/main/java/org/sagebionetworks/bridge/config/BridgeWebMvcConfig.java
@@ -1,0 +1,23 @@
+package org.sagebionetworks.bridge.config;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/config/BridgeWebMvcConfig.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @Autowired
+    public final void setHttpsForwardingInterceptor(HttpsForwardingInterceptor httpsForwardingInterceptor) {
+        this.httpsForwardingInterceptor = httpsForwardingInterceptor;
+    }
+
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        registry.addInterceptor(httpsForwardingInterceptor);
+    }
```
alxdarksage, line 22: Boy this is nice

alxdarksage (approved):

PR #9, opened 2019-05-02
Adding controllers to BridgeServer2

alxdarksage:


src/main/resources/bridge-server.conf
```diff
@@ -0,0 +1,155 @@
+channel.throttle.max.requests = 1
+channel.throttle.timeout.seconds = 300
```
DwayneJengSage: Why are we duplicating this file? It's error-prone to have to make the same changes in both projects. We should have both projects reading from the conf file.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -0,0 +1,71 @@
...
+        this.cacheProvider = cacheProvider;
+    }
+
+    @Override
+    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
+            throws IOException, ServletException {
+        final Metrics metrics = initMetrics((HttpServletRequest)req);
+        
+        CacheKey metricsKey = CacheKey.metricsKey(metrics.getCacheKey());
+        cacheProvider.setObject(metricsKey, metrics, BridgeConstants.METRICS_EXPIRE_SECONDS);
```
DwayneJengSage: Does it make sense to use Redis for this? A request is handled by at most a single host (and oftentimes a single thread). Would it make more sense to store metrics in the RequestContext?

pom.xml
```diff
@@ -57,7 +62,14 @@
...
         </dependency>
-
+        <!-- Apparently needed for PowerMock tests to run. I needed to add this for the tests to run. See: 
+        	https://stackoverflow.com/questions/32854688/powermock-java-lang-illegalstateexception-failed-to-transform-class
+        -->
+		<dependency>
+        	<groupId>org.javassist</groupId>
+        	<artifactId>javassist</artifactId>
+        	<version>3.18.2-GA</version>
+    	</dependency>
```
DwayneJengSage: Looks like the formatting here mixes tabs and spaces.

Also, should this be declared as a test dependency? (And moved to be with the other test dependencies in this file?)

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
```diff
@@ -38,7 +142,621 @@ public Config bridgeConfig() {
...
+
+    @Bean(name = "annotationBasedTableCreator")
+    public AnnotationBasedTableCreator annotationBasedTableCreator(DynamoNamingHelper dynamoNamingHelper) {
+        return new AnnotationBasedTableCreator(dynamoNamingHelper);
+    }
+
+    @Bean(name = "healthCodeEncryptor")
+    public BridgeEncryptor healthCodeEncryptor() {
+        return new BridgeEncryptor(new AesGcmEncryptor(bridgeConfig().getHealthCodeKey()));
+    }
```
DwayneJengSage: Minor: Pretty sure healthCodeEncryptor is no longer used.

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
```diff
@@ -38,7 +142,621 @@ public Config bridgeConfig() {
...
+        props.put("hibernate.connection.password", config.get("hibernate.connection.password"));
+        props.put("hibernate.connection.username", config.get("hibernate.connection.username"));
+
+        
+        String url = config.get("hibernate.connection.url");
+        
+        // Append SSL props to URL
+        boolean useSsl = Boolean.valueOf(config.get("hibernate.connection.useSSL"));
+        url += (url.contains("?")) ? "&" : "?";
+        url += "requireSSL="+useSsl+"&useSSL="+useSsl+"&verifyServerCertificate="+useSsl;
```
DwayneJengSage: As per IM conversation, we should just hardcode the timezone to always be UTC.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,60 @@
...
+
+@CrossOrigin
+@RestController("activityEventController")
+@RequestMapping("/v1/activityevents")
+public class ActivityEventController extends BaseController {
+
+    private ActivityEventService activityEventService;
+
+    @Autowired
+    public void setActivityEventService(ActivityEventService activityEventService) {
```
DwayneJengSage: All spring setters should be marked final.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ActivityEventController.java
```diff
@@ -0,0 +1,60 @@
...
+    @GetMapping
+    public String getSelfActivityEvents() throws Exception {
+        UserSession session = getAuthenticatedAndConsentedSession();
+
+        List<ActivityEvent> activityEvents = activityEventService.getActivityEventList(session.getHealthCode());
+        
+        // I do not like the fact we are serializing in the controller, but that's the only way to access
+        // the ObjectWriter and that's currently how we suppress healthCode.
+        ResourceList<ActivityEvent> list = new ResourceList<>(activityEvents);
+        return ACTIVITY_EVENT_WRITER.writeValueAsString(list);
```
DwayneJengSage, line 58: Is this being returned with the correct content type?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CacheAdminController.java
```diff
@@ -0,0 +1,45 @@
...
+
+@CrossOrigin
+@RestController("cacheAdminController")
+@RequestMapping("/v3/cache")
+public class CacheAdminController extends BaseController {
+    
+    private CacheAdminService cacheAdminService;
+
+    @Autowired
+    public void setCacheAdminService(CacheAdminService cacheService) {
```
DwayneJengSage: All spring setters should be marked final.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
```diff
@@ -82,46 +84,43 @@ public void setConfig(Config config) {
             case "DELETE":
                 pfRequest = Request.Delete(fullUrl);
                 break;
+            case "HEAD":
+                pfRequest = Request.Head(fullUrl);
+                break;
```
DwayneJengSage, line 94: We should add a test for HEAD to maintain code coverage.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
```diff
@@ -82,46 +84,43 @@ public void setConfig(Config config) {
...
-            }
-
-            if (headerName.equalsIgnoreCase(HEADER_REQUEST_ID)) {
-                requestId = headerValue;
-            }
+        // Only carry over the headers that Bridge server directly examines. Let Spring do the Cross-Origin headers.
+        String requestId = request.getHeader(X_REQUEST_ID_HEADER);
+        String userAgent = request.getHeader(USER_AGENT);
+        String acceptLanguage = request.getHeader(ACCEPT_LANGUAGE);
+        String sessionToken = request.getHeader(SESSION_TOKEN_HEADER);
```
DwayneJengSage, line 106: Content-Type?

src/test/java/org/sagebionetworks/bridge/spring/controllers/PassthroughControllerTest.java
```diff
@@ -228,26 +226,16 @@ public void unrecognizedStatusCode() throws Exception {
...
-                .put(PassthroughController.HEADER_REQUEST_ID, OTHER_REQUEST_ID)
-                .build();
-
         HttpServletRequest mockRequest = mock(HttpServletRequest.class);
         when(mockRequest.getContentType()).thenReturn(MIME_TYPE_TEXT_PLAIN);
-        when(mockRequest.getHeaderNames()).thenReturn(new Vector<>(requestHeaderMap.keySet()).elements());
+        when(mockRequest.getHeader("Dummy-Header")).thenReturn("dummy header value");
+        when(mockRequest.getHeader("Content-Length")).thenReturn("10");
+        when(mockRequest.getHeader(X_FORWARDED_FOR_HEADER)).thenReturn(OTHER_IP_ADDRESS);
+        when(mockRequest.getHeader(X_REQUEST_ID_HEADER)).thenReturn(OTHER_REQUEST_ID);
```
DwayneJengSage, line 270: We should add user agent, accept language, session token, and content-type headers, to make sure we're passing those through correctly.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
```diff
@@ -143,22 +142,28 @@ public void setConfig(Config config) {
...
 
-        // Response headers.
+        // Response headers. We only pass back Bridge specific headers so there's no interference with Spring filters.
         HttpHeaders springHeaders = new HttpHeaders();
-        for (Header header : pfResponse.getAllHeaders()) {
-            springHeaders.add(header.getName(), header.getValue());
+        copyHeader(springHeaders, pfResponse, SESSION_TOKEN_HEADER);
+        copyHeader(springHeaders, pfResponse, BRIDGE_API_STATUS_HEADER);
+        copyHeader(springHeaders, pfResponse, HttpHeaders.CONTENT_LENGTH);
+        copyHeader(springHeaders, pfResponse, HttpHeaders.CONTENT_TYPE);
```
DwayneJengSage: Set-Cookie?

src/test/java/org/sagebionetworks/bridge/spring/controllers/PassthroughControllerTest.java
```diff
@@ -275,25 +262,32 @@ public void requestWithHeaders() throws Exception {
...
+        mockHttpResponseForRequest(mockPfRequest, 200, responseHeaderMap, expectedResponseBody);
 
         mockStatic(Request.class);
         when(Request.Get(anyString())).thenReturn(mockPfRequest);
 
+        // Here's what we should copy back from the Bridge response. We do not include every
+        // header.
+        Map<String, String> expectedResponseHeaderMap = ImmutableMap.<String, String>builder()
+                .put("Content-Type", MIME_TYPE_TEXT_PLAIN)
+                .put("Content-Length", "22")
```
DwayneJengSage: We should add Bridge API Status, Session Token, and Set-Cookie.

DwayneJengSage (changes requested):

src/main/resources/bridge-server.conf
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
alxdarksage: Yes. Switched to the use of RequestContext.

pom.xml
alxdarksage: Yes, moved. Also fixed Eclipse to use 4 spaces instead of a tab in XML files.

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
alxdarksage: Removed (I can't find any use outside of HealthCodeUtil in BridgeServerLogic and I'm not sure that's still useful).

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
alxdarksage: Switched to ?serverTimezone=UTC...

src/main/java/org/sagebionetworks/bridge/spring/controllers/ActivityEventController.java
alxdarksage: Updated (they also don't need to be public)

src/main/java/org/sagebionetworks/bridge/spring/controllers/ActivityEventController.java
alxdarksage: Yes, the @RestController annotation includes the @ResponseBody annotation. The default of that annotation is to return JSON. It returns Content-Type: text/plain;charset=UTF-8 and doesn't escape the string, even if you don't send an Accept header indicating you'll take JSON.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CacheAdminController.java
alxdarksage: Updated.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
alxdarksage: Not from the request, but the response needs to return it for our html based pages. I've tested this works. If we ever varied content based on the request, we'd need to pass through the Accept header.

src/test/java/org/sagebionetworks/bridge/spring/controllers/PassthroughControllerTest.java
alxdarksage: Added UA, Accept-Language and Bridge-Session. Added headers to the test of passing back headers from BridgePF.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PassthroughController.java
alxdarksage: Yes, good one.  Added and added to test

alxdarksage: Should be passed back through the passthrough controller as the "Cookie" header, so added that as well.

src/test/java/org/sagebionetworks/bridge/spring/controllers/PassthroughControllerTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #17, opened 2019-05-09
03 Consent and participant controllers.

alxdarksage:
This PR is larger than the others because we want to move over certain endpoints at the same time to avoid matching errors where a path like /v3/foo/recent matches /v3/foo/:id, which is not the desired behavior. To avoid this, ParticipantReportController is included along with ParticipantController and StudyConsentController is included along with ConsentController (originally this was just ParticipantController and ConsentController).


src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportControllerTest.java
```diff
@@ -0,0 +1,566 @@
...
+        assertEquals(reportData.getData().get(2).asText(), "C");
+    }
+    
+    @Test
+    public void saveParticipantReportForWorkerRequiresHealthCode() throws Exception {
+        String json = TestUtils.createJson("{'date':'2015-02-12','data':['A','B','C']}");
+        mockRequestBody(mockRequest, json);
+        try {
+            controller.saveParticipantReportForWorker(REPORT_ID);    
+        } catch(BadRequestException e) {
```
DwayneJengSage, line 394: Minor: We should add a fail() to ensure that the exception is being thrown.

DwayneJengSage (approved):
One minor comment about a pre-existing issue. Other than that, looks good to merge.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportControllerTest.java
alxdarksage: Added.

PR #18, opened 2019-05-10
04 CompountActivityDefinitionController, EmailController, ExportController

alxdarksage:
- CompountActivityDefinitionController
- EmailController
- ExportController


src/main/java/org/sagebionetworks/bridge/spring/controllers/EmailController.java
```diff
@@ -0,0 +1,79 @@
...
+    @RequestMapping(method = {GET, POST}, path="/v3/users/self/unsubscribeEmail", produces="text/plain")
+    public String unsubscribeFromEmail() throws Exception {
+        // We catch and return 200s because MailChimp makes a validation call when configuring the web hook, and if it fails,
+        // MailChimp won't persist the configuration. We could also detect the validation call because it has a different
+        // User-Agent than the real callbacks, "MailChimp.com WebHook Validator" versus "MailChimp.com", and always return 
+        // 200 for that validation call.
+        try {
+            // The servlet getParameter() method will extract the value from a query string or from 
+            // POSTed form data, making this much simpler than in Play.
+            String token = request().getParameter("token");
```
DwayneJengSage, line 41: Under the old code, we would check both the query params and the post body. Does getParameter() check both query params and post body? If not, is this an intended change? Do we know that it works with MailChimp?

src/test/java/org/sagebionetworks/bridge/spring/controllers/EmailControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private void mockContext(String... values) {
+        Map<String, String[]> paramMap = Maps.newHashMap();
+        for (int i = 0; i <= values.length - 2; i += 2) {
+            paramMap.put(values[i], new String[] { values[i + 1] });
+        }
+        for (Map.Entry<String, String[]> param : paramMap.entrySet()) {
+            when(mockRequest.getParameter(param.getKey())).thenReturn(param.getValue()[0]);
+        }
+    }
```
DwayneJengSage, line 88: In the old tests, we had separate tests for getting from query param or getting from the post body. Is this abstracted away in request.getParameter(). If not, do we need to separate these out again? What about the mixed() test?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/EmailController.java
alxdarksage: Yes, it definitely does (I have tested this manually, and verified it in the documentation). If MailChimp sends the same variables as when we wrote this, then it'll work (the formats are very well specified for query strings and application/x-www-form-urlencoded POST data).

src/test/java/org/sagebionetworks/bridge/spring/controllers/EmailControllerTest.java
alxdarksage: See above. The getParameter() method will check both places. Documented here: https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getParameter(java.lang.String). We expect one parameter but if MailChimp sends two, we wouldn't be able to pick between them anyway, so it's fine that getParameter() takes the first one. This is how it worked in the old implementation as well.

DwayneJengSage (approved):

PR #19, opened 2019-05-10
05 ExternalIdControllerV4, FPHSController, HealthDataController

alxdarksage:


src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataController.java
```diff
@@ -0,0 +1,139 @@
...
+        if (metrics != null) {
+            metrics.setRecordId(savedRecord.getId());
+        }
+
+        // Return the record produced by this submission. Filter out Health Code, of course.
+        return HealthDataRecord.PUBLIC_RECORD_WRITER.writeValueAsString(savedRecord);
+    }
+
+    // Not mapped in the BridgePF routes file.
+    /* public StatusMessage updateRecordsStatus() {
```
DwayneJengSage: This is used by /v3/recordexportstatuses and /v3/recordExportStatuses (Note there was a capitalization snafu when this was implemented. I'm not sure which one is actually in use right now.) This is used by the Bridge Exporter.

Similarly, we should also add the unit test back.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataController.java
alxdarksage: Thanks for finding this... added back along with test.

DwayneJengSage (approved):

PR #21, opened 2019-05-13
07 OAuthController, ScheduleController, ScheduledActivityController

alxdarksage:


src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduleController.java
```diff
@@ -0,0 +1,96 @@
...
+@CrossOrigin
+@RestController
+public class ScheduleController extends BaseController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ScheduleController.class);
+    
+    private SchedulePlanService schedulePlanService;
+    
+    @Autowired
+    public void setSchedulePlanService(SchedulePlanService schedulePlanService) {
```
DwayneJengSage: Minor: This should be set final.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityController.java
```diff
@@ -0,0 +1,264 @@
...
+    
+    static final TypeReference<ArrayList<ScheduledActivity>> SCHEDULED_ACTIVITY_TYPE_REF = new TypeReference<ArrayList<ScheduledActivity>>() {};
+    static final StatusMessage UPDATED_MSG = new StatusMessage("Activities updated.");
+    static final String MISSING_TIMESTAMP_ERROR = "startsOn and endsOn are both required and must be ISO 8601 timestamps.";
+    static final String AMBIGUOUS_TIMEZONE_ERROR = "startsOn and endsOn must be in the same time zone.";
+
+    private ScheduledActivityService scheduledActivityService;
+
+    @Autowired
+    void setScheduledActivityService(ScheduledActivityService scheduledActivityService) {
```
DwayneJengSage: Minor: final

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
```diff
@@ -0,0 +1,568 @@
...
+    @Test
+    public void timeZoneCapturedFirstTime() throws Exception {
+        mockEditAccount(mockAccountDao, mockAccount);
+        
+        DateTimeZone MSK = DateTimeZone.forOffsetHours(3);
+        controller.getScheduledActivities(null, "+03:00", "3", "5");
+        
+        verify(mockAccount).setTimeZone(MSK);
+        verify(sessionUpdateService).updateTimeZone(session, MSK);
+    }
```
DwayneJengSage, line 222: Add the check for scheduledActivityService.getScheduledActivities() to main parity with old tests.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
```diff
@@ -0,0 +1,568 @@
...
+        assertEquals(critContext.getStudyIdentifier().getIdentifier(), TEST_STUDY_IDENTIFIER);
+        assertEquals(critContext.getClientInfo(), CLIENT_INFO);
+        
+        verify(mockCacheProvider).updateRequestInfo(requestInfoCaptor.capture());
+        RequestInfo requestInfo = requestInfoCaptor.getValue();
+        assertEquals(requestInfo.getUserId(), "id");
+        assertEquals(requestInfo.getLanguages(), LANGUAGES);
+        assertEquals(requestInfo.getUserDataGroups(), USER_DATA_GROUPS);
+        assertEquals(requestInfo.getUserSubstudyIds(), USER_SUBSTUDY_IDS);
+        assertEquals(requestInfo.getStudyIdentifier(), TEST_STUDY);
```
DwayneJengSage, line 280: There are some attributes of requestInfo that aren't part of sessionUpdateService.updateTimeZone() that should be tested here.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
```diff
@@ -0,0 +1,568 @@
...
+        assertEquals(requestInfo.getUserId(), "id");
+        assertEquals(requestInfo.getClientInfo(), CLIENT_INFO);
+        assertEquals(requestInfo.getUserAgent(), USER_AGENT);
+        assertEquals(requestInfo.getLanguages(), LANGUAGES);
+        assertEquals(requestInfo.getUserDataGroups(), USER_DATA_GROUPS);
+        assertEquals(requestInfo.getUserSubstudyIds(), USER_SUBSTUDY_IDS);
+        assertTrue(requestInfo.getActivitiesAccessedOn().isAfter(startsOn));
+        assertNull(requestInfo.getSignedInOn());
+        assertEquals(requestInfo.getStudyIdentifier(), TEST_STUDY);
+        verify(controller).persistTimeZone(session, zone);
```
DwayneJengSage: Similarly, we still need to check requestInfo.getTimeZone().

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
alxdarksage: Added these and a test that the mock account is having its time zone updated.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
alxdarksage: RequestInfo.getTimeZone() doesn't change from its default value which is UTC. Although the user has no initial time zone set and this call is setting it, without the SessionUpdateService to actually update the session before creating RequestInfo, it's not updated. Added a verify that the mock account is being updated with the time zone.

Not sure how else to handle this since it was incorrect to use the SessionUpdateService implementation instead of mocking it. the call to the service is being made correctly and that's tested separately.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityController.java
alxdarksage: Changed

DwayneJengSage: Mark this as final.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
```diff
@@ -0,0 +1,584 @@
...
+        assertEquals(critContext.getClientInfo(), CLIENT_INFO);
+        
+        verify(mockCacheProvider).updateRequestInfo(requestInfoCaptor.capture());
+        RequestInfo requestInfo = requestInfoCaptor.getValue();
+        assertEquals(requestInfo.getUserId(), "id");
+        assertEquals(requestInfo.getLanguages(), LANGUAGES);
+        assertEquals(requestInfo.getUserDataGroups(), USER_DATA_GROUPS);
+        assertEquals(requestInfo.getUserSubstudyIds(), USER_SUBSTUDY_IDS);
+        assertEquals(requestInfo.getStudyIdentifier(), TEST_STUDY);
+        assertNotNull(requestInfo.getActivitiesAccessedOn());
```
DwayneJengSage: Need to check requestInfo.getActivitiesAccessedOn().getZone()

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityController.java
alxdarksage: Sorry I think I changed BridgePF by accident... changed

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
alxdarksage: Different from line 220?

src/test/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityControllerTest.java
alxdarksage: Added. This exposed what I think is a bug, so I changed the controller to get the desired behavior (it comes down to: what is the time zone in RequestInfo, is it the user's initial time zone, or is it the time zone of the most recent request? It looks like here, we're expecting it to be the latter, so I've changed the code so the test passes. This is now different than BridgePF.

DwayneJengSage (approved):

PR #22, opened 2019-05-14
08 SchedulePlanController, SharedModuleMetadataController, SharedModuleController

alxdarksage:
Discovered that mostrecent and includeDeleted differ in case (probably should have picked one approach). They are supported exactly as they are supported in BridgePF (with different casing).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/SchedulePlanController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/SchedulePlanController.java
@@ -1,0 +1,113 @@
+package org.sagebionetworks.bridge.spring.controllers;
+
+import static org.sagebionetworks.bridge.Roles.ADMIN;
+import static org.sagebionetworks.bridge.Roles.DEVELOPER;
+import static org.sagebionetworks.bridge.Roles.RESEARCHER;
+import static org.sagebionetworks.bridge.Roles.WORKER;
+
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/SchedulePlanController.java
```diff
@@ -0,0 +1,113 @@
...
+
+@CrossOrigin
+@RestController
+public class SchedulePlanController extends BaseController {
+
+    static final StatusMessage DELETE_MSG = new StatusMessage("Schedule plan deleted.");
+    private SchedulePlanService schedulePlanService;
+    
+    @Autowired
+    void setSchedulePlanService(SchedulePlanService schedulePlanService) {
```
DwayneJengSage, line 42: Minor: final

src/main/java/org/sagebionetworks/bridge/spring/controllers/SharedModuleController.java
```diff
@@ -0,0 +1,41 @@
...
+
+@CrossOrigin
+@RestController
+public class SharedModuleController extends BaseController {
+    
+    private SharedModuleService moduleService;
+
+    /** Shared Module Service, configured by Spring. */
+    @Autowired
+    void setModuleService(SharedModuleService moduleService) {
```
DwayneJengSage: Minor: final

DwayneJengSage (approved):
Looks good. Minor comments about setting Spring setters as final.

src/main/java/org/sagebionetworks/bridge/spring/controllers/SharedModuleController.java
alxdarksage: Updated

PR #39, opened 2019-05-29
delete ErrorController and PassthroughController

DwayneJengSage:
Deleting PassthroughController since all controllers are now migrated and we don't need to forward traffic to BridgePF.

Deleting ErrorController, because for some reason, it interferes with invalid URLs.

Adding a DefaultController to handle unrecognized URLs. The response is
```
{
    "statusCode": 404,
    "message": "Endpoint not found.",
    "type": "EndpointNotFoundException"
}
```

which is the same as BridgePF. Also verified that it's logged correctly, without a stack trace, but with an entry in MetricsFilter.


alxdarksage (approved):

alxdarksage (approved):
Great solution, thanks

PR #42, opened 2019-06-06
Merging in BridgeServerLogic

alxdarksage:
I believe the exclusions have been moved to the right package given the comment in the original pom.xml file.


pom.xml
```diff
@@ -114,6 +84,149 @@
             <version>4.8.0</version>
             <scope>runtime</scope>
         </dependency>
```
DwayneJengSage: Minor: Can you move these 3 dependencies with the other runtime dependencies to keep everything in the same place?

pom.xml
```diff
@@ -161,9 +303,22 @@
...
     </repositories>
 
     <build>
+        <extensions>
+            <extension>
+                <groupId>org.springframework.build</groupId>
+                <artifactId>aws-maven</artifactId>
+                <version>5.0.0.RELEASE</version>
+            </extension>
+        </extensions>
```
DwayneJengSage: I believe this is used to publish to a Maven repo, which is not necessary.

DwayneJengSage (approved):
Minor pom issues. Otherwise, looks good.

pom.xml
alxdarksage: Moved

pom.xml
alxdarksage: removed

PR #43, opened 2019-06-07
Elminating FindBugs errors.

alxdarksage:
The only issue I could not eliminate were calls like this: `Iterables.getFirst(iter, null)` the second argument is marked `@Nullable` but FindBugs still doesn't like it. Suppressed with `@SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION")`


src/main/java/org/sagebionetworks/bridge/services/ExternalIdService.java
```diff
@@ -74,6 +77,7 @@ public final void setSubstudyService(SubstudyService substudyService) {
         return externalIdDao.getExternalIds(studyId, offsetKey, pageSize, idFilter, assignmentFilter);
     }
     
+    @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION")
```
DwayneJengSage: It seems like this particular bug report is just, well, buggy. Instead of suppressing warnings piece-meal, it might be better to just use findbugs-exclude.xml to exclude this particular bug altogether.

```
  <Match>
    <!-- This bug pattern doesn't work correctly. -->
    <Bug pattern="NP_NONNULL_PARAM_VIOLATION" />
  </Match>
```

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -489,8 +490,13 @@ public void updateParticipant(Study study, StudyParticipant participant) {
                 externalIdService, substudyService, study, false);
         Validate.entityThrowingException(validator, participant);
         
-        Set<Roles> callerRoles = BridgeUtils.getRequestContext().getCallerRoles();
+        // FindBugs: Null passed for non-null parameter of BridgeUtils.collectExternalIds(Account).
+        // In actuality this will not happen (an exception will be thrown first).
+        if (account == null) {
+            return;
+        }
```
DwayneJengSage: Similarly, if we know for sure that this will never be null, would it be better to exclude the warning than to write unreachable code?

DwayneJengSage (approved):
Looks good. Two comments about excludes.

src/main/java/org/sagebionetworks/bridge/services/ExternalIdService.java
alxdarksage: I changed this as described, removed the annotation package.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Okay this one forced me to rethink what I was doing. Apparently all these shenanigans are to throw a 404 error rather than an invalid entity exception, but the validator catches the invalid substudy of the caller correctly, and this can be simplified if you just leave it with the invalid entity exception. And I don't see now why I insisted on a 404. So: cleaned up code, findbug errors go away, tests adjusted.

DwayneJengSage (approved):

PR #51, opened 2019-06-21
Templates

alxdarksage:
These are the metadata records for a stream of actual documents (TBD), and they are not articulated in any way with the existing system.


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -769,6 +771,17 @@ public void setAutoVerificationPhoneSuppressed(boolean autoVerificationPhoneSupp
         this.autoVerificationPhoneSuppressed = autoVerificationPhoneSuppressed;
     }
     
+    @DynamoDBTypeConvertedJson
```
DwayneJengSage: Do we want to use DynamoDB's default support for String to String maps instead?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        String defaultGuid = study.getDefaultTemplates().get(template.getTemplateType().name().toLowerCase());
+        if (template.getGuid().equals(defaultGuid)) {
+            throw new ConstraintViolationException.Builder().withMessage("The default template for a type cannot be deleted.")
+                .withEntityKey("guid", template.getGuid()).build();
+        }
+    }
+    
+    private String getKey(Template template) {
+        return "template:" + template.getGuid();
+    }
```
DwayneJengSage, line 252: It looks like instead of storing a CriteriaKey, we just always use "template:[templateGuid]". Can we update the design doc to reflect this? https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/808615943/Enhanced+Template+Repository

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplateDao.java
```diff
@@ -0,0 +1,106 @@
...
+                .withRequestParam(INCLUDE_DELETED, includeDeleted);
+    }
+    
+    @Override
+    public Optional<Template> getTemplate(StudyIdentifier studyId, String guid) { 
+        checkNotNull(guid);
+        
+        Template template = hibernateHelper.getById(HibernateTemplate.class, guid);
+        if (template != null && !template.getStudyId().equals(studyId.getIdentifier())) {
+            return Optional.empty();
```
DwayneJengSage, line 77: Unit test for template in wrong study?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplateDao.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public void deleteTemplatePermanently(StudyIdentifier studyId, String guid) {
+        checkNotNull(guid);
+        
+        Optional<Template> template = getTemplate(studyId, guid);
+        if (template.isPresent()) {
+            hibernateHelper.deleteById(HibernateTemplate.class, guid);    
+        }
```
DwayneJengSage, line 103: Unit test for deleting a template that doesn't exist?

src/test/java/org/sagebionetworks/bridge/hibernate/TemplatePersistenceExceptionConverterTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.testng.annotations.Test;
+
+import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
+
+public class TemplatePersistenceExceptionConverterTest {
+    
+    private TemplatePersistenceExceptionConverter converter;
+    
+    @Mock
+    private ConstraintViolationException cve;
```
DwayneJengSage: Minor: Unused?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        // The ideal case: one and only one template matches the user's context
+        if (templates.size() == 1) {
+            return templates.get(0);
+        }
+        // If not, fall back to the default specified for this study:
+        Study study = studyService.getStudy(context.getStudyIdentifier());
+        String defaultGuid = study.getDefaultTemplates().get(type.name().toLowerCase());
+        if (defaultGuid != null) {
+            return getTemplate(context.getStudyIdentifier(), defaultGuid);
+        }
```
DwayneJengSage, line 99: If more than one template matches, wouldn't it make more sense to pick one of the matching templates?

DwayneJengSage, line 99: Should we have a unit test for multiple templates and a study-specified default?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+
+        if (offset == null) {
+            offset = 0;
+        }
+        if (pageSize == null) {
+            pageSize = API_DEFAULT_PAGE_SIZE;
+        }
+        if (offset < 0) {
+            throw new BadRequestException("Invalid negative offset value");
+        } else if (pageSize < API_MINIMUM_PAGE_SIZE || pageSize > API_MAXIMUM_PAGE_SIZE) {
```
DwayneJengSage: Is there any particular reason to enforce a minimum page size?

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    @Test
+    public void getTemplate() {
+        Template template = Template.create();
+        template.setStudyId(TEST_STUDY.getIdentifier());
+        when(mockTemplateDao.getTemplate(TEST_STUDY, GUID)).thenReturn(Optional.of(template));
+        
+        Template result = service.getTemplate(TEST_STUDY, GUID);
+        assertSame(result, template);
```
DwayneJengSage, line 309: Should this test also check criteria?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        template.setGuid(generateGuid());
+        DateTime timestamp = getTimestamp();
+        template.setCreatedOn(timestamp);
+        template.setModifiedOn(timestamp);
+
+        Study study = studyService.getStudy(studyId);
+        Set<String> substudyIds = substudyService.getSubstudyIds(study.getStudyIdentifier());
+        
+        TemplateValidator validator = new TemplateValidator(study.getDataGroups(), substudyIds);
+        Validate.entityThrowingException(validator, template);
```
DwayneJengSage, line 171: Unit test for creating invalid template?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+    public GuidVersionHolder updateTemplate(StudyIdentifier studyId, Template template) {
+        checkNotNull(studyId);
+        checkNotNull(template);
+        
+        // If the entity is deleted and staying deleted, don't allow this operation.
+        Template existing = getTemplate(studyId, template.getGuid());
+        if (existing.isDeleted() && template.isDeleted()) {
+            throw new EntityNotFoundException(Template.class);
+        }
+        checkNotDeletingDefault(template, studyId);
```
DwayneJengSage: Doesn't this just prevent all updates of the default template?

DwayneJengSage: Unit test for deleting the default template via the update API?

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -0,0 +1,459 @@
...
+        template.setTemplateType(EMAIL_SIGN_IN);
+        template.setCriteria(criteria);
+        
+        GuidVersionHolder result = service.updateTemplate(TEST_STUDY, template);
+        assertEquals(result.getGuid(), GUID);
+        assertEquals(result.getVersion(), new Long(10));
+        
+        assertEquals(template.getStudyId(), TEST_STUDY_IDENTIFIER);
+        assertEquals(template.getGuid(), GUID);
+        assertEquals(template.getVersion(), 10);
```
DwayneJengSage, line 411: modifiedOn?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        template.setModifiedOn(getTimestamp());
+        // no reason for these to be updated after creation
+        template.setTemplateType(existing.getTemplateType());
+        template.setCreatedOn(existing.getCreatedOn());
+        
+        Study study = studyService.getStudy(studyId);
+        Set<String> substudyIds = substudyService.getSubstudyIds(study.getStudyIdentifier());
+        
+        TemplateValidator validator = new TemplateValidator(study.getDataGroups(), substudyIds);
+        Validate.entityThrowingException(validator, template);
```
DwayneJengSage, line 200: Unit test for updating an invalid template?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        
+        return new GuidVersionHolder(template.getGuid(), new Long(template.getVersion()));
+    }
+    
+    public void deleteTemplate(StudyIdentifier studyId, String guid) {
+        // This not only verifies the template exists, it verifies it is in the caller's study
+        Template existing = getTemplate(studyId, guid);
+        if (existing.isDeleted()) {
+            throw new EntityNotFoundException(Template.class);
+        }
```
DwayneJengSage, line 218: Unit test for deleting a deleted template?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        
+        templateDao.updateTemplate(existing);
+    }
+    
+    public void deleteTemplatePermanently(StudyIdentifier studyId, String guid) {
+        // Deletes fail quietly, we don't throw 404s if they're unnecessary 
+        Optional<Template> optional = templateDao.getTemplate(studyId, guid);
+        if (!optional.isPresent()) {
+            return;
+        }
```
DwayneJengSage: Unit test for deleting non-existent template?

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+    private String getKey(Template template) {
+        return "template:" + template.getGuid();
+    }
+    
+    private Criteria persistCriteria(Template template) {
+        Criteria criteria = template.getCriteria();
+        if (criteria == null) {
+            criteria = Criteria.create();
+        }
+        criteria.setKey(getKey(template));
```
DwayneJengSage, line 259: Unit tests for create and update should check that we set the criteria key.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -0,0 +1,254 @@
...
+        return criteriaDao.createOrUpdateCriteria(criteria);
+    }
+
+    private void loadCriteria(Template template) {
+        Criteria criteria = criteriaDao.getCriteria(getKey(template));
+        if (criteria == null) {
+            criteria = Criteria.create();
+        }
+        criteria.setKey(getKey(template));
```
DwayneJengSage: Unit tests for get should check that we set the criteria key.

src/test/java/org/sagebionetworks/bridge/spring/controllers/TemplateControllerTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Test
+    public void deleteTemplateDefault() throws Exception {
+        StatusMessage message = controller.deleteTemplate(GUID, null);
+        assertEquals(message.getMessage(), "Template deleted.");
+        
+        verify(mockTemplateService).deleteTemplate(TEST_STUDY, GUID);
+    }
+
+    @Test
+    public void deleteAppConfig() throws Exception {
```
DwayneJengSage: Nitpick: Copy-paste error on test name?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
alxdarksage: This worked fine, removing annotation.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Updated

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplateDao.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplateDao.java
alxdarksage: In a rush I didn't look at the design document for this... in fact we're supposed to throw a ENFE if the template does not exist, but that behavior is in the service... I'm changing the service and adding a test in the DAO that it silently does nothing (in practice this should not happen if you access through the service).

src/test/java/org/sagebionetworks/bridge/hibernate/TemplatePersistenceExceptionConverterTest.java
alxdarksage: Yes, removed

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Because matching occurs only if match criteria are provided by the user, it is easy to get false positives (an account with no metadata will match all the records in some cases). So we want developers to unambiguously indicate the fallback. This is also clearer for developers (app configs just return the first one and this has led to implementations that are routinely relying on this behavior, probably without knowing their matching criteria are ambiguous).

alxdarksage: Based on discussion I've made this more robust and added tests for the many fallback scenarios we now go through to try and return a template if at all possible.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Only that we have done it before. I can make it min=1.

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Added. Added for update as well.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Aaargh. Yes it does. Changed this and added a test for it.

alxdarksage: I believe cannotUpdateToDeleteDefaultTemplate() does this. Added a test that you *can* update the default template as long as you're not trying to mark it deleted.

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Should be "deleteTemplateNotFound" - renamed test and added test for case where template is entirely missing.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: This behavior was changed based on the design docs for logical/physical delete and tests were updated. It now throws an ENFE.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: added

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/spring/controllers/TemplateControllerTest.java
alxdarksage: yes. fixed

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -0,0 +1,575 @@
...
+    @Test(expectedExceptions = BadRequestException.class)
+    public void getTemplatesPageSizeAboveMax() {
+        service.getTemplatesForType(TEST_STUDY, EMAIL_RESET_PASSWORD, null, 1000, true);
+    }
+
+    @Test
+    public void getTemplate() {
+        Criteria criteria = Criteria.create();
+        criteria.setKey("template:"+GUID1); // this is persisted as part of criteria, not set on load
+        when(mockCriteriaDao.getCriteria("template:"+GUID1)).thenReturn(criteria);
```
DwayneJengSage, line 301: > this is persisted as part of criteria, not set on load

This may be logically true, but loadCriteria() has code to set the criteriaKey. Is this redundant code that should be removed? Or is it code that we need and that we should test it?

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -0,0 +1,575 @@
...
+        Template existing = Template.create();
+        existing.setStudyId(TEST_STUDY_IDENTIFIER);
+        existing.setGuid(GUID1);
+        existing.setTemplateType(EMAIL_ACCOUNT_EXISTS);
+        when(mockTemplateDao.getTemplate(TEST_STUDY, GUID1)).thenReturn(Optional.of(existing));
+
+        service.deleteTemplatePermanently(TEST_STUDY, GUID1);
+
+        verify(mockTemplateDao).deleteTemplatePermanently(TEST_STUDY, GUID1);
+    }
```
DwayneJengSage, line 521: Should we also verify that we delete the criteria?

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -0,0 +1,575 @@
...
+    public void createTemplateInvalid() {
+        service.createTemplate(TEST_STUDY, Template.create());
+    }
+    
+    @Test
+    public void updateTemplate() {
+        doAnswer(answer -> {
+            Template captured = answer.getArgument(0);
+            captured.setVersion(10);
+            captured.setModifiedOn(TIMESTAMP);
```
DwayneJengSage: TemplateService.updateTemplate() already sets the modifiedOn, and the dao doesn't. Wouldn't setting the template here actually be incorrect?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: So thinking about this further, we don't serialize the key in the JSON we return, and we always set it before persistence, so we don't really need the key set when loading (it's going to be there from loading though). I removed this and a couple of lines in the test that are looking for the key being set by the service on a load.

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: added

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: You're right, removed this, test is unaffected

DwayneJengSage (approved):

PR #60, opened 2019-07-02
Template revisions.

alxdarksage:
I would also like to add a method that creates all the default templates for when a new study is created, I will probably add that to this CR and the TemplateRevisionsService. This will begin the move of all the default templates out of the StudyService.


src/main/java/org/sagebionetworks/bridge/hibernate/TemplateRevisionPersistenceExceptionConverter.java
```diff
@@ -0,0 +1,22 @@
...
+
+import javax.persistence.OptimisticLockException;
+import javax.persistence.PersistenceException;
+
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
+
+@Component
+public class TemplateRevisionPersistenceExceptionConverter implements PersistenceExceptionConverter {
```
DwayneJengSage: This is identical to TemplatePersistenceExceptionConverter. Should we consider consolidating them to a single PersistenceExceptionConverter and just cast the entity to a BridgeEntity?

src/main/java/org/sagebionetworks/bridge/services/TemplateRevisionService.java
```diff
@@ -0,0 +1,127 @@
...
+    public void publishTemplateRevision(StudyIdentifier studyId, String templateGuid, DateTime createdOn) {
+        checkNotNull(studyId);
+        checkNotNull(templateGuid);
+        checkNotNull(createdOn);
+        
+        Template template = templateDao.getTemplate(studyId, templateGuid)
+                .orElseThrow(() -> new EntityNotFoundException(Template.class));
+        
+        templateRevisionDao.getTemplateRevision(templateGuid, createdOn)
+                .orElseThrow(() -> new EntityNotFoundException(TemplateRevision.class));
```
DwayneJengSage, line 114: Is it worth calling getTemplateRevision() instead of re-writing the same code?

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
```diff
@@ -355,9 +398,16 @@ public void createTemplate() {
...
         
+        verify(mockTemplateRevisionDao).createTemplateRevision(revisionCaptor.capture());
         verify(mockCriteriaDao).createOrUpdateCriteria(criteria);
-        verify(mockTemplateDao).createTemplate(template);
+        verify(mockTemplateDao).createTemplate(eq(template), any());
+        
+        TemplateRevision revision = revisionCaptor.getValue();
+        assertEquals(revision.getSubject(), EMAIL_RESET_PASSWORD.name());
+        assertEquals(revision.getDocumentContent(), EMAIL_RESET_PASSWORD.name());
+        assertEquals(revision.getMimeType(), HTML);
```
DwayneJengSage, line 417: Should we verify the revision properties set by the create API?
* createdBy
* createdOn
* templateGuid
* storagePath

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
```diff
@@ -151,30 +292,52 @@ public Template getTemplate(StudyIdentifier studyId, String guid) {
...
+        template.setPublishedCreatedOn(timestamp);
         
         Criteria criteria = persistCriteria(template);
         template.setCriteria(criteria);
         
-        templateDao.createTemplate(template);
+        revision.setCreatedBy(getUserId());
+        revision.setCreatedOn(timestamp);
+        revision.setTemplateGuid(templateGuid);
+        revision.setCreatedBy(getUserId());
```
DwayneJengSage: Minor: createdBy listed twice

src/main/java/org/sagebionetworks/bridge/validators/TemplateRevisionValidator.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        if (isBlank(revision.getStoragePath())) {
+            errors.rejectValue("storagePath", "cannot be blank");
+        }
+        if (revision.getMimeType() == null) {
+            errors.rejectValue("mimeType", "cannot be null");
+        }
+        if (revision.getSubject() != null && revision.getSubject().length() > 250) {
+            errors.rejectValue("subject", "must be 250 characters of less");
+        }
```
DwayneJengSage, line 40: Unit test for subject too long?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/TemplateRevisionPersistenceExceptionConverter.java
alxdarksage: Converted three uninteresting persistence exception converters into one BasicPersistenceExceptionConverter. (The AccountSecretsExceptionConverter also did nothing but pass along the exception).

src/main/java/org/sagebionetworks/bridge/services/TemplateRevisionService.java
alxdarksage: The getTemplateRevision() method does not return the Template which we need to update, so we'd be retrieving the Template object twice. I would leave it like this.

src/test/java/org/sagebionetworks/bridge/services/TemplateServiceTest.java
alxdarksage: Definitely, added.

src/main/java/org/sagebionetworks/bridge/services/TemplateService.java
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/validators/TemplateRevisionValidator.java
alxdarksage: Added

DwayneJengSage (approved):

PR #61, opened 2019-07-03
add fitbit scopes to study

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2522

Prelimary work in Bridge Server to add the FitBit scopes to a study. Used plain old strings instead of a JSON so we can add new scopes to the Worker without needing to make the corresponding change to Bridge Server.


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
```diff
@@ -61,6 +61,7 @@
     private String identifier;
     private Map<String, String> automaticCustomEvents;
     private boolean autoVerificationEmailSuppressed;
+    private List<String> fitBitScopes;
```
alxdarksage, line 64: Not a set?

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoStudy.java
DwayneJengSage: For some strange reason, DynamoDB supports empty lists but not empty sets. I feel like it's cleaner to use "this is never null" semantics other than "this is never empty".

PR #65, opened 2019-07-15
External ids cleanup

alxdarksage:
- With this commit, the server always treats external IDs as managed. They must always be created in the system (associated to a substudy) before they are declared for an account.
- Added a migration pathway to migrate selected accounts to use a managed external ID.
- Switched delete external ID to be an admin function (there is no logical delete).
- When the set/getExternalId() field is removed from Account in a later commit, it may be possible to simplify some of the business logic around this. Not yet though.


src/main/java/org/sagebionetworks/bridge/services/ExternalIdService.java
```diff
@@ -104,10 +104,6 @@ public void deleteExternalIdPermanently(Study study, ExternalIdentifier external
         checkNotNull(study);
         checkNotNull(externalId);
         
-        if (study.isExternalIdValidationEnabled()) {
-            throw new BadRequestException("Cannot delete IDs while externalId validation is enabled for this study.");
-        }
```
alxdarksage: This has been a real pain, so I am switching to making this call and ADMIN-only call (because we only have physical deletes for external IDs). This doesn't remove the account and is pretty safe because you can undo this later if need be by restoring the ID and re-associating it to an account.

DwayneJengSage (approved):

PR #69, opened 2019-07-23
Reauth enabled for new studies by default

DwayneJengSage:
iOS SDK assumes that reauth is enabled for all studies. Also, now that we've fixed the 409s issue, there's no reason not to enable reauth for all new studies.

This change makes the reauthEnabled flag nullable, and if not specified on study create, it defaults to false. This is so that we don't suddenly accidentally turn on reauth for old studies.


alxdarksage (approved):

alxdarksage (changes requested):
A caller could null this out on an update and in effect, disable reauthentication. I would require this flag to be set in the validator, and set the default on create before calling the validator. This would mean (in Swagger) that x-nullable = false (we'll never return it as a null field or a missing field) and default = true, required = false (doesn't need to be provided, we'll default it to true).

alxdarksage (approved):

PR #70, opened 2019-07-24
Channel SignIn no longer resets throttle timer

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2413

Note that we don't have a good way to test Redis expiration, so neither the unit tests nor integ tests have changed. This was tested manually by setting channel throttle timeout to 60 seconds and then making requests 45 seconds apart. Under the old code, the third request would still be throttled. Under the new code, the third request would go through.

Alternate implementations:

* Instead of resetting the expiration, set the expiration to the current TTL. This however will allow the participant to burst requests at the end of the expiration period, and it's more complicated to implement.
* Instead of setting a count, just set a boolean that expires after the timeout. This is simple to implement, but we won't be able to use X requests in Y minutes logic. (That said, currently, X is always 1.)

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
@@ -9,7 +9,6 @@ import static org.sagebionetworks.bridge.validators.SignInValidator.EMAIL_SIGNIN
 import static org.sagebionetworks.bridge.validators.SignInValidator.PHONE_SIGNIN_REQUEST;
 
 import java.io.IOException;
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -708,16 +710,29 @@ private boolean isRequestThrottled(ThrottleRequestType type, String userId) {
...
+            // exceed the throttle limit by making a bunch of requests at the end of the throttle window.
+            jedisOps.setex(cacheKey, channelThrottleTimeoutSeconds, String.valueOf(numRequests + 1));
+            return false;
+        } else {
+            // We've seen at least as many requests as our limit (or more), so this next request will push us over the
+            // limit. Throttle this.
+            // Don't update Redis, since we throttle email/SMS sent, not requests. This allows the expiration to expire
+            // naturally.
+            return true;
+        }
```
alxdarksage, line 675: Doesn't need to happen here, but this throttling logic seems very useful to factor out so we can use it a number of places. In addition, I think it's fair to throw an exception that would return 429 status (too many requests) when the user hits the limit, so they understand what happened (and we see throttling in the logs). See BRIDGE-2415.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -708,16 +710,29 @@ private boolean isRequestThrottled(ThrottleRequestType type, String userId) {
         // Generate key, which is in the form of channel-throttling:[type]:[userId].
         String cacheKey = "channel-throttling:" + type.toString().toLowerCase() + ":" + userId;
```
alxdarksage: If we use CacheProvider and CacheKey, we have a definitive list of all the keys under which we're writing values... I would prefer this but only because I'm pretty OCD.

alxdarksage (approved):

PR #72, opened 2019-07-25
add dayInStudy to health data record

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2531

Add dayInStudy to the health data record, following the convention for all the other fields we add to the health data record.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
+++ b/src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
@@ -2,7 +2,6 @@ package org.sagebionetworks.bridge.upload;
 
 import static org.sagebionetworks.bridge.BridgeUtils.mapSubstudyMemberships;
 
-import java.util.Map;
 import javax.annotation.Nonnull;
 
 import org.joda.time.DateTime;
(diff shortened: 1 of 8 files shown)


src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
```diff
@@ -34,6 +48,23 @@ public void handle(@Nonnull UploadValidationContext context) {
...
+                // activities_retrieved with calendar date 2019-07-24. Therefore, 2019-07-24 would be day 1, 2019-07-25
+                // would be day 2, etc.
+                LocalDate activitiesRetrievedLocalDate = activitiesRetrievedDateTime
+                        .withZone(BridgeConstants.LOCAL_TIME_ZONE).toLocalDate();
+                LocalDate todayLocalDate = DateUtils.getCurrentCalendarDateInLocalTime();
+
+                // "dayInStudy" is 1-indexed. The first day of the study is day 1. Therefore, add 1.
+                int dayInStudy = Days.daysBetween(activitiesRetrievedLocalDate, todayLocalDate).getDays() + 1;
+                record.setDayInStudy(dayInStudy);
+            }
```
alxdarksage: There may be studies that don't use scheduling, for which there will not be an activities retrieved timestamp. You might consider falling back to enrollment, then to the creation timestamp of the account record (it might be easiest to add that creation timestamp to the event map from ActivityEventService, if we're not doing that already).

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
DwayneJengSage: Made a fallback from activities_retrieved to enrollment to account.createdOn. I decided against writing a new ActivityEvent, because we'd then have to backfill it.

Made this a new method in ParticipantService. Might be useful for the report that Janae is building.

alxdarksage (approved):

PR #77, opened 2019-08-02
Move some functionality of CriteriaContext to RequestContext which is available across all code paths.

alxdarksage:
We're now heading toward selecting templates all over the place based on what we know about the user... most importantly their app version and possibly their preferred language. Toward that end I think CriteriaContext can be phased out in favor of carrying over the same information in RequestContext as a thread local variable. (SchedulingContext is different because a lot of the context is passed in by the user and will probably stay.) RequestContext is all about what we know about the caller that could influence the response.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
@@ -77,6 +77,7 @@ import org.sagebionetworks.bridge.services.StudyService;
 import org.sagebionetworks.bridge.services.AuthenticationService.ChannelType;
 
 public class AuthenticationControllerTest extends Mockito {
+    private static final String USER_AGENT_STRING = "App/14 (Unknown iPhone; iOS/9.0.2) BridgeSDK/4";
     private static final String DOMAIN = "localhost";
     private static final DateTime NOW = DateTime.now();
     private static final String REAUTH_TOKEN = "reauthToken";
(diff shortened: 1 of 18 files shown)


src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -714,10 +717,15 @@ public void emailSignInBlockedByVersionKillSwitch() throws Exception {
...
-        when(mockRequest.getHeader(USER_AGENT)).thenReturn("App/14 (Unknown iPhone; iOS/9.0.2) BridgeSDK/4");
+        
+        BridgeUtils.setRequestContext(new RequestContext.Builder()
+                .withCallerClientInfo(ClientInfo.fromUserAgentCache("App/14 (Unknown iPhone; iOS/9.0.2) BridgeSDK/4"))
+                .build());
         study.getMinSupportedAppVersions().put(OperatingSystem.IOS, 20);
         
         controller.emailSignIn();
+        
+        BridgeUtils.setRequestContext(null);
```
DwayneJengSage: Minor: This will never actually run, because controller.emailSignIn() throws. You would need a finally block.

DwayneJengSage (approved):

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
alxdarksage: This exposed a real can of worms. I moved this clear to an @AfterMethod, and that in turn caused subsequent tests to fail. Thread local variables aren't cleared between test runs so you must always remember to clear them or there will be trouble (test will succeed individually, but fail when run as part of the test suite). I've found and fixed places where I was not doing this correctly.

PR #78, opened 2019-08-02
Some final changes before adding migration code

alxdarksage:
- Changed TemplateService's `getTemplateForUser` to `getTemplateRevisionForUser` because we actually want the revision, not the template;
- Added the `TemplateMigrationService` which can be called on update idempotently to move over the study templates to the enhanced templates system.
- On a physical delete of a study, physical delete all the templates (this in turn removes all the revisions). A lot of junk was being left behind by tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -66,7 +66,6 @@ import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
-import org.sagebionetworks.bridge.models.PagedResourceList;
 import org.sagebionetworks.bridge.models.accounts.IdentifierHolder;
 import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
 import org.sagebionetworks.bridge.models.studies.EmailTemplate;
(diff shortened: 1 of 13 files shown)


src/test/java/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -853,6 +863,15 @@ public void createStudyWithoutConsentNotificationEmailDoesNotSendNotification()
...
     public void physicallyDeleteStudy() {
+        PagedResourceList<? extends Template> page1 = new PagedResourceList<>(
+                ImmutableList.of(createTemplate("guid1"), createTemplate("guid2"), createTemplate("guid3")), 3);
+        PagedResourceList<? extends Template> page2 = new PagedResourceList<>(ImmutableList.of(), 3);
+
+        doReturn(page1, page2).when(mockTemplateService).getTemplatesForType(
+                TEST_STUDY_IDENTIFIER, EMAIL_ACCOUNT_EXISTS, 0, 50, true);
+        doReturn(page2).when(mockTemplateService).getTemplatesForType(eq(TEST_STUDY_IDENTIFIER), 
+                not(eq(EMAIL_ACCOUNT_EXISTS)), eq(0), eq(50), eq(true));
+
```
DwayneJengSage, line 874: We should verify calls to deleteTemplatePermanently().

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -733,6 +741,18 @@ public void deleteStudy(String identifier, boolean physical) {
...
     
+    private void deleteAllTemplates(StudyIdentifier studyId) {
+        for (TemplateType type : TemplateType.values()) {
+            PagedResourceList<? extends Template> page;
+            do {
+                page = templateService.getTemplatesForType(studyId, type, 0, 50, true);
+                for (Template template : page.getItems()) {
+                    templateService.deleteTemplatePermanently(studyId, template.getGuid());    
+                }
+            } while(!page.getItems().isEmpty());
```
DwayneJengSage: How do indices in SQL work? Would it be better to create a getTemplatesForStudy() API instead of making a bunch of micro-calls for each template type?

DwayneJengSage: We should also consider adding a maximum here, in case a bug somewhere causes this code to go into an infinite loop.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Added a method to specifically do this delete... there's no reason the DAO can't do this.

src/test/java/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #86, opened 2019-08-15
Need to preserve the externalId field for backwards compatibility.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/models/accounts/StudyParticipantTest.java
```diff
@@ -407,4 +407,22 @@ public void duplicateLanguagesAreRemoved() {
         
         assertEquals(participant.getLanguages(), Lists.newArrayList("en","fr","de"));
     }
+    
+    @Test
+    public void setExternalIdReturned() {
+        StudyParticipant participant = new StudyParticipant.Builder().withExternalId("oneExternalId").build();
+        assertEquals(participant.getExternalId(), "oneExternalId");
+    }
```
DwayneJengSage, line 415: Need test for getting external ID from map.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #88, opened 2019-08-19
Selectable by criteria

alxdarksage:
BRIDGE-2542: we currently select the first language a user provides as an acceptable alternative, rather than looking to see if we have the best fit language for the user. Fixed here with some new utility methods that take the entire set of Criteria-filterable objects and sorting all matches by language weight.

Also refactored CriteriaUtilsTest because I was having a hard time following what was being set in the criteria for each test.


src/main/java/org/sagebionetworks/bridge/services/AppConfigService.java
```diff
@@ -112,10 +111,8 @@ public AppConfig getAppConfigForUser(CriteriaContext context, boolean throwExcep
 
         List<AppConfig> appConfigs = getAppConfigs(context.getStudyIdentifier(), false);
 
-        List<AppConfig> matches = appConfigs.stream().filter(oneAppConfig -> {
-            return CriteriaUtils.matchCriteria(context, oneAppConfig.getCriteria());
-        }).sorted(Comparator.comparingLong(AppConfig::getCreatedOn))
-          .collect(BridgeCollectors.toImmutableList());
+        List<AppConfig> matches = CriteriaUtils.filterByCriteria(context, appConfigs);
+        matches.sort(Comparator.comparingLong(AppConfig::getCreatedOn));
```
alxdarksage: So are we returning the most recent match and *not* sorting by language here? It seems we need to support language sorting here too.

src/main/java/org/sagebionetworks/bridge/models/CriteriaUtils.java
```diff
@@ -24,16 +27,39 @@
...
+        final List<String> langs = context.getLanguages();
+        Comparator<T> comparator = (sel1, sel2) -> {
+            int posLang1 = langs.indexOf(sel1.getCriteria().getLanguage());
+            int posLang2 = langs.indexOf(sel2.getCriteria().getLanguage());
+            return posLang1 - posLang2;
+        };
+        // In the app config case, sort by createdOn timestamp as well
+        if (secondComparator != null) {
+            comparator = comparator.thenComparing(secondComparator);
+        }
```
DwayneJengSage, line 45: Do we have a unit test for sorting by a secondary comparator?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/models/CriteriaUtils.java
alxdarksage: Added test.

DwayneJengSage (approved):

PR #89, opened 2019-08-20
call OAuth Introspect API to get scopes

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2522

There will be a subsequent pull request to delete the FitBitScopes field from the study.


src/main/java/org/sagebionetworks/bridge/services/OAuthProviderService.java
```diff
@@ -242,7 +281,31 @@ protected OAuthAccessGrant jsonToGrant(JsonNode node) {
...
+                continue;
+            }
+
+            int equalIdx = scopePair.indexOf('=');
+            scopeList.add(scopePair.substring(0, equalIdx));
+        }
+
+        return scopeList;
+    }
+
```
alxdarksage, line 327: Are we asking for all the scopes that Fitbit tells us exist? If a partner asks us to get scopes A, B, and C, and Fitbit has A, B, C, and D, do we ask for A-C or A-D? Just asking for clarification...

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/services/OAuthProviderService.java
DwayneJengSage: RFC7662 defines this as scopes associated with the token, not all possible scopes. I'll add to the comment to clarify and add an example of the full response in the comments.

DwayneJengSage: Further clarification: Bridge doesn't ask OAuth for specific scopes. Scopes are defined by the call to OAuth authorization, which happens in the app. Scopes are locked in at this time, and Bridge Server cannot change them. What Bridge Server does is query the token for what scopes are associated with it.

alxdarksage: Okay great, thanks.

PR #93, opened 2019-08-22
Move signature block into the content of the consent.

alxdarksage:
See: https://sagebionetworks.jira.com/browse/BRIDGE-769

The signature block can now be changed by end users, the labels can be localized, etc. As long as they leave the name and date somewhere in the document (otherwise, well add the signature block to the end of the document). This will change existing templates as well so no migration is needed.

There is also a very minor change to the integration tests once this has been changed: https://github.com/Sage-Bionetworks/BridgeIntegrationTests/pull/378


src/main/java/org/sagebionetworks/bridge/services/StudyConsentService.java
```diff
@@ -271,23 +296,27 @@ private String loadDocumentContent(StudyConsent consent) {
...
-        return document.body().html();
+    private String appendSignatureBlockIfNeeded(String content) {
+        // The user can change the signature block, they can remove parts of the signature block,
+        // but if the person's name or the signing date are gone, we're assuming that they've 
+        // removed too much, and we re-append the signature block.
+        if (!content.contains("${participant.name}") || !content.contains("${participant.signing.date}")) {
+            content = content + SIGNATURE_BLOCK;
+        }
+        return content;
     }
```
DwayneJengSage, line 315: Do we have any unit tests for where if the signature block already exists, we don't add another signature block?

src/test/java/org/sagebionetworks/bridge/services/ConsentPdfTest.java
```diff
@@ -207,21 +184,13 @@ private static ConsentSignature makeInvalidSignature() {
...
-        assertTrue(bodyContent.contains(dateStr), "Signing date correct");
-        assertTrue(bodyContent.contains("|Test Person|"), "Name correct");
-        assertTrue(bodyContent.contains("|user@user.com|"), "User email correct");
-        assertTrue(bodyContent.contains("|Not Sharing|"), "Sharing correct");
-        assertTrue(bodyContent.contains("<html><head></head><body>Passed through as is."), "HTML markup preserved");
-    }
-
-    private static void validateNewDocBody(String bodyContent) throws Exception {
+    private static void validateDocBody(String bodyContent) throws Exception {
+        System.out.println(bodyContent);
```
DwayneJengSage: Minor: Remove println

DwayneJengSage (changes requested):
Some minor comments.

src/main/java/org/sagebionetworks/bridge/services/StudyConsentService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/ConsentPdfTest.java
alxdarksage: removed

DwayneJengSage (approved):

PR #94, opened 2019-08-23
Create study_start_date event 

janaechan:
Made new getActivityEventList() and getActivityEventMap() that accept an Account instead of the healthCode string because I needed the studyId as well to find a StudyParticipant by healthCode.


src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -147,9 +165,58 @@ public void publishActivityEvent(ActivityEvent event) {
...
+
+            DateTime timestamp = entry.getValue();
+            if (timestamp !=null) {
+                event.setTimestamp(timestamp.getMillis());
+            }
+
+            activityEventList.add(event);
+        }
+        return activityEventList;
+    }
```
alxdarksage: The problem with this approach is that the existing methods that take healthCode will now return different results than the methods that take an account, which is unnecessarily confusing. What about removing the path through ParticipantService entirely, and adding the AccountDao to the ActivityEventService? If you need the account, you can load it from the methods that take a health code (you don't always need it, as we discussed, and the ParticipantService isn't really needed at all since createdOn is part of the basic Account object). Although looking at it, you may just be able to pass in the Account from the ParticipantService rather than calling back to it... if that makes sense, then we should probably remove the healthCode related methods so there's only one code path that assembles the events map.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -241,6 +242,12 @@ public StudyParticipant getParticipant(Study study, String userId, boolean inclu
         return getParticipant(study, account, includeHistory);
     }
     
+    public StudyParticipant getParticipant(Study study, String healthCode) {
+        AccountId accountId = AccountId.forHealthCode(study.getIdentifier(), healthCode);
+        Account account = getAccountThrowingException(accountId);
+        return getParticipant(study, account, false);
```
alxdarksage: I'm not sure why you need this since ParticipantController.getActivityEventsForWorker() calls ParticipantService.getActivityEvents() and that already loads an Account. Probably you want to remove the activityEventService.getActivityEventList() method that takes a healthCode, and just pass that Account to it... ?

alxdarksage:
Basically, the service method that loads the persisted events is ActivityEventService.getActivityEventMap(String healthCode). You'll have to do all the work there, so that no view of the activity events map is inconsistent. Because scheduling only has the healthCode, you'd have to do a lot of refactoring to make that method take an account. It would be easier to load the account (again in some code paths, unfortunately) within that method, if you need it. You already have that code in AccountDao and don't need to create it again in ParticipantService, which loads consents, etc. and is  more heavy.

alxdarksage:
Also remember that changes need to be accompanied with unit tests.

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -147,9 +165,58 @@ public void publishActivityEvent(ActivityEvent event) {
...
+        
+        if (activitiesRetrievedDateTime != null) {
+            studyStartDateTime = activitiesRetrievedDateTime;
+        } else if (enrollmentDateTime != null) {
+            studyStartDateTime = enrollmentDateTime;
+        } else if (studyStartDateTime == null) {
+            Study study = studyService.getStudy(account.getStudyId());
+            StudyParticipant studyParticipant = participantService.getParticipant(study, account.getHealthCode());
+            studyStartDateTime = studyParticipant.getCreatedOn();
+            builder.put(ActivityEventObjectType.STUDY_START_DATE.name().toLowerCase(), studyStartDateTime);
```
DwayneJengSage: Note that the first 2 branches don't actually save the study_start_date anywhere.

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -147,9 +165,58 @@ public void publishActivityEvent(ActivityEvent event) {
...
 
+    /**
+     * Gets the activity events times for a specific user in order to schedule against them.
+     */
+    public Map<String, DateTime> getActivityEventMap(Account account) {
+        checkNotNull(account.getHealthCode());
+        Map<String, DateTime> activityMap = activityEventDao.getActivityEventMap(account.getHealthCode());
+        Builder<String, DateTime> builder = ImmutableMap.<String, DateTime>builder();
+        
+        DateTime studyStartDateTime = activityMap.get(ActivityEventObjectType.STUDY_START_DATE.name().toLowerCase());
```
DwayneJengSage: Is it worth loading study_start_date from the activities map? This is likely to change, and it's not that expensive to re-compute. Also note that this activity event isn't actually being saved anywhere.

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -147,9 +165,58 @@ public void publishActivityEvent(ActivityEvent event) {
...
+        } else if (studyStartDateTime == null) {
+            Study study = studyService.getStudy(account.getStudyId());
+            StudyParticipant studyParticipant = participantService.getParticipant(study, account.getHealthCode());
+            studyStartDateTime = studyParticipant.getCreatedOn();
+            builder.put(ActivityEventObjectType.STUDY_START_DATE.name().toLowerCase(), studyStartDateTime);
+        }
+        
+        for (Map.Entry<String, DateTime> entry : activityMap.entrySet()) {
+            builder.put(entry.getKey(), entry.getValue());
+        }
```
DwayneJengSage, line 194: Would builder.putAll() be better?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -141,17 +177,44 @@ public void publishActivityEvent(ActivityEvent event) {
...
+        DateTime activitiesRetrieved = activityMap.get(ActivityEventObjectType.ACTIVITIES_RETRIEVED.name().toLowerCase());
+        DateTime enrollment = activityMap.get(ActivityEventObjectType.ENROLLMENT.name().toLowerCase());
+        DateTime createdOn = activityMap.get(ActivityEventObjectType.CREATED_ON.name().toLowerCase());
+        if (createdOn == null) {
+            Study study = studyService.getStudy(studyId);
+            AccountId accountId = AccountId.forHealthCode(studyId, healthCode);
+            Account account = BridgeUtils.filterForSubstudy(accountDao.getAccount(accountId));
+            if (account == null) {
+                throw new EntityNotFoundException(Account.class);
+            }
```
DwayneJengSage: You don't need AccountDao. You can just call ParticipantService.getParticipant() and pass in "healthcode:(healthCode)" https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java#L251

src/test/java/org/sagebionetworks/bridge/services/ActivityEventServiceTest.java
```diff
@@ -112,20 +132,142 @@ public void canPublishEvent() {
...
+        
+        Map<String,DateTime> map = Maps.newHashMap();
+        map.put("activities_retrieved", activitiesRetrieved);
+        map.put("enrollment", enrollment);
+        map.put("created_on", createdOn);
+        when(activityEventDao.getActivityEventMap("BBB")).thenReturn(map);
+        
+        Study study = Study.create();
+        study.setIdentifier(STUDY_ID);
+        when(mockStudyService.getStudy(STUDY_ID)).thenReturn(study);
```
DwayneJengSage: This test case has createdOn in the events map, so studyService, accountDao, and participantService will never be called. Not only do you not need to mock them, we should verify they are never called.

`verify(mockStudyService, never()).getStudy(any());`

src/test/java/org/sagebionetworks/bridge/services/ActivityEventServiceTest.java
```diff
@@ -112,20 +132,142 @@ public void canPublishEvent() {
...
+        when(mockAccountDao.getAccount(accountId)).thenReturn(account);
+        
+        StudyParticipant studyParticipant = new StudyParticipant.Builder().withCreatedOn(now).build();
+        when(mockParticipantService.getParticipant(study, account, false)).thenReturn(studyParticipant);
+        
+        Map<String, DateTime> results = activityEventService.getActivityEventMap(STUDY_ID, "BBB");
         assertEquals(results.get("enrollment"), now);
-        assertEquals(results.size(), 1);
+        assertEquals(results.get("study_start_date"), now);
+        assertEquals(results.size(), 3);
```
DwayneJengSage, line 180: We should verify that created_on was created. Also, we should use different a different date for created_on, so we can make sure the correct date was used.

Also, maybe have only one test case that creates created_on (probably in a separate test case), so you don't have to set up the mock every time.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -296,6 +296,21 @@ public String getParticipantForWorker(@PathVariable String studyId, @PathVariabl
         ObjectWriter writer = StudyParticipant.API_WITH_HEALTH_CODE_WRITER;
         return writer.writeValueAsString(participant);
     }
+    
+    @GetMapping(path="/v3/studies/{studyId}/participants/{userId}/requestInfo")
+    public RequestInfo getRequestInfoForWorker(@PathVariable String studyId, @PathVariable String userId) throws Exception {
+        getAuthenticatedSession(WORKER);
+        Study study = studyService.getStudy(studyId);
```
DwayneJengSage: We don't need to load the study if all we're using is the studyId.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -296,6 +296,21 @@ public String getParticipantForWorker(@PathVariable String studyId, @PathVariabl
...
+
+        // Verify it's in the same study as the researcher.
+        RequestInfo requestInfo = cacheProvider.getRequestInfo(userId);
```
DwayneJengSage: Note that we recently changed this, so now requestInfo comes from RequestInfoService.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantControllerTest.java
```diff
@@ -556,6 +559,35 @@ public void getParticipantRequestInfoOnlyReturnsCurrentStudyInfo() throws Except
...
+    
+    @Test
+    public void getParticipantRequestInfoForWorkerIsNullsafe() throws Exception {
+        participant = new StudyParticipant.Builder().copyOf(participant).withRoles(ImmutableSet.of(WORKER)).build();
+        session.setParticipant(participant);
+        // There is no request info.
+        RequestInfo result = controller.getRequestInfoForWorker(study.getIdentifier(), "userId");
+
+        assertNotNull(result); // values are all null, but object is returned
+    }
```
DwayneJengSage, line 593: Add a test case for when the requestInfo is from another study. (Equivalent of getParticipantRequestInfoOnlyReturnsCurrentStudyInfo.)

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/ActivityEventServiceTest.java
```diff
@@ -113,29 +133,104 @@ public void canPublishEvent() {
...
+        assertEquals(results.get("created_on"), CREATED_ON);
+        assertEquals(results.get("study_start_date"), ACTIVITIES_RETRIEVED);
+        assertEquals(results.size(), 4);
+        
+        verify(activityEventDao).getActivityEventMap(HEALTH_CODE);
+        verify(mockStudyService, never()).getStudy(anyString());
+        verify(mockParticipantService, never()).getParticipant(any(), anyString(), anyBoolean());
+    }
+    
+    public void canSetStudyStartDateWithEnrollment() {
```
DwayneJengSage, line 170: It looks like the Test annotation got lost on a bunch of tests.

DwayneJengSage (changes requested):
Looks good. Just one minor comment about missing Test annotations.

Also, be sure to run `mvn verify` on your workspace one last time. This runs all the unit tests, FindBugs tests, and runs code coverage. I'd also recommend running `mvn spring-boot:run` and running your integration tests (I usually just run the ones that call the code you changed) to make sure everything's working correctly.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -507,6 +507,18 @@ public Study updateStudy(Study study, boolean isAdminUpdate) {
...
+            for (TemplateType type: TemplateType.values()) {
+                String typeName = type.name().toLowerCase();
+                Template template = Template.create();
+                template.setName(BridgeUtils.templateTypeToLabel(type));
+                template.setTemplateType(type);
+                GuidVersionHolder keys = templateService.createTemplate(study, template);
+                map.put(typeName, keys.getGuid());
+            }
+            study.setDefaultTemplates(map);
+        }
```
DwayneJengSage: You should delete this. It was a temporary patch to get your local server running, but we don't actually want to check this in.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #96, opened 2019-08-27
Relax the rules around adding substudies and external IDs to accounts.

alxdarksage:
The "new rules" (BRIDGE-2548):

Users can sign up with external IDs/substudies as before, and they can use the updateIdentifiers API to add an external ID after account creation.

Researchers can add external IDs and substudies on creation, and add external IDs after creation. This in turn would associate the account to new substudies, but they cannot delete these relationships.

Admins can add these relationships on create/update, but unlike everyone else, they can also remove them. This would screw up production data, so it remains difficult to do.

I did have to move code around in the ParticipantService to avoid hacks, for example, I had to move substudy authorization checks out of the validator.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -190,32 +190,6 @@ public class ParticipantService {
         this.requestInfoService = requestInfoService;
     }
     
-    public ExternalIdentifier assignExternalId(AccountId accountId, String externalId) {
-        Account account = getAccountThrowingException(accountId);
-        ExternalIdentifier extIdObj = beginAssignExternalId(account, externalId);
-        if (extIdObj != null) {
-            AccountSubstudy acctSubstudy = AccountSubstudy.create(account.getStudyId(),
-                    extIdObj.getSubstudyId(), account.getId());
-            // If a substudy relationship exists without the external ID, remove it because
-            // we're about to create it with an external ID
-            if (account.getAccountSubstudies().contains(acctSubstudy)) {
-                account.getAccountSubstudies().remove(acctSubstudy);
-            }
-            acctSubstudy.setExternalId(extIdObj.getIdentifier());
-            account.getAccountSubstudies().add(acctSubstudy);
-            try {
-                accountDao.updateAccount(account,
-                        (modifiedAccount) -> externalIdService.commitAssignExternalId(extIdObj));
-            } catch(Exception e) {
-                if (extIdObj != null) {
-                    externalIdService.unassignExternalId(account, extIdObj.getIdentifier());    
-                }
-                throw e;
-            }
-        }
-        return extIdObj;
-    }
-
     /**
      * This is a researcher API to backfill SMS notification registrations for a user. We generally prefer the app
      * register notifications, but sometimes the work can't be done on time, so we want study developers to have the
@@ -856,6 +830,9 @@ public class ParticipantService {
         return activityEventService.getActivityEventList(account.getHealthCode());
     }
     
+    /**
+     * This method is only executed on the authenticated caller, not on behalf of any other person.
+     */
     public StudyParticipant updateIdentifiers(Study study, CriteriaContext context, IdentifierUpdate update) {
         checkNotNull(study);
         checkNotNull(context);
@@ -878,9 +855,7 @@ public class ParticipantService {
         }
         
         // reload account, or you will get an optimistic lock exception
-        AccountId accountId = AccountId.forId(study.getIdentifier(), account.getId());
         boolean sendEmailVerification = false;
-        boolean assignExternalId = false;
         boolean accountUpdated = false;
         if (update.getPhoneUpdate() != null && account.getPhone() == null) {
             account.setPhone(update.getPhoneUpdate());
@@ -893,18 +868,26 @@ public class ParticipantService {
             sendEmailVerification = true;
             accountUpdated = true;
         }
-        Set<String> externalIds = BridgeUtils.collectExternalIds(account);
-        if (update.getExternalIdUpdate() != null && !externalIds.contains(update.getExternalIdUpdate())) {
-            assignExternalId = true;
-            accountUpdated = true;
-        }
-        if (accountUpdated) {
-            if (assignExternalId) {
-                ExternalIdentifier externalId = assignExternalId(accountId, update.getExternalIdUpdate());
-                updateRequestContext(externalId);
-            } else {
-                accountDao.updateAccount(account, null);
+        ExternalIdentifier externalId = beginAssignExternalId(account, update.getExternalIdUpdate());
+        if (externalId != null) {
+            AccountSubstudy acctSubstudy = AccountSubstudy.create(account.getStudyId(),
+                    externalId.getSubstudyId(), account.getId());
+            // Highly unlikely this was an admin account, but just in case
+            if (account.getAccountSubstudies().contains(acctSubstudy)) {
+                account.getAccountSubstudies().remove(acctSubstudy);
             }
+            acctSubstudy.setExternalId(externalId.getIdentifier());
+            account.getAccountSubstudies().add(acctSubstudy);
+            try {
+                accountDao.updateAccount(account,
+                        (modifiedAccount) -> externalIdService.commitAssignExternalId(externalId));
+            } catch(Exception e) {
+                externalIdService.unassignExternalId(account, externalId.getIdentifier());    
+                throw e;
+            }
+            updateRequestContext(externalId);
+        } else if (accountUpdated) {
+            accountDao.updateAccount(account, null);
         }
         if (sendEmailVerification && 
             study.isEmailVerificationEnabled() && 
--- a/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -537,9 +537,7 @@ public class ParticipantService {
             accountDao.updateAccount(account,
                     (modifiedAccount) -> externalIdService.commitAssignExternalId(externalId));
         } catch (Exception e) {
-            if (externalId != null) {
-                externalIdService.unassignExternalId(account, externalId.getIdentifier());
-            }
+            externalIdService.unassignExternalId(account, externalId.getIdentifier());
             throw e;
         }
     }
--- a/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -3,13 +3,12 @@ package org.sagebionetworks.bridge.services;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.Boolean.TRUE;
-import static java.util.stream.Collectors.toSet;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.sagebionetworks.bridge.BridgeUtils.collectSubstudyIds;
 import static org.sagebionetworks.bridge.BridgeUtils.substudyAssociationsVisibleToCaller;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.ADMINISTRATIVE_ROLES;
 import static org.sagebionetworks.bridge.Roles.CAN_BE_EDITED_BY;
-import static org.sagebionetworks.bridge.Roles.RESEARCHER;
 import static org.sagebionetworks.bridge.Roles.WORKER;
 
 import java.security.InvalidKeyException;
@@ -18,7 +17,6 @@ import java.security.spec.InvalidKeySpecException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 import com.google.common.collect.ImmutableSet;
@@ -41,6 +39,7 @@ import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.dao.ScheduledActivityDao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
+import org.sagebionetworks.bridge.exceptions.ConstraintViolationException;
 import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.LimitExceededException;
@@ -191,12 +190,32 @@ public class ParticipantService {
         this.requestInfoService = requestInfoService;
     }
     
-    public void assignExternalId(AccountId accountId, ExternalIdentifier externalId) {
+    public ExternalIdentifier assignExternalId(AccountId accountId, String externalId) {
         Account account = getAccountThrowingException(accountId);
-        ExternalIdentifier extIdObj = beginAssignExternalId(account, externalId.getIdentifier());
-        externalIdService.commitAssignExternalId(extIdObj);
+        ExternalIdentifier extIdObj = beginAssignExternalId(account, externalId);
+        if (extIdObj != null) {
+            AccountSubstudy acctSubstudy = AccountSubstudy.create(account.getStudyId(),
+                    extIdObj.getSubstudyId(), account.getId());
+            // If a substudy relationship exists without the external ID, remove it because
+            // we're about to create it with an external ID
+            if (account.getAccountSubstudies().contains(acctSubstudy)) {
+                account.getAccountSubstudies().remove(acctSubstudy);
+            }
+            acctSubstudy.setExternalId(extIdObj.getIdentifier());
+            account.getAccountSubstudies().add(acctSubstudy);
+            try {
+                accountDao.updateAccount(account,
+                        (modifiedAccount) -> externalIdService.commitAssignExternalId(extIdObj));
+            } catch(Exception e) {
+                if (extIdObj != null) {
+                    externalIdService.unassignExternalId(account, extIdObj.getIdentifier());    
+                }
+                throw e;
+            }
+        }
+        return extIdObj;
     }
-    
+
     /**
      * This is a researcher API to backfill SMS notification registrations for a user. We generally prefer the app
      * register notifications, but sometimes the work can't be done on time, so we want study developers to have the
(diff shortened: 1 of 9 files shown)


src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -450,7 +453,7 @@ public void createParticipantWithExternalIdAndSubstudyCallerValidates() {
         participantService.createParticipant(STUDY, participant, false);
     }
     
-    @Test(expectedExceptions = InvalidEntityException.class, 
+    @Test(enabled = false, expectedExceptions = InvalidEntityException.class,
```
DwayneJengSage: What are we going to do with all these disabled and commented-out tests?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -191,12 +190,32 @@ final void setRequestInfoService(RequestInfoService requestInfoService) {
         this.requestInfoService = requestInfoService;
     }
     
-    public void assignExternalId(AccountId accountId, ExternalIdentifier externalId) {
+    public ExternalIdentifier assignExternalId(AccountId accountId, String externalId) {
         Account account = getAccountThrowingException(accountId);
-        ExternalIdentifier extIdObj = beginAssignExternalId(account, externalId.getIdentifier());
-        externalIdService.commitAssignExternalId(extIdObj);
+        ExternalIdentifier extIdObj = beginAssignExternalId(account, externalId);
+        if (extIdObj != null) {
```
DwayneJengSage: What does it mean if beginAssignExternalId() returns null?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -191,12 +190,32 @@ final void setRequestInfoService(RequestInfoService requestInfoService) {
...
+            AccountSubstudy acctSubstudy = AccountSubstudy.create(account.getStudyId(),
+                    extIdObj.getSubstudyId(), account.getId());
+            // If a substudy relationship exists without the external ID, remove it because
+            // we're about to create it with an external ID
+            if (account.getAccountSubstudies().contains(acctSubstudy)) {
+                account.getAccountSubstudies().remove(acctSubstudy);
+            }
```
DwayneJengSage: What happens if the participant is already assigned to the substudy with a different external ID?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -191,12 +190,32 @@ final void setRequestInfoService(RequestInfoService requestInfoService) {
         this.requestInfoService = requestInfoService;
     }
     
-    public void assignExternalId(AccountId accountId, ExternalIdentifier externalId) {
+    public ExternalIdentifier assignExternalId(AccountId accountId, String externalId) {
```
DwayneJengSage: I'm unable to find unit tests for assignExternalId(). Need the following test cases:

* null externalId in input
* externalId not in externalIdService
* externalId already assigned to the participant
* externalId assigned to a different participant
* participant already in the substudy with a different externalId
* error assigning externalId causes externalId to be unassigned
* participant without a substudy assigns an externalId in a substudy
* participant with a substudy assigns an externalId in the same substudy
* participant with a substudy assigns an externalId in a different substudy

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -412,20 +430,6 @@ public IdentifierHolder createParticipant(Study study, StudyParticipant particip
...
-        Optional<ExternalIdentifier> optionalId = externalIdService.getExternalId(
-                study.getStudyIdentifier(), participant.getExternalId());
-        if (optionalId.isPresent()) {
-            Set<String> substudyIds = new ImmutableSet.Builder<String>()
-                    .addAll(participant.getSubstudyIds())
-                    .add(optionalId.get().getSubstudyId()).build();
-            participant = new StudyParticipant.Builder()
-                    .copyOf(participant)
-                    .withSubstudyIds(substudyIds).build();
-        }
```
DwayneJengSage, line 427: Why is this code being deleted?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -914,31 +920,27 @@ protected ExternalIdentifier beginAssignExternalId(Account account, String exter
...
         }
-        ExternalIdentifier identifier = optionalId.get();
         if (identifier.getHealthCode() != null && !account.getHealthCode().equals(identifier.getHealthCode())) {
             throw new EntityAlreadyExistsException(ExternalIdentifier.class, "identifier", identifier.getIdentifier()); 
         }
-        // Whether already assigned or not, we will adjust the account, in case we are repairing
-        // an existing broken data association
-
-        identifier.setHealthCode(account.getHealthCode());
-        if (identifier.getSubstudyId() != null) {
```
DwayneJengSage, line 934: Is it possible for an externalId to not be assigned to a substudy?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -191,12 +190,32 @@ final void setRequestInfoService(RequestInfoService requestInfoService) {
...
+            }
+            acctSubstudy.setExternalId(extIdObj.getIdentifier());
+            account.getAccountSubstudies().add(acctSubstudy);
+            try {
+                accountDao.updateAccount(account,
+                        (modifiedAccount) -> externalIdService.commitAssignExternalId(extIdObj));
+            } catch(Exception e) {
+                if (extIdObj != null) {
+                    externalIdService.unassignExternalId(account, extIdObj.getIdentifier());    
+                }
```
DwayneJengSage: Is this null-check necessary? How does extIdObj become null in this codepath?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -532,33 +536,21 @@ public void updateParticipant(Study study, StudyParticipant participant) {
...
-        // Legacy behavior: a user can add an external ID to their account on an update, we refer to 
-        // this as a "simple add." A researcher can assign an external ID to any user if it has not yet 
-        // been assigned (if it has been assigned, beginAssignExternalId() throws an exception and 
-        // aborts this entire call).
-        
-        boolean isSimpleAdd = allExternalIds.isEmpty() && participant.getExternalId() != null;
-        boolean isResearcherAdd = callerRoles.contains(RESEARCHER)
-                && !allExternalIds.contains(participant.getExternalId())
-                && participant.getExternalId() != null;
-        boolean assigningExternalId = isSimpleAdd || isResearcherAdd;
```
DwayneJengSage, line 549: Why is this code being deleted?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -597,51 +589,57 @@ private void updateAccountAndRoles(Study study, Account account, StudyParticipan
...
-                    .map(AccountSubstudy::getSubstudyId).collect(toSet());
+            // Add substudy relationship
+            Set<String> existingSubstudyIds = collectSubstudyIds(account);
             for (String substudyId : participant.getSubstudyIds()) {
-                // Edge case: when creating user if the external ID and the substudy ID it is associated to are 
-                // both set, the persist fails. To work correctly, we want to ignore the substudy ID assignment 
-                // because it's going to be established by the external ID assignment.
-                if (externalId != null && externalId.getSubstudyId().equals(substudyId)) {
-                    continue;
-                }
```
DwayneJengSage, line 629: Why is this code being deleted?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -597,51 +589,57 @@ private void updateAccountAndRoles(Study study, Account account, StudyParticipan
...
-            // We have to clear the cache if we make changes that can alter the security profile of 
-            // the account, otherwise very strange behavior can occur if that user is signed in with 
-            // a stale session.
-            if (!isNew && clearCache) {
-                cacheProvider.removeSessionByUserId(account.getId());    
+            // If a substudy relationship exists without the external ID, remove it because
+            // we're about to create it with an external ID
+            if (account.getAccountSubstudies().contains(acctSubstudy)) {
+                account.getAccountSubstudies().remove(acctSubstudy);
             }
```
DwayneJengSage, line 603: What happens if the participant is already assigned to the substudy with a different externalId? Do we have unit tests for this case?

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -2902,6 +2891,478 @@ public void getActivityHistory() {
...
+    @Test
+    public void substudyResearcherCannotAddSubstudyIdOnUpdate() {
+        BridgeUtils.setRequestContext(new RequestContext.Builder()
+                .withCallerSubstudies(ImmutableSet.of(SUBSTUDY_ID, "secondSubstudyId"))
+                .withCallerRoles(ImmutableSet.of(RESEARCHER)).build());
+        when(substudyService.getSubstudy(TEST_STUDY, SUBSTUDY_ID, false)).thenReturn(Substudy.create());
+        when(substudyService.getSubstudy(TEST_STUDY, "secondSubstudyId", false)).thenReturn(Substudy.create());
+        account.setId(ID);
+        account.getAccountSubstudies().add(AccountSubstudy.create(TEST_STUDY_IDENTIFIER, SUBSTUDY_ID, ID));
+        when(accountDao.getAccount(ACCOUNT_ID)).thenReturn(account);
```
DwayneJengSage, line 3119: Unit test for substudy researcher assigning an externalId to a participant not in any substudy.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -862,9 +878,7 @@ public StudyParticipant updateIdentifiers(Study study, CriteriaContext context,
         }
         
         // reload account, or you will get an optimistic lock exception
-        account = getAccountThrowingExceptionIfSubstudyMatches(AccountId.forId(study.getIdentifier(), account.getId()));
```
DwayneJengSage, line 865: Is this no longer needed?

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -2165,8 +2169,11 @@ public void updatingBlankExternalIdFails() {
         verify(externalIdService, never()).commitAssignExternalId(any());
     }
 
-    @Test
-    public void changingManagedExternalIdWorks() {
+    // To change the ID associated to a substudy, you'd have to be an researcher/admin and do it in two 
+    // steps, unassigning the first ID and then assigning the second ID. Non-admin users cannot do this, 
+    // they can only add external IDs.
+    @Test(expectedExceptions = ConstraintViolationException.class)
+    public void changingExternalIdDoesNotWork() throws Exception {
```
DwayneJengSage, line 2158: Is this test redundant with updateParticipantExternalIdsExistOneAddedFails() and normalUserCannotChangeExternalIdOnUpdate()?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Ack. These need to be re-enabled and adjusted to pass. I have done this now.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: If beginAssignExternalId returns an externalIdentifier object, then assign the external identifier. If the precondition checks of beginAssignExternalId don't pass (some are silent and some throw exceptions), then don't assign the external identifier.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Nothing in the code, but the database will throw an error. I think we want this if people try and change the external ID because of the chaos that results. Admins can delete the first external ID, then add another one, if it really needs to be done. The code here only makes it possible to add an external ID that would "update" an existing relationship to a substudy by adding the external ID to it.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: All good tests... added them.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: This was a hack so the validator would see the account substudy records in their final state, since we were validating before updating that collection. I took these substudy security checks out of the validator and put them after the update of accountSubstudies, making this hack adjustment unnecessary.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Not any longer, no, they must always be assigned to a substudy. external = somebody besides us came up with it.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: You're right, it's impossible, removing

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: These are the rules we're "simplifying" as described. We're allowing adds all the time, as long as they come from assignment of an external ID and don't change an existing external ID/substudy relationship. Admins can directly manipulate substudy IDs and external IDs after an account is created to fix any problems that come up.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Because the new record being added by the assignment of an external ID now happens later in the method after we check substudy collection changes by an admin. So, we no longer need these special exceptions in the iteration to compare the persisted and submitted sets of substudy IDs. First we make substudy ID changes if the user is an admin, then we add an external ID if one is being assigned.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: This code doesn't execute because the external IDs will make the two associative records different, but the database will reject two records assigning to the same substudy. 

rollbackCreateParticipantWhenAccountCreationFails examines this case for create participant (the test throws a ConcurrentModificationException which might actually be the exception you get in this case, but the code handles any persistence exception the same way), and rollbackUpdateParticipantWhenAccountUpdateFails does the same for updateParticipant.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: If the user is not in the substudy researcher's substudy, then they cannot see it to assign to it (the user would have to assign to self, or an admin or global researcher would have to do it).

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Yes now that we've moved the reauth token rotation to a different table, we no longer need to reload the account like this.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: changingExternalIdDoesNotWork and updateParticipantExternalIdsExistOneAddedFails are the same. normalUserCannotChangeExternalIdOnUpdate differs only slightly in that the caller is not in the substudy of the external IDs being changed. I deleted updateParticipantExternalIdsExistOneAddedFails.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1803,40 +1771,44 @@ public void throwLimitExceededException() {
...
-            participantService.updateIdentifiers(STUDY, CONTEXT, update);
-            fail("Should have thrown exception");
-        } catch(EntityNotFoundException e) {
-            assertEquals(e.getMessage(), "Account not found.");
-        }
-        verify(accountDao, never()).updateAccount(any(), any());
+        IdentifierUpdate update = new IdentifierUpdate(EMAIL_PASSWORD_SIGN_IN, null, null, "newExtId");
+        participantService.updateIdentifiers(STUDY, CONTEXT, update);
+        
+        verify(accountDao).updateAccount(accountCaptor.capture(), any());
```
DwayneJengSage, line 1795: We should also verify that we call externalIdService.commitAssignExternalId()

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1803,40 +1771,44 @@ public void throwLimitExceededException() {
...
     }
     
     @Test
     public void updateIdentifiersAssignsExternalIdEvenWhenAlreadyAssigned() {
         // Fully associated external ID can be changed by an update.
+        BridgeUtils.setRequestContext(new RequestContext.Builder()
+                .withCallerSubstudies(ImmutableSet.of("substudyB")).build());
         mockHealthCodeAndAccountRetrieval();
-        AccountSubstudy as = AccountSubstudy.create(STUDY.getIdentifier(), "substudyB", ID);
+        AccountSubstudy as = AccountSubstudy.create(TEST_STUDY_IDENTIFIER, "substudyB", ID);
```
DwayneJengSage, line 1809: Do we also have a test that updateIdentifiers will ignore if we attempt to assign the same externalId to the same account?

DwayneJengSage (changes requested):
A few comments about updateIdentifiers tests. Otherwise looks good.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
alxdarksage: Added updateIdentifiersDoesNothingWhenExternalIdResubmitted() and renamed updateIdentifiersDoesNotReassignExistingExternalId() to updateIdentifiersDoesNotReassignExternalIdOnOtherUpdate() to slightly clarify the difference between the two tests.

DwayneJengSage (approved):

PR #99, opened 2019-08-28
Export user languages to Synapse

alxdarksage:
Including the user's languages as part of the health data record.


DwayneJengSage:
Pretty sure we decided to punt on this until Bridge Exporter 3.0.

PR #105, opened 2019-09-05
Better verification messaging

alxdarksage:
We now keep the token around so we can reidentify the account and return a more helpful message when we can conclude the token was already used to verify the email address/phone number.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.services;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.Boolean.TRUE;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.sagebionetworks.bridge.models.templates.TemplateType.EMAIL_ACCOUNT_EXISTS;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -208,8 +217,9 @@ public void sendEmailVerificationToken(Study study, String userId, String recipi
         }
 
         String sptoken = getNextToken();
+        long expiresOn = getDateTimeInMillis() + (VERIFY_OR_RESET_EXPIRE_IN_SECONDS*1000);
 
-        saveVerification(sptoken, new VerificationData(study.getIdentifier(), ChannelType.EMAIL, userId));
+        saveVerification(sptoken, new VerificationData(study.getIdentifier(), ChannelType.EMAIL, userId, expiresOn));
```
DwayneJengSage, line 222: Do we have any unit tests that verify that we're saving the expiresOn to the cacheProvider correctly? (For both email and phone case.)

DwayneJengSage (changes requested):
Overall looks good. One question about unit tests and the VerificationData.

DwayneJengSage (changes requested):
I think there was a misunderstanding. I was asking for a test that verifies that we set expiresOn correctly when creating VerificationData.

alxdarksage:
Updated based on your latest comment (checking the timestamp that's being persisted in VerificationData).

DwayneJengSage (approved):

PR #106, opened 2019-09-09
Context refactor

alxdarksage:
Move the extraction of IP address from the HTTP request from the CriteriaContext to the RequestContext. It's about the request and not needed for criteria filtering.

This is probably the only cleanup I'm going to make for BRIDGE-2536.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
@@ -229,7 +229,7 @@ public class AuthenticationServiceMockTest {
         account.setAccountSubstudies(ImmutableSet.of(as1, as2));
         account.setId(USER_ID);
         
-        BridgeUtils.setRequestContext(new RequestContext.Builder().withCallerIpAddress(IP_ADDRESS).build());
+        setIpAddress(IP_ADDRESS);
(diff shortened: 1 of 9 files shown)


src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -209,7 +210,11 @@ public void before() {
     
     @AfterMethod
     public void after() {
-        BridgeUtils.setRequestContext(RequestContext.NULL_INSTANCE);
+        BridgeUtils.setRequestContext(NULL_INSTANCE);
+    }
+    
+    void setIpAddress(String ipAddress) {
+        BridgeUtils.setRequestContext(new RequestContext.Builder().withCallerIpAddress(ipAddress).build());
```
DwayneJengSage, line 217: Minor: We create this nifty little helper method, but never use it anywhere.

DwayneJengSage (approved):
One minor comment. Looks good to check in regardless.

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: We now use it

PR #107, opened 2019-09-11
Study admin

alxdarksage:
BRIDGE-2561

Allow the creation of admins within a study that can do all the normal admin things... for that study only. There are some protections that need to be added throughout the API:

1) API calls that operate on all studies (create a study, delete a study, list cache keys, etc.) need to verify the admin is a "full admin" (account is defined in the API study);

2) API calls that allow an admin, need to read the study being used from the caller's session. Study admins can only authenticate to that study and full admins can put themselves in any study, so this is checked by using the session study. We have a very few calls that do not look at this study... in those cases, we need to add a check in the controller to prevent someone authenticating as an admin in study A, doing something like retrieving upload records from study B.

3) FPHS is an odd edge case there are one or two others, that are handled as well.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
@@ -131,11 +131,10 @@ public class StudyController extends BaseController {
     public Study getStudy(@PathVariable String identifier) throws Exception {
         UserSession session = getAuthenticatedSession(ADMIN, WORKER);
         
-        // Caller must have permissions to operate on this study. However worker may be used
+        // Caller must be a full admin to get any study, however worker may be used
         // cross study, so exclude it from this check
-        if (!session.isInRole(WORKER) &&
-            !session.getStudyIdentifier().getIdentifier().equals(identifier)) {
-            throw new UnauthorizedException("Study admin cannot retrieve other studies.");
+        if (!session.isInRole(WORKER)) {
+            verifyCrossStudyAdmin(session.getId(), "Study admin cannot retrieve other studies.");
         }
         // since only admin and worker can call this method, we need to return all studies including deactivated ones
         return studyService.getStudy(identifier, true);
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
@@ -93,6 +93,11 @@ public class StudyController extends BaseController {
     final void setUploadService(UploadService uploadService) {
         this.uploadService = uploadService;
     }
+    
+    // To enable mocking of values.
+    Set<String> getStudyWhitelist() {
+        return studyWhitelist;
+    }
 
     @GetMapping("/v3/studies/self")
     public Study getCurrentStudy() throws Exception {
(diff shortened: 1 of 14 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -114,8 +114,13 @@ public VersionHolder updateStudyForDeveloper() throws Exception {
...
     @PostMapping("/v3/studies/{identifier}")
     public VersionHolder updateStudy(@PathVariable String identifier) throws Exception {
-        getAuthenticatedSession(ADMIN);
-
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        // ADMIN must have permissions to operate on this study.
+        if (!session.getStudyIdentifier().getIdentifier().equals(identifier)) {
+            throw new UnauthorizedException("Study admin cannot update another study");
+        }
```
DwayneJengSage, line 127: It looks like there are no positive cases for any of these API calls. Should we add those to StudyControllerTest?

src/test/java/org/sagebionetworks/bridge/spring/controllers/CacheAdminControllerTest.java
```diff
@@ -60,6 +75,21 @@ public void verifyAnnotations() throws Exception {
...
+        
+        // This should be a ResourceList, but it's not currently, so we're maintaining that.
+        Set<String> cacheItems = controller.listItems();
+        assertEquals(items, cacheItems);
+        
+        verify(mockCacheAdminService).listItems();
+    }
+    
+    @Test(expectedExceptions = UnauthorizedException.class)
+    public void listItemsRejectsStudyAdmin() throws Exception {
```
DwayneJengSage, line 90: Minor: This test has verifies after the call to controller.listItems() tha twill never be called. (Same with removeItemRejectsStudyAdmin().)

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadController.java
```diff
@@ -175,6 +175,10 @@ public UploadView getUpload(@PathVariable String uploadId) {
             if (record == null) {
                 throw new EntityNotFoundException(HealthDataRecord.class);
             }
+            // You cannot authenticate in one study, and then retrieve a record from another study
+            if (!session.getStudyIdentifier().getIdentifier().equals(record.getStudyId())) {
+                throw new UnauthorizedException("Study admin cannot retrieve upload in another study.");
+            }
```
DwayneJengSage, line 188: It's very common for me to have an upload ID but not know which study it belongs to. I think it would be better to use verifyCrossAdminStudy() instead.

Also, the Worker also calls this API, and this would kill the Worker.

src/test/java/org/sagebionetworks/bridge/spring/controllers/UploadControllerTest.java
```diff
@@ -321,11 +322,13 @@ public void getUploadById() throws Exception {
         assertEquals(node.get("healthData").get("healthCode").textValue(), HEALTH_CODE);
     }
 
-    @Test
-    public void getUploadByRecordId() throws Exception {
+    @Test(expectedExceptions = UnauthorizedException.class)
+    public void getUploadByRecordIdRejectsStudyAdmin() throws Exception {
```
DwayneJengSage: Shouldn't there be a positive test for this as well?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
alxdarksage: Added a positive case alongside all these study admin tests.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CacheAdminControllerTest.java
alxdarksage: Removed in both places.

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadController.java
alxdarksage: It should be an either/or thing I think (study admins can find uploads in their study, full admins in any study). Updated with an additional test.

src/test/java/org/sagebionetworks/bridge/spring/controllers/UploadControllerTest.java
alxdarksage: Yes, added this back.

DwayneJengSage (approved):

PR #108, opened 2019-09-12
Master Scheduler Model/API

janaechan:
Master Scheduler Design Doc: https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/436862979/Master+Scheduler


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerConfigDao.java
```diff
@@ -0,0 +1,123 @@
...
+        this.mapper = mapper;
+    }
+
+    @Override
+    public List<MasterSchedulerConfig> getAllSchedulerConfig() {
+        DynamoMasterSchedulerConfig configKey = new DynamoMasterSchedulerConfig();
+        // Mapper.scan
+        DynamoDBQueryExpression<DynamoMasterSchedulerConfig> query = new DynamoDBQueryExpression<DynamoMasterSchedulerConfig>()
+                .withHashKeyValues(configKey);
+        PaginatedQueryList<DynamoMasterSchedulerConfig> results = mapper.query(DynamoMasterSchedulerConfig.class, query);
```
DwayneJengSage: You want mapper.scan, not mapper.query.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerConfigDao.java
```diff
@@ -0,0 +1,123 @@
...
+        
+        List<MasterSchedulerConfig> list = Lists.newArrayListWithCapacity(results.size());
+        for (DynamoMasterSchedulerConfig config : results) {
+            list.add(config);
+        }
+        return null;
```
DwayneJengSage: Return the list, not null. (If the list is empty, return the empty list.)

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDao.java
```diff
@@ -0,0 +1,28 @@
...
+    @Resource(name = "masterSchedulerStatusMapper")
+    final void setMasterSchedulerStatusMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    // getLastProcessedTime
+    @Override
+    public Long getLastProcessedTime() {
+        DynamoMasterSchedulerStatus statusKey = new DynamoMasterSchedulerStatus();
+        statusKey.setHashKey("BridgeMasterSheduler");
```
DwayneJengSage: By convention, we sould create a static final String (constant) for BridgeMasterScheduler. That way, if we use it in more than once place, we can refer to the constant.

Also, uh, you mispelled Scheduler.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDao.java
```diff
@@ -0,0 +1,28 @@
...
+        
+        return mapper.load(statusKey).getLastProcessedTime();
```
DwayneJengSage: We should have a null check, in case the lastProcessedTime in your environment hasn't been initialized. (In this case, I think returning null makes the most sense.)

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
```diff
@@ -0,0 +1,71 @@
...
+        return schedulerConfigDao.createSchedulerConfig(schedulerConfig);
+    }
+
+    public MasterSchedulerConfig getSchedulerConfig(String scheduleId) {
+        checkNotNull(scheduleId);
+        
+        MasterSchedulerConfig schedulerConfig = schedulerConfigDao.getSchedulerConfig(scheduleId);
+        if (schedulerConfig == null) {
+            throw new BadRequestException("Can't get scheduler config for scheduleId=" + scheduleId 
+                    + ": scheduler does not exists");
```
DwayneJengSage: This should be EntityNotFoundException. Also, between the DAO and the Service, only one of them needs null checks. I feel like having the null check in the Service makes sense, as that's where the validation happens.

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        
+        return schedulerConfig;
+    }
+
+    public MasterSchedulerConfig updateSchedulerConfig(String scheduleId, MasterSchedulerConfig schedulerConfig) {
+        checkNotNull(scheduleId);
+        checkNotNull(schedulerConfig);
+        
+        Validate.entityThrowingException(MasterSchedulerConfigValidator.INSTANCE, schedulerConfig);
```
DwayneJengSage, line 60: We should also set the scheduleId to the schedulerConfig, to ensure we don't accidentally create a separate schedulerConfig.

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
```diff
@@ -0,0 +1,71 @@
...
+import org.sagebionetworks.bridge.dao.MasterSchedulerStatusDao;
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.models.schedules.MasterSchedulerConfig;
+import org.sagebionetworks.bridge.validators.MasterSchedulerConfigValidator;
+import org.sagebionetworks.bridge.validators.Validate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MasterSchedulerService {
```
DwayneJengSage, line 17: Should we also have a call to getLastProcessedTime?

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,85 @@
...
+    @Autowired
+    MasterSchedulerService schedulerService;
+
+    public void setMasterSchedulerService(MasterSchedulerService schedulerService) {
+        this.schedulerService = schedulerService;
+    }
+
+    @GetMapping("/v3/schedulerconfigs")
+    public ResourceList<MasterSchedulerConfig> getAllSchedulerConfigs() {
+        getAuthenticatedSession(Roles.ADMIN);
```
DwayneJengSage: When you merge in Alx's pending changes, you'll need to include verifyCrossStudyAdmin() here and in all the other methods.

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @PostMapping("/v3/schedulerconfigs")
+    @ResponseStatus(HttpStatus.CREATED)
+    public StatusMessage createSchedulerConfig() {
+        getAuthenticatedSession(Roles.ADMIN);
+
+        MasterSchedulerConfig schedulerConfig = parseJson(MasterSchedulerConfig.class);
+        
+        schedulerService.createSchedulerConfig(schedulerConfig);
+        return CREATED_MSG;
```
DwayneJengSage: I think we decided to return the created object in the create and update methods?

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+@CrossOrigin
+@RestController
+public class MasterSchedulerController extends BaseController {
```
DwayneJengSage, line 24: Similarly, a call to getLastProcessedTime.

DwayneJengSage (changes requested):
Also unit tests. Also, JavaSDK and IntegrationTests.

Also, wasn't there a DefaultStudyBootstrapper fix we were including?

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
janaechan: Should I create a holder for the time in the sdk?

DwayneJengSage: Yes. Did we end up creating that generic timestamp holder?

janaechan: No, it was a study_start_date holder (but ended up removing it). Should the holder be specific to this lastProcessedTime or a generic timestamp?

DwayneJengSage: I think a generic timestamp holder would be better.

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
janaechan: I noticed in that in other DAOs the delete() method has the null check and doesn't throw any exception if it is null. Should I handle it similarly in the scheduler?

DwayneJengSage: I think in this case, it should throw an error, since this is used as an admin API, and having some feedback when something unexpected happens would be useful.

src/main/java/org/sagebionetworks/bridge/services/MasterSchedulerService.java
```diff
@@ -0,0 +1,78 @@
...
+            throw new EntityNotFoundException(MasterSchedulerConfig.class);
+        }
+        
+        return schedulerConfigDao.updateSchedulerConfig(schedulerConfig);
+    }
+
+    public void deleteSchedulerConfig(String scheduleId) {
+        checkNotNull(scheduleId);
+        
+        schedulerConfigDao.deleteSchedulerConfig(scheduleId);
```
DwayneJengSage, line 77: Let's get the schedule config and throw an EntityNotFoundException if it doesn't exist.

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,96 @@
...
+    @Autowired
+    MasterSchedulerService schedulerService;
+
+    public void setMasterSchedulerService(MasterSchedulerService schedulerService) {
+        this.schedulerService = schedulerService;
+    }
+
+    @GetMapping("/v3/schedulerconfigs")
+    public ResourceList<MasterSchedulerConfig> getAllSchedulerConfigs() {
+        getAuthenticatedSession(ADMIN);
```
DwayneJengSage: You'll need to merge in Alx's changes and call verifyCrossStudyAdmin() here and in all the other methods.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDaoTest.java
```diff
@@ -0,0 +1,5 @@
+package org.sagebionetworks.bridge.dynamodb;
+
+public class DynamoMasterSchedulerStatusDaoTest {
+
+}
```
DwayneJengSage, line 55: TODO: Fill this in.

src/test/java/org/sagebionetworks/bridge/models/TimestampHolderTest.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final long TIME_STAMP = DateTime.parse("2018-03-27T18:30-07:00").getMillis();
+    
+    @Test
+    public void canSerialize() throws Exception {
+        TimestampHolder holder = new TimestampHolder(TIME_STAMP);
+        
+        JsonNode node = BridgeObjectMapper.get().valueToTree(holder);
+        
+        assertEquals(TIME_STAMP, node.get("timestamp").asLong());
+        assertEquals("TimestampHolder", node.get("type").textValue());
```
DwayneJengSage: Do we want to serialize this as an IS8601 timestamp instead of long epoch milliseconds? See DynamoHealthDataRecord.getCreatedOn() for an example of serializing a long into an ISO timestamp.

src/test/java/org/sagebionetworks/bridge/models/schedules/MasterSchedulerConfigTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.models.TimestampHolder;
+import org.testng.annotations.Test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class MasterSchedulerConfigTest {
```
DwayneJengSage: You don't need a separate test for MasterSchedulerConfigTest when it's identical to DynamoMasteSchedulerConfigTest. (Same with MasterSchedulerStatusTest.)

src/test/java/org/sagebionetworks/bridge/services/MasterSchedulerServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+        assertEquals(config.getRequestTemplate().get("b").textValue(), "string");
+        assertEquals(config.getSqsQueueUrl(), SQS_QUEUE_URL);
+        assertEquals(config.getVersion(), new Long(VERSION));
+        assertSame(mockConfig, config);
+        verify(configDao).getSchedulerConfig(SCHEDULE_ID);
+    }
+    
+    @Test(expectedExceptions = EntityNotFoundException.class)
+    public void testGetSchedulerConfigThatDoesNotExist() {
+        when(configDao.getSchedulerConfig(SCHEDULE_ID)).thenReturn(any());
```
DwayneJengSage: Minor: any() is for matching arguments. If you want to return null, thenReturn(null) will do.

src/test/java/org/sagebionetworks/bridge/services/MasterSchedulerServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+    public void testUpdateSchedulerConfig() {
+        MasterSchedulerConfig mockConfig = TestUtils.getMasterSchedulerConfig();
+        when(configDao.getSchedulerConfig(SCHEDULE_ID)).thenReturn(mockConfig);
+        
+        ObjectNode objNode = JsonNodeFactory.instance.objectNode();
+        objNode.put("a", true);
+        objNode.put("b", "string2");
+        
+        MasterSchedulerConfig updatedConfig = MasterSchedulerConfig.create();
+        updatedConfig.setScheduleId(SCHEDULE_ID);
```
DwayneJengSage: We should set a different schedule ID here to make sure that the update() method sets schedule ID correctly.

src/test/java/org/sagebionetworks/bridge/services/MasterSchedulerServiceTest.java
```diff
@@ -0,0 +1,202 @@
...
+        verify(configDao).createSchedulerConfig(configCaptor.capture());
+    }
+    
+    @Test(expectedExceptions = InvalidEntityException.class)
+    public void testCreateSchedulerConfigValidates() {
+        MasterSchedulerConfig mockConfig = TestUtils.getMasterSchedulerConfig();
+        mockConfig.setScheduleId(null);
+        
+        schedulerService.createSchedulerConfig(mockConfig);
+        verify(configDao, never()).createSchedulerConfig(any());
```
DwayneJengSage, line 110: Note: If createSchedulerConfig() throws, the verify will never get called. If you want to verify that the dao is never called, the best thing to do is to wrap it in a try catch:
```
try {
  schedulerService.createSchedulerConfig(mockConfig);
  fail("expected exception");
} catch (InvalidEntityException ex) {
  // expected exception
}
verify(configDao, never()).createSchedulerConfig(any());
```

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@CrossOrigin
+@RestController
+public class MasterSchedulerController extends BaseController {
+    static final StatusMessage DELETED_MSG = new StatusMessage("Scheduler config deleted.");
+
+    
+    @Autowired
+    MasterSchedulerService schedulerService;
```
DwayneJengSage: By convention, this should be private, and the Autowired annotation should be on the setter.

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void setMasterSchedulerService(MasterSchedulerService schedulerService) {
```
DwayneJengSage: We usually mark these setters as final. It's unlikely to come up again, but if you subclass a class and override the setter, the parent class's setter will never be called, and then bad things happen.

src/test/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerControllerTest.java
```diff
@@ -0,0 +1,242 @@
...
+    
+    Study study;
+    
+    @BeforeMethod
+    private void before() {
+        MockitoAnnotations.initMocks(this);
+        
+        mockConfig = TestUtils.getMasterSchedulerConfig();
+        
+        study = new DynamoStudy();
```
DwayneJengSage: We should use Study.create(), so the Dynamo details don't leak outside the Dynamo classes.

src/test/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerControllerTest.java
```diff
@@ -0,0 +1,242 @@
...
+        session.setParticipant(new StudyParticipant.Builder().withRoles(ImmutableSet.of(ADMIN)).build());
+        List<MasterSchedulerConfig> configs = ImmutableList.of(mockConfig, mockConfig, mockConfig);
+        
+        // Mock service.
+        when(mockSchedulerService.getAllSchedulerConfigs()).thenReturn(configs);
+        
+        // Execute and validate.
+        ResourceList<MasterSchedulerConfig> result = controller.getAllSchedulerConfigs();
+        
+        assertEquals(result.getItems().size(), 3);
```
DwayneJengSage, line 125: We should verify not just the size but also the content.

src/test/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerControllerTest.java
```diff
@@ -0,0 +1,242 @@
...
+        
+        mockRequestBody(mockRequest, mockConfig);
+        
+        // Mock service.
+        when(mockSchedulerService.createSchedulerConfig(any())).thenReturn(mockConfig);
+        
+        // Execute and validate.
+        MasterSchedulerConfig result = controller.createSchedulerConfig();
+        
+        assertEquals(result.getScheduleId(), SCHEDULE_ID);
```
DwayneJengSage, line 142: Similarly, we should verify the config returned by the controller is the same as the config that was returned by the service. (Same for update().)

src/test/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerControllerTest.java
```diff
@@ -0,0 +1,242 @@
...
+        MasterSchedulerConfig updatedConfig = MasterSchedulerConfig.create();
+        updatedConfig.setScheduleId("test-schedule-id");
+        updatedConfig.setCronSchedule(UPDATED_CRON_SCHEDULE);
+        updatedConfig.setRequestTemplate(mockConfig.getRequestTemplate());
+        updatedConfig.setSqsQueueUrl(UPDATED_SQS_QUEUE_URL);
+        updatedConfig.setVersion(2L);
+        mockRequestBody(mockRequest, updatedConfig);
+        
+        // Mock service.
+        when(mockSchedulerService.getSchedulerConfig(SCHEDULE_ID)).thenReturn(mockConfig);
```
DwayneJengSage: You don't need to mock schedulerService.get(), because since it's a mock, it will never be called. Also, we don't want implementation details from the service to leak into our tests for the controller.

src/main/java/org/sagebionetworks/bridge/validators/MasterSchedulerConfigValidator.java
```diff
@@ -0,0 +1,52 @@
+package org.sagebionetworks.bridge.validators;
+
+import org.sagebionetworks.bridge.models.schedules.MasterSchedulerConfig;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+/** Validator for MasterSchedulerConfig. */
+public class MasterSchedulerConfigValidator implements Validator {
```
DwayneJengSage, line 10: Need unit test for validator.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/models/TimestampHolder.java
```diff
@@ -0,0 +1,19 @@
+package org.sagebionetworks.bridge.models;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class TimestampHolder {
```
alxdarksage: Usually in our APIs, time values are expressed using ISO 8601 string values... where we store longs in the persistence tier, these values are always in UTC ("Z") time, but are still immediately recognizable in the API as time values. I would personally prefer we return something like a DateTimeHolder than a LongHolder (which is all this is).

src/main/java/org/sagebionetworks/bridge/models/schedules/MasterSchedulerStatus.java
```diff
@@ -0,0 +1,30 @@
...
+     * The hashkey is always BridgeMasterScheduler.
+     */
+    public String getHashKey();
+    void setHashKey(String hashKey);
+    
+    /**
+     * The last time the Master Scheduler ran.
+     */
+    public Long getLastProcessedTime();
+    public void setLastProcessedTime(Long lastProcessedTime);
```
alxdarksage, line 27: Same thing: can we translate this into a DateTime value? It can still be persisted as a Long.

src/main/java/org/sagebionetworks/bridge/spring/controllers/MasterSchedulerController.java
```diff
@@ -0,0 +1,96 @@
...
+        schedulerService.deleteSchedulerConfig(scheduleId);
+        return DELETED_MSG;
+    }
+    
+
+    /**
+     * Get the last time scheduler ran.
+     */
+    @GetMapping("/v3/schedulerstatus")
+    public TimestampHolder getSchedulerStatus() {
```
alxdarksage: See prior comment (I would prefer a less ambiguous DateTimeHolder).

alxdarksage (changes requested):

janaechan:
The lastProcessedTime is still stored as a Long in MasterSchedulerStatus but is converted into a DateTime in the StatusDao. Will this work?

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDao.java
```diff
@@ -0,0 +1,35 @@
...
+        DynamoMasterSchedulerStatus statusKey = new DynamoMasterSchedulerStatus();
+        statusKey.setHashKey(SCHEDULER_STATUS_HASH_KEY);
+        
+        DynamoMasterSchedulerStatus status = mapper.load(statusKey);
+        if (status == null) {
+            return null;
+        }
+        return new DateTimeHolder(new DateTime(status.getLastProcessedTime()));
+    }
+}
```
alxdarksage, line 35: This will work. There may be an annotation that you can put on DynamoMasterSchedulerStatus which does the same thing if you return status from the API... it'll convert a long to a JSON DateTime value:

@JsonDeserialize(using = DateTimeToLongDeserializer.class)

As status has only one field, it kind of makes sense to just return it from the API rather than the holder object, but this also works.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDaoTest.java
```diff
@@ -0,0 +1,62 @@
...
+    PaginatedScanList<DynamoMasterSchedulerStatus> mockScanList;
+    
+    @Captor
+    private ArgumentCaptor<DynamoMasterSchedulerStatus> configCaptor;
+    
+    @Captor
+    ArgumentCaptor<DynamoMasterSchedulerStatus> scanCaptor;
+    
+    @Captor
+    private ArgumentCaptor<DynamoDBSaveExpression> saveExpressionCaptor;
```
DwayneJengSage: Minor: A lot of these mocks and captors aren't used in this test.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerStatusDaoTest.java
```diff
@@ -0,0 +1,62 @@
...
+        DateTimeHolder timestamp = new DateTimeHolder(TIMESTAMP);
+        DynamoMasterSchedulerStatus status = new DynamoMasterSchedulerStatus();
+        status.setHashKey(DynamoMasterSchedulerStatusDao.SCHEDULER_STATUS_HASH_KEY);
+        status.setLastProcessedTime(TIMESTAMP.getMillis());
+        
+        when(mockMapper.load(any())).thenReturn(status);
+        
+        DateTimeHolder result = dao.getLastProcessedTime();
+        assertEquals(result.getDateTime().getMillis(), timestamp.getDateTime().getMillis());
+    }
```
DwayneJengSage, line 46: We should also have a test case for the null branch, to ensure we've covered all of our code paths.

src/test/java/org/sagebionetworks/bridge/models/DateTimeHolderTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.testng.annotations.Test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class DateTimeHolderTest {
+    
+    @Test
+    public void canSerialize() throws Exception {
+        DateTimeHolder holder = new DateTimeHolder(TIMESTAMP);
```
DwayneJengSage: We should use a timestamp with a non-UTC timezone to make sure our serialization also handles timezones correctly.

src/test/java/org/sagebionetworks/bridge/models/DateTimeHolderTest.java
```diff
@@ -0,0 +1,25 @@
...
+        
+        JsonNode node = BridgeObjectMapper.get().valueToTree(holder);
+        
+        assertEquals(TIMESTAMP.toString(), node.get("dateTime").textValue());
+        assertEquals("DateTimeHolder", node.get("type").textValue());
+        
+        DateTimeHolder deser = BridgeObjectMapper.get().readValue(node.toString(), DateTimeHolder.class);
+        assertEquals(deser.getDateTime(), TIMESTAMP);
```
DwayneJengSage: Similarly, DateTime.equals() doesn't check timestamp. We want to verify the timezone is correct. I believe TestUtils.assertDatesWithTimeZoneEqual() will do this for you.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerConfigDaoTest.java
```diff
@@ -0,0 +1,203 @@
...
+        
+        // verify mapper call
+        verify(mockMapper).save(configCaptor.capture(), any(DynamoDBSaveExpression.class));
+        
+        DynamoMasterSchedulerConfig mapperConfig = configCaptor.getValue();
+        assertEquals(mapperConfig.getScheduleId(), SCHEDULE_ID);
+        assertEquals(mapperConfig.getCronSchedule(), CRON_SCHEDULE);
+        assertTrue(mapperConfig.getRequestTemplate().get("a").booleanValue());
+        assertEquals(mapperConfig.getRequestTemplate().get("b").textValue(), "string");
+        assertEquals(mapperConfig.getSqsQueueUrl(), SQS_QUEUE_URL);
```
DwayneJengSage, line 107: We should also verify that create() nulls out the version number.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #111, opened 2019-09-18
File hosting

alxdarksage:
Adding the metadata for a hosted file (for which there will be revisions).


src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
```diff
@@ -0,0 +1,221 @@
...
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.exceptions.InvalidEntityException;
+import org.sagebionetworks.bridge.models.files.FileMetadata;
+
+public class FileServiceTest extends Mockito {
+
+    private static final String NAME = "oneName";
+    
+    @Spy
+    FileMetadataDao mockFileDao;
```
DwayneJengSage, line 36: Did you mean Mock? Is there any situation where we want to call through to the actual mockFileDao?

src/main/resources/db/changelog/changelog.sql
```diff
@@ -208,3 +208,29 @@ CREATE TABLE IF NOT EXISTS `RequestInfos` (
...
+    `guid` varchar(60) NOT NULL,
+    `name` varchar(255) DEFAULT NULL,
+    `createdOn` BIGINT UNSIGNED NOT NULL,
+    `modifiedOn` BIGINT UNSIGNED NOT NULL,
+    `description` text,
+    `mimeType` varchar(255) DEFAULT NULL,
+    `deleted` tinyint(1) NOT NULL DEFAULT '0',
+    `version` int(10) unsigned NOT NULL,
+    PRIMARY KEY (`guid`),
+    KEY `Studies_idx` (`studyId`)
```
DwayneJengSage: Do we also want a compound index by studyId and guid?

DwayneJengSage (approved):
Looks good. Just a few minor questions.

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
alxdarksage: Yes this should be a mock.

src/main/resources/db/changelog/changelog.sql
alxdarksage: Yes, we ensure study correctness by using both so we should add an index.

DwayneJengSage (approved):

PR #113, opened 2019-09-26
Allow survey API to accept identifier or GUID

janaechan:
https://sagebionetworks.jira.com/browse/BRIDGE-2259

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
@@ -48,6 +48,8 @@ import com.google.common.collect.Maps;
 @Component
 public class DynamoSurveyDao implements SurveyDao {
 
+    static final String IDENTIFIER_PREFIX = "identifier:";
+    
     class QueryBuilder {
         
         private static final String PUBLISHED_PROPERTY = "published";
@@ -91,7 +93,7 @@ public class DynamoSurveyDao implements SurveyDao {
             List<DynamoSurvey> dynamoSurveys = null;
             if (surveyGuid == null) {
                 dynamoSurveys = queryBySecondaryIndex();
-            } else if (surveyGuid.startsWith("identifier:")) { 
+            } else if (surveyGuid.toLowerCase().startsWith(IDENTIFIER_PREFIX)) {
                 dynamoSurveys = queryBySurveyIdentifier();
             } else {
                 dynamoSurveys = query();
@@ -296,8 +298,8 @@ public class DynamoSurveyDao implements SurveyDao {
     }
     
     @Override
-    public Survey updateSurvey(Survey survey) {
-        Survey existing = getSurvey(survey, false);
+    public Survey updateSurvey(StudyIdentifier studyIdentifier, Survey survey) {
+        Survey existing = getSurvey(studyIdentifier, survey, false);
         
         // copy over mutable fields
         existing.setName(survey.getName());
@@ -316,8 +318,8 @@ public class DynamoSurveyDao implements SurveyDao {
     }
     
     @Override
-    public Survey versionSurvey(GuidCreatedOnVersionHolder keys) {
-        DynamoSurvey existing = (DynamoSurvey)getSurvey(keys, true);
+    public Survey versionSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
+        DynamoSurvey existing = (DynamoSurvey)getSurvey(studyIdentifier, keys, true);
         DynamoSurvey copy = new DynamoSurvey(existing);
         copy.setPublished(false);
         copy.setDeleted(false);
@@ -341,8 +343,9 @@ public class DynamoSurveyDao implements SurveyDao {
     }
 
     @Override
-    public void deleteSurveyPermanently(GuidCreatedOnVersionHolder keys) {
-        Survey existing = getSurvey(keys, false);
+    public void deleteSurveyPermanently(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
+        Survey existing = getSurvey(studyIdentifier, keys, false);
+
         if (existing != null) {
             deleteAllElements(existing.getGuid(), existing.getCreatedOn());
             surveyMapper.delete(existing);
@@ -392,7 +395,11 @@ public class DynamoSurveyDao implements SurveyDao {
      * version (not a specific timestamped version), this method should be rarely called.
      */
     @Override
-    public Survey getSurvey(GuidCreatedOnVersionHolder keys, boolean includeElements) {
+    public Survey getSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys, boolean includeElements) {
+        if (keys.getGuid() != null && keys.getGuid().toLowerCase().startsWith(IDENTIFIER_PREFIX)) {
+            return new QueryBuilder().setStudy(studyIdentifier).setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
+                    .setSkipElements(!includeElements).getOne();
+        }
         return new QueryBuilder().setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
                 .setSkipElements(!includeElements).getOne();
     }
--- a/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
@@ -73,7 +73,11 @@ public class DynamoSurveyDao implements SurveyDao {
             return this;
         }
         QueryBuilder setStudy(StudyIdentifier studyIdentifier) {
-            this.studyIdentifier = studyIdentifier.getIdentifier();
+            if (studyIdentifier == null) {
+                this.studyIdentifier = null;
+            } else {
+                this.studyIdentifier = studyIdentifier.getIdentifier();
+            }
             return this;
         }
         QueryBuilder setCreatedOn(long createdOn) {
@@ -396,11 +400,7 @@ public class DynamoSurveyDao implements SurveyDao {
      */
     @Override
     public Survey getSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys, boolean includeElements) {
-        if (keys.getGuid() != null && keys.getGuid().toLowerCase().startsWith(IDENTIFIER_PREFIX)) {
-            return new QueryBuilder().setStudy(studyIdentifier).setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
-                    .setSkipElements(!includeElements).getOne();
-        }
-        return new QueryBuilder().setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
+        return new QueryBuilder().setStudy(studyIdentifier).setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
                 .setSkipElements(!includeElements).getOne();
     }
 
--- a/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
@@ -2,6 +2,9 @@ package org.sagebionetworks.bridge.dynamodb;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 
@@ -14,6 +17,8 @@ import org.sagebionetworks.bridge.exceptions.ConcurrentModificationException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.time.DateUtils;
 import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolder;
+import org.sagebionetworks.bridge.models.appconfig.AppConfigElement;
+import org.sagebionetworks.bridge.models.schedules.ScheduledActivity;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifierImpl;
 import org.sagebionetworks.bridge.models.surveys.Constraints;
@@ -21,6 +26,7 @@ import org.sagebionetworks.bridge.models.surveys.Survey;
 import org.sagebionetworks.bridge.models.surveys.SurveyElement;
 import org.sagebionetworks.bridge.models.surveys.SurveyElementFactory;
 import org.sagebionetworks.bridge.models.surveys.SurveyQuestion;
+import org.sagebionetworks.bridge.models.upload.Upload;
 import org.sagebionetworks.bridge.models.upload.UploadSchema;
 import org.sagebionetworks.bridge.services.UploadSchemaService;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -140,6 +142,30 @@ Survey getOne() {
             }
             return surveyMapper.queryPage(DynamoSurvey.class, query).getResults();
         }
+        
+        private List<DynamoSurvey> queryBySurveyIdentifier() {
+            if (studyIdentifier == null) {
+                throw new IllegalStateException("Calculated the need to query by secondary index, but study identifier is not set");
```
alxdarksage: Language: "Retrieval of survey by identifier requires a study identifier"

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -85,6 +85,8 @@ QueryBuilder setDeleted(boolean includeDeleted) {
             List<DynamoSurvey> dynamoSurveys = null;
             if (surveyGuid == null) {
                 dynamoSurveys = queryBySecondaryIndex();
+            } else if (surveyGuid.startsWith("identifier:")) {
```
alxdarksage: I would use some constants: "identifier:" can be a string constant, and then instead of a magic number (11), you can use the constant string's length, consolidating any future change in one place.

Trivial: in the account case, the prefix is case insensitive, this should be case insensitive too.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -140,6 +142,30 @@ Survey getOne() {
...
+                query.withQueryFilterEntry(PUBLISHED_PROPERTY, equalsNumber("1"));
+            }
+            if (notDeleted) {
+                query.withQueryFilterEntry(DELETED_PROPERTY, equalsNumber("0"));
+            }
+            if (createdOn != 0L) {
+                query.withRangeKeyCondition(CREATED_ON_PROPERTY, equalsNumber(Long.toString(createdOn)));
+            }
+            
+            return surveyMapper.queryPage(DynamoSurvey.class, query).getResults();
```
alxdarksage: I think, for this to be effective, you have to use the index specified (in this case, the studyKey-identifier-index), via the Index object. Look at the use of DynamoIndexHelper in the DynamoScheduledActivityDao for an example of this.

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
DwayneJengSage: DynamoIndexHelper is buggy and should not be used. mapper.query() handles secondary indices just fine. See queryBySecondaryIndex().

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -140,6 +142,30 @@ Survey getOne() {
...
             return surveyMapper.queryPage(DynamoSurvey.class, query).getResults();
         }
+        
+        private List<DynamoSurvey> queryBySurveyIdentifier() {
+            if (studyIdentifier == null) {
+                throw new IllegalStateException("Calculated the need to query by secondary index, but study identifier is not set");
+            }
+            DynamoSurvey hashKey = new DynamoSurvey();
+            hashKey.setStudyIdentifier(studyIdentifier);
+            hashKey.setIdentifier(surveyGuid.substring(11));
```
DwayneJengSage: Minor: I believe the withHashKeyValues() only looks at hash key, so setting the identifier won't actually do anything. Since you specify the identifier in the range key, you don't have to specify the identifier here.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -140,6 +142,30 @@ Survey getOne() {
...
+            query.withConsistentRead(false);
+            query.withRangeKeyCondition("identifier", equalsString(surveyGuid.substring(11)));
+            if (published) {
+                query.withQueryFilterEntry(PUBLISHED_PROPERTY, equalsNumber("1"));
+            }
+            if (notDeleted) {
+                query.withQueryFilterEntry(DELETED_PROPERTY, equalsNumber("0"));
+            }
+            if (createdOn != 0L) {
+                query.withRangeKeyCondition(CREATED_ON_PROPERTY, equalsNumber(Long.toString(createdOn)));
```
DwayneJengSage: You can't specify two range keys. You'll need a filter for this to work.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
DwayneJengSage: Querying by guid returns the surveys from newest to oldest (descending createdOn order). Since we don't have the range key to do that for us, we'll need to sort the results ourselves before returning them.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -675,6 +676,92 @@ public void constraintRulesOnQuestionOverwriteConstraintRules() {
...
+        List<DynamoSurvey> surveyList = ImmutableList.of(new DynamoSurvey(), new DynamoSurvey());
+        mockSurveyMapper(surveyList);
+        
+        
+        List<Survey> results = dao.getSurveyAllVersions(TEST_STUDY, SURVEY_IDENTIFIER, true);
+        assertEquals(results, surveyList);
+        
+        verify(mockSurveyMapper).queryPage(eq(DynamoSurvey.class), queryCaptor.capture());
+        DynamoDBQueryExpression<DynamoSurvey> query = queryCaptor.getValue();
+        assertEquals(query.getHashKeyValues().getIdentifier(), SURVEY_IDENTIFIER.substring(11));
```
DwayneJengSage: It's more important to test the studyIdentifier, since that's the actual hash key. (Here and in all the other tests.)

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -675,6 +676,92 @@ public void constraintRulesOnQuestionOverwriteConstraintRules() {
...
+        assertEquals(rangeKeyCondition.getAttributeValueList().get(0).getS(), SURVEY_ID);
+        
+        assertEquals(query.getQueryFilter().size(), 1);
+        verifyIsDeletedQueryCondition(query);
+    }
+    
+    @Test
+    public void getSurveyMostRecentVersionByIdentifier() {
+        DynamoSurvey persisted = new DynamoSurvey();
+        mockSurveyMapper(persisted);
```
DwayneJengSage: Consider creating multiple surveys with createdOn in some arbitrary order to make sure that we're sorting it correctly and returning the most recent one. (Same with getSurveyMostRecentlyPublishedVersion().)

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -675,6 +676,92 @@ public void constraintRulesOnQuestionOverwriteConstraintRules() {
...
+        Condition rangeKeyCondition = query.getRangeKeyConditions().get("identifier");
+        assertEquals(rangeKeyCondition.getComparisonOperator(), EQ.name());
+        assertEquals(rangeKeyCondition.getAttributeValueList().get(0).getS(), SURVEY_ID);
+        
+        assertEquals(query.getQueryFilter().size(), 2);
+        verifyIsDeletedQueryCondition(query);
+        verifyPublishedQueryCondition(query);
+        
+        verify(mockSurveyElementMapper).queryPage(eq(DynamoSurveyElement.class), any());
+    }
```
DwayneJengSage, line 860: We also need a test for getSurvey(), particularly with createdOn.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -85,6 +91,8 @@ QueryBuilder setDeleted(boolean includeDeleted) {
             List<DynamoSurvey> dynamoSurveys = null;
             if (surveyGuid == null) {
                 dynamoSurveys = queryBySecondaryIndex();
+            } else if (surveyGuid.startsWith("identifier:")) {
```
DwayneJengSage: We should move "identifier:" to a constant.

We should make this check case-insensitive.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -140,6 +148,34 @@ Survey getOne() {
...
+            
+            if (published) {
+                query.withQueryFilterEntry(PUBLISHED_PROPERTY, equalsNumber("1"));
+            }
+            if (notDeleted) {
+                query.withQueryFilterEntry(DELETED_PROPERTY, equalsNumber("0"));
+            }
+            if (createdOn != 0L) {
+                query.withQueryFilterEntry(CREATED_ON_PROPERTY, equalsNumber(Long.toString(createdOn)));
+            }
```
DwayneJengSage, line 177: We need a test for getSurvey() by identifier with createdOn.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -669,12 +672,216 @@ public void constraintRulesOnQuestionOverwriteConstraintRules() {
...
+        when(mockSurveyElementMapper.queryPage(eq(DynamoSurveyElement.class), any())).thenReturn(mockElementResultsPage);
+        when(mockElementResultsPage.getResults()).thenReturn(ImmutableList.of(element));
+        
+        Survey result = dao.getSurvey(TEST_STUDY, SURVEY_IDENTIFIER_KEYS, true);
+        assertSame(result, survey);
+        assertEquals(result.getElements().size(), 1);
+        
+        verify(mockSurveyMapper).queryPage(eq(DynamoSurvey.class), queryCaptor.capture());
+        DynamoDBQueryExpression<DynamoSurvey> query = queryCaptor.getValue();
+        assertEquals(query.getHashKeyValues().getStudyIdentifier(), TEST_STUDY_IDENTIFIER);
```
DwayneJengSage, line 705: We should also verify that the query contains (a) the survey identifier as a range key (c) the createdOn as a query filter.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDao.java
```diff
@@ -356,7 +395,11 @@ public Survey getSurveyMostRecentlyPublishedVersion(StudyIdentifier studyIdentif
      * version (not a specific timestamped version), this method should be rarely called.
      */
     @Override
-    public Survey getSurvey(GuidCreatedOnVersionHolder keys, boolean includeElements) {
+    public Survey getSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys, boolean includeElements) {
+        if (keys.getGuid() != null && keys.getGuid().toLowerCase().startsWith(IDENTIFIER_PREFIX)) {
+            return new QueryBuilder().setStudy(studyIdentifier).setSurvey(keys.getGuid()).setCreatedOn(keys.getCreatedOn())
+                    .setSkipElements(!includeElements).getOne();
+        }
```
DwayneJengSage: This duplicates a lot of the same code as getAll(). Why not always pass in studyIdentifier?

I would go as far as to put the null-check in QueryBuilder.setStudy(). If the object studyIdentifier is null, then it sets the string studyIdentifier to null.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -627,8 +630,8 @@ public void getAllSurveysMostRecentVersionExcludeDeleted() {
     
     @Test(expectedExceptions = IllegalStateException.class, expectedExceptionsMessageRegExp = 
             "Calculated the need to query by secondary index, but study identifier is not set")
-    public void secondaryIndexRequiresStudyId() {
-        dao.versionSurvey(new DynamoSurvey());
+    public void secondaryIndexRequiresStudyIdWithStudyIdentifier() {
+        dao.versionSurvey(TEST_STUDY, new DynamoSurvey());
```
DwayneJengSage: This test is a little bit contrived. In practice, versionSurvey() will only ever be used to get a survey by guid+createdOn or by studyId+identifier+createdOn. I would rewrite this test to use getAllSurveysMostRecentVersion().

I would also go into SurveyService and add a checkNotNull() for survey guid and createdOn for any API that calls getSurvey(), such as get, delete, update, version, and publish.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoSurveyDaoTest.java
```diff
@@ -669,12 +672,216 @@ public void constraintRulesOnQuestionOverwriteConstraintRules() {
...
+        survey.setCreatedOn(TIMESTAMP);
+        mockSurveyMapper(survey);
+        
+        // There is a question that should be retrieved.
+        DynamoSurveyQuestion element = new DynamoSurveyQuestion();
+        element.setType(SurveyElementConstants.SURVEY_QUESTION_TYPE);
+        element.setUiHint(UIHint.BLOODPRESSURE);
+        element.setConstraints(new BloodPressureConstraints());
+        when(mockSurveyElementMapper.queryPage(eq(DynamoSurveyElement.class), any())).thenReturn(mockElementResultsPage);
+        when(mockElementResultsPage.getResults()).thenReturn(ImmutableList.of(element));
```
DwayneJengSage: There is no need to mock the mockSurveyElementMapper.queryPage(), as it will never be called.

src/test/java/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -336,7 +336,7 @@ public void successfulDeletePermanently() {
         assertEquals(paramsCaptor.getValue().get("surveyGuid"), survey.getGuid());
         assertEquals(paramsCaptor.getValue().get("surveyCreatedOn"), survey.getCreatedOn());
         
-        verify(mockSurveyDao).deleteSurveyPermanently(keysCaptor.capture());
+        verify(mockSurveyDao).deleteSurveyPermanently(any(), keysCaptor.capture());
```
DwayneJengSage: We should verify the study ID.

src/test/java/org/sagebionetworks/bridge/services/SurveyServiceMockTest.java
```diff
@@ -595,32 +595,32 @@ public void updateSurveyAlreadyPublishedThrowsException() {
...
         Survey update = Survey.create();
-        update.setStudyIdentifier(TestConstants.TEST_STUDY_IDENTIFIER);
+        update.setStudyIdentifier(TEST_STUDY_IDENTIFIER);
         update.setDeleted(false);
         
-        service.updateSurvey(TestConstants.TEST_STUDY, update);
+        service.updateSurvey(TEST_STUDY, update);
         
-        verify(mockSurveyDao).updateSurvey(surveyCaptor.capture());
+        verify(mockSurveyDao).updateSurvey(any(), surveyCaptor.capture());
```
DwayneJengSage: We should verify the study ID.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #116, opened 2019-10-03
File revisions

alxdarksage:
Need to add the new credentials to the infrastructure build scripts before this can be merged.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/FileController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/FileController.java
@@ -4,8 +4,7 @@ import static org.sagebionetworks.bridge.BridgeConstants.API_DEFAULT_PAGE_SIZE;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.DEVELOPER;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-
+import org.joda.time.DateTime;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.CrossOrigin;
@@ -18,7 +17,6 @@ import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 
 import org.sagebionetworks.bridge.BridgeUtils;
-import org.sagebionetworks.bridge.json.BridgeObjectMapper;
 import org.sagebionetworks.bridge.models.GuidVersionHolder;
 import org.sagebionetworks.bridge.models.PagedResourceList;
 import org.sagebionetworks.bridge.models.ResourceList;
@@ -33,6 +31,8 @@ import org.sagebionetworks.bridge.services.FileService;
 public class FileController extends BaseController {
     
     static final StatusMessage DELETE_MSG = new StatusMessage("File metadata and revisions deleted.");
+    static final StatusMessage UPLOAD_FINISHED_MSG = new StatusMessage("File revision upload completed.");
+
     private FileService fileService;
     
     @Autowired
@@ -114,5 +114,15 @@ public class FileController extends BaseController {
         revision.setFileGuid(guid);
         
         return fileService.createFileRevision(session.getStudyIdentifier(), revision);
+    }
+    
+    @PostMapping("/v3/files/{guid}/revisions/{createdOn}")
+    public StatusMessage finishFileRevision(@PathVariable String guid, @PathVariable("createdOn") String createdOnStr) throws Exception {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        
+        DateTime createdOn = DateTime.parse(createdOnStr);
+        
+        fileService.finishFileRevision(session.getStudyIdentifier(), guid, createdOn);
+        return UPLOAD_FINISHED_MSG;
     }
 }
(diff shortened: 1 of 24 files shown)


src/main/resources/BridgeServer2.conf
```diff
@@ -141,6 +141,10 @@ dev.consents.bucket = org-sagebridge-consents-dev
 uat.consents.bucket = org-sagebridge-consents-uat
 prod.consents.bucket = org-sagebridge-consents-prod
 
+# AWS credentials for presigned URL to PUT in the docs folder
+aws.key.file.revisions = dummy-value 
+aws.secret.key.file.revisions = dummy-value
```
DwayneJengSage: I can't find where these are being used.

Also, if we _are_ using these, we should consider not using them. My understanding is we wanted to consolidate to using a single AWS IAM user for Bridge, to make it easier to manage the credentials.

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
```diff
@@ -218,5 +257,126 @@ public void deleteAllStudyFiles() {
...
+        service.getFileRevisions(TEST_STUDY,GUID, 0, 1);
+    }
+    
+    @Test(expectedExceptions = BadRequestException.class)
+    public void getFileRevisionsSizeTooLarge() {
+        FileMetadata metadata = new FileMetadata();
+        doReturn(metadata).when(service).getFile(TEST_STUDY, GUID);
+        
+        service.getFileRevisions(TEST_STUDY, GUID, 0, 1000);
+    }
```
DwayneJengSage, line 300: Do we also want a getFileRevisionsNotFound()?

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
```diff
@@ -218,5 +257,126 @@ public void deleteAllStudyFiles() {
...
+    
+    @Test(expectedExceptions = EntityNotFoundException.class, 
+            expectedExceptionsMessageRegExp=".*FileMetadata not found.*")
+    public void createFileRevisionNotFound() throws Exception {
+        FileRevision revision = new FileRevision();
+        revision.setFileGuid(GUID);
+        
+        service.createFileRevision(TEST_STUDY, revision);
+    }
```
DwayneJengSage: What about createFileRevisionInvalid()? (Same for update.)

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
```diff
@@ -218,5 +257,126 @@ public void deleteAllStudyFiles() {
...
+        
+    @Test
+    public void finishFileRevision() {
+        FileRevision existing = new FileRevision();
+        existing.setFileGuid(GUID);        
+        existing.setCreatedOn(TIMESTAMP);
```
DwayneJengSage, line 361: Should we set a dummy upload URL so we can verify that it's actually being blanked out?

src/main/java/org/sagebionetworks/bridge/spring/controllers/FileController.java
```diff
@@ -88,4 +93,37 @@ public StatusMessage deleteFile(@PathVariable String guid,
...
+    @GetMapping("/v3/files/{guid}/revisions")
+    public PagedResourceList<FileRevision> getFileRevisions(@PathVariable String guid,
+            @RequestParam(required = false) String offsetBy, @RequestParam(required = false) String pageSize) {
+        UserSession session = getAuthenticatedSession(DEVELOPER);
+        
+        int offsetInt = BridgeUtils.getIntOrDefault(offsetBy, 0);
+        int pageSizeInt = BridgeUtils.getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+
+        return fileService.getFileRevisions(session.getStudyIdentifier(), guid, offsetInt, pageSizeInt);
+    }
```
DwayneJengSage, line 106: Where are the unit tests for getFileRevisions()?

DwayneJengSage (changes requested):

src/main/resources/BridgeServer2.conf
alxdarksage: After discussing with Aaron and Khai, it was decided that the full s3 client can make credentials that are scoped in the presigned URLs without having a user with scoped permissions like this. I'll remove these.

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
alxdarksage: Added this, moved around the checks so that we don't trigger an NPE before discovering the revision has no file GUID (as called from the controller, there will always be something in the fileGuid field though).

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/FileServiceTest.java
alxdarksage: (There's no update call... finish changes two fields on an existing record with no user input.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/FileController.java
alxdarksage: totally missed these... added.

DwayneJengSage (approved):

PR #118, opened 2019-10-09
Add hosted file references to the app config file.

alxdarksage:


src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
```diff
@@ -148,6 +157,12 @@ public void canSerialize() throws Exception {
...
         assertEquals(node.get("surveyReferences").get(1).get("createdOn").textValue(), SURVEY_PUB_DATE.toString());
         
+        assertEquals(node.get("fileReferences").size(), 2);
+        assertEquals(node.get("fileReferences").get(0).get("guid").textValue(), GUID);
+        assertEquals(node.get("fileReferences").get(0).get("createdOn").textValue(), TIMESTAMP.toString());
+        assertEquals(node.get("fileReferences").get(1).get("guid").textValue(), "twoGuid");
+        assertEquals(node.get("fileReferences").get(1).get("createdOn").textValue(), TIMESTAMP.toString());
+        
         AppConfig deser = BridgeObjectMapper.get().treeToValue(node, AppConfig.class);
         assertNull(deser.getStudyId());
```
DwayneJengSage, line 167: Should we check file references in the deserialized object?

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,32 @@
...
+            .getHostnameWithPostfix("docs") + "/" + GUID + "." + TIMESTAMP.getMillis();
+
+    @Test
+    public void test() {
+        FileReference ref = new FileReference(GUID, TIMESTAMP);
+        
+        assertEquals(ref.getGuid(), GUID);
+        assertEquals(ref.getCreatedOn(), TIMESTAMP);
+        assertTrue(ref.getHref().contains(HREF));
+    }
```
DwayneJengSage: Would it be better to mock the Config so we can verify our logic and the exact URL?

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.sagebionetworks.bridge.TestConstants.TIMESTAMP;
+import static org.sagebionetworks.bridge.config.Environment.PROD;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+import org.testng.annotations.Test;
+
+import org.sagebionetworks.bridge.config.BridgeConfigFactory;
+
+public class FileReferenceTest {
```
DwayneJengSage: What about the null guid case and the null createdOn case?

We also need an equalsVerifier test and a JSON serialization test.

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,32 @@
...
+        FileReference ref = new FileReference(GUID, TIMESTAMP);
+        
+        assertEquals(ref.getGuid(), GUID);
+        assertEquals(ref.getCreatedOn(), TIMESTAMP);
+        assertTrue(ref.getHref().contains(HREF));
+    }
+ 
+    @Test
+    public void testWithEnv() {
+        FileReference ref = new FileReference(PROD, "docs.test.com", GUID, TIMESTAMP);
```
DwayneJengSage: We need to test both Prod and non-Prod.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoAppConfigTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: Added tests that nulls don't break anything and added equalsVerifier.

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,57 @@
...
+        EqualsVerifier.forClass(FileReference.class).allFieldsShouldBeUsed().verify();
+    }
+    
+    @Test
+    public void nullValues() throws Exception {
+        FileReference ref = BridgeObjectMapper.get().readValue("{}", FileReference.class);
+        assertNull(ref.getGuid());
+        assertNull(ref.getCreatedOn());
+        assertNull(ref.getHref());
+    }
```
DwayneJengSage: I think we need two separate tests for null guid and null createdOn, because they're two separate branches in getHref().

Also, is it necessary to use JSON deserialization for this test? It would be more clear to call the constructor directly.

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.testng.annotations.Test;
+
+import org.sagebionetworks.bridge.config.BridgeConfigFactory;
+import org.sagebionetworks.bridge.config.Environment;
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+
+public class FileReferenceTest {
```
DwayneJengSage: Still need a JSON serialization test.

src/main/java/org/sagebionetworks/bridge/models/files/FileReference.java
```diff
@@ -0,0 +1,78 @@
...
+    private final String guid;
+    private final DateTime createdOn;
+    
+    @JsonCreator
+    public FileReference(@JsonProperty("guid") String guid, @JsonProperty("createdOn") DateTime createdOn) {
+        this.guid = guid;
+        this.createdOn = (createdOn == null) ? null : createdOn.withZone(DateTimeZone.UTC);
+        this.env = BridgeConfigFactory.getConfig().getEnvironment();
+        this.baseUrl = BridgeConfigFactory.getConfig().getHostnameWithPostfix("docs");
+    }
```
DwayneJengSage: If we don't want to use PowerMock, how about using a FileReferenceFactory?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: Split and removed Jackson here.

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/models/files/FileReference.java
alxdarksage: Switched to PowerMock.

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: We're using PowerMock now to mock the config

alxdarksage:
This is ready to be reviewed

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
```diff
@@ -0,0 +1,93 @@
...
+    public void canSerialization() throws Exception {
+        String href = "http://" + BridgeConfigFactory.getConfig()
+                .getHostnameWithPostfix("docs") + "/" + GUID + "." + TIMESTAMP.getMillis();
+        
+        FileReference ref = new FileReference(GUID, TIMESTAMP);
+        
+        JsonNode node = BridgeObjectMapper.get().valueToTree(ref);
+        assertEquals(node.get("guid").textValue(), GUID);
+        assertEquals(node.get("createdOn").textValue(), TIMESTAMP.toString());
+        assertEquals(node.get("type").textValue(), "FileReference");
```
DwayneJengSage, line 61: We should also check that href is being serialized correctly.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/models/files/FileReferenceTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #121, opened 2019-10-25
Add synapse user id to Bridge accounts.

alxdarksage:


src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -324,6 +329,7 @@ public StudyParticipant getParticipant(Study study, Account account, boolean inc
         builder.withAttributes(account.getAttributes());
         builder.withSubstudyIds(assoc.getSubstudyIdsVisibleToCaller());
         builder.withExternalIds(assoc.getExternalIdsVisibleToCaller());
+        builder.withSynapseUserId(account.getSynapseUserId());
         return builder;
```
DwayneJengSage, line 333: It looks like this is being called by getParticipant() and getSelfParticipant(). Should we update the tests for these APIs to check for synapseUserId?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -500,6 +509,11 @@ private boolean shouldEnableCompleteExternalIdAccount(StudyParticipant participa
         return participant.getEmail() == null && participant.getPhone() == null && 
             participant.getExternalId() != null && participant.getPassword() != null;
     }
+    
+    private boolean shouldEnableCompleteSynapseUserIdAccount(StudyParticipant participant) {
+        return participant.getEmail() == null && participant.getPhone() == null && 
+            participant.getSynapseUserId() != null && participant.getPassword() == null;
+    }
```
DwayneJengSage, line 517: Is there any reason to gate this on email, phone, or password? If they have a Synapse User ID, even if they have a email, phone, or password, is there any reason we need to block them from signing in via Synapse?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -559,7 +573,8 @@ private void updateAccountAndRoles(Study study, Account account, StudyParticipan
         account.setNotifyByEmail(participant.isNotifyByEmail());
         account.setDataGroups(participant.getDataGroups());
         account.setLanguages(participant.getLanguages());
-        account.setMigrationVersion(AccountDao.MIGRATION_VERSION);
+        account.setSynapseUserId(participant.getSynapseUserId());
```
DwayneJengSage: Should this be in updateIdentifiers() instead?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: It's an interesting point. We have a global status flag that can be set to UNVERIFIED and if the user has provided an email or phone number, we would normally leave that flag as UNVERIFIED. But we have separate flags for email and phone so perhaps we could ignore this flag, or just make it UNVERIFIED if all the non-null verification flags are set to false (or all set to true depending on what we mean by this... I think though that if you have proven you can access the system with one credential, you are ENABLED). Right now though, we would not allow the account to be globally verified if an email or phone were provided.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: It didn't seem like we needed to prevent updates to this value after account creation, because it isn't validated (if it's wrong, you just can't sign in to that study with your Synapse account until it's fixed).

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Something like this:
```
    public AccountStatus getStatus() {
        if (status == AccountStatus.DISABLED) {
            return AccountStatus.DISABLED;
        }
        boolean verified = false;
        if (Boolean.TRUE.equals(emailVerified)) {
            verified = true;
        }
        if (Boolean.TRUE.equals(phoneVerified)) {
            verified = true;
        }
        return (verified) ? AccountStatus.ENABLED : AccountStatus.UNVERIFIED;
    }
```

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
DwayneJengSage: I think part of the confusion is that it’s not clear what the AccountStatus means. Back when Email was the only way to access Bridge, it was to prevent you from using Bridge before you verify your email, because a lot of things could use your email address. Now that we’ve split off email verification, the semantics is a little bit unclear.

From a purely technical perspective, UNVERIFIED means you can’t access Bridge until you’ve verified your email/phone number. If someone, for whatever reason, creates a Bridge account with both an email address and a Synpase ID, do we want to block them from using Bridge until they’ve verified that email address? Or do we want to allow them access to Bridge, but not send them any emails until they’ve verified their email address?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
DwayneJengSage: The reason why that exists is to force users to pass in their credentials in order to change their email or phone number, to prevent someone from stealing your session, hijacking your account, and then using that to take over your account permanently.

I can see this not being an issue with externalId, because externalId is an arbitrary string, and there is no external ID equivalent of reset password email or email sign-in.

With Synapse User ID, it’s a little trickier, since if the user doesn’t have a password, you can’t ask for a password to change the Synapse User ID it’s associated with. That said, I still think the risk of someone hijacking the Bridge account is high enough that we should prevent this from happening. It would be safer to block updating Synapse User ID altogether, except through an admin API. I’ll check with Aaron.

alxdarksage: This is moved to the updateIdentifiers API.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Filed as https://sagebionetworks.jira.com/browse/BRIDGE-2628. I'll email Bridge group about this change today.

src/test/java/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -314,6 +314,7 @@ public static void assertValidatorMessage(Validator validator, Object object, St
             Validate.entityThrowingException(validator, object);
             fail("Should have thrown exception");
         } catch(InvalidEntityException e) {
+            System.out.println(e.getErrors());
```
DwayneJengSage: Minor: Remove println

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -866,6 +880,10 @@ public StudyParticipant updateIdentifiers(Study study, CriteriaContext context,
             sendEmailVerification = true;
             accountUpdated = true;
         }
+        if (update.getSynapseUserIdUpdate() != null && account.getSynapseUserId() == null) {
+            account.setSynapseUserId(update.getSynapseUserIdUpdate());
+            accountUpdated = true;
+        }
```
DwayneJengSage, line 887: Should there be a unit test for updating Synapse ID via updateIdentifiers()?

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantControllerTest.java
```diff
@@ -1260,11 +1263,32 @@ public void updateIdentifiersWithExternalId() throws Exception {
                 identifierUpdateCaptor.capture());
 
         IdentifierUpdate update = identifierUpdateCaptor.getValue();
-        assertEquals(update.getSignIn().getPhone(), PHONE_PASSWORD_SIGN_IN_REQUEST.getPhone());
+        assertEquals(update.getSignIn().getEmail(), PHONE_PASSWORD_SIGN_IN_REQUEST.getEmail());
```
DwayneJengSage: Was this change intentional?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantControllerTest.java
alxdarksage: No, reverted

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/TestUtils.java
alxdarksage: Removed

DwayneJengSage (approved):

PR #122, opened 2019-11-05
Synapse OAuth support

alxdarksage:


src/main/java/org/sagebionetworks/bridge/services/OAuthProviderService.java
```diff
@@ -136,6 +177,72 @@ protected DateTime getDateTime() {
...
+     * an authentication token, returned with the userid. The information required is very similar to an 
+     * OAuth provider as configured by a study administrator, but it's internal to our system and so it is 
+     * configured like other components of the server software.
+     *   
+     * @param authToken that was passed from the Synapse server back to the authenticating client
+     * @return synapseUserId if the exchange is successful
+     * 
+     * @throws BadRequestException
+     */
+    public String oauthSignIn(OAuthAuthorizationToken authToken) {
```
DwayneJengSage: Minor: This API is specific to Synapse. Should the API name reflect that?

src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -491,12 +496,38 @@ protected String generateReauthToken() {
...
+        }
+        
+        clearSession(authToken.getStudyId(), account.getId());
+        Study study = studyService.getStudy(authToken.getStudyId());
+        UserSession session = getSessionFromAccount(study, context, account);
+
+        cacheProvider.setUserSession(session);
+        if (!session.doesConsent() && !session.isInRole(Roles.ADMINISTRATIVE_ROLES)) {
+            throw new ConsentRequiredException(session);
+        }
```
DwayneJengSage: Is the consent check necessary? We don't intend for participants to sign-in with Synapse. (In fact, we explicitly do _not_ want to do that.) Would it be better to check for Admin privileges and throw a 403 if they're missing?

src/test/java/org/sagebionetworks/bridge/services/OAuthProviderServiceTest.java
```diff
@@ -332,4 +376,72 @@ public void introspect_ErrorCallingIntrospect() throws Exception {
...
+    @Test
+    public void oauthSignIn() throws Exception {
+        // This is not encrypted, the real token is public/private key encrypted. We mock the parser
+        // to avoid having to sign the payload.
+        mockAccessGrantCall(201, "{\"access_token\":\"not used\",\"id_token\":{\"userid\":\"12345\"}}");
+        
+        when(service.getJwtParser()).thenReturn(mockJwtParser);
+        when(mockJwtParser.parseClaimsJws(any())).thenReturn(mockJwtClaims);
+        when(mockJwtClaims.getBody()).thenReturn(mockClaims);
+        when(mockClaims.get(SYNAPSE_USERID_KEY, String.class)).thenReturn("12345");
```
DwayneJengSage, line 390: For testing purposes, let's make the "unencrypted" value different from the "encrypted" value to make sure our code is plumbing the values through correctly.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -291,6 +292,26 @@ public StatusMessage resetPassword() throws Exception {
...
+        
+        UserSession session = null;
+        try {
+            session = authenticationService.oauthSignIn(context, token);
+        } catch(ConsentRequiredException e) {
+            setCookieAndRecordMetrics(e.getUserSession());
+            throw e;
+        }
+        setCookieAndRecordMetrics(session);
+        verifySupportedVersionOrThrowException(study);
```
DwayneJengSage: Similarly, I don't think it's necessary to include the shut-off switch, since we aren't intending Synapse sign-in for participants.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -291,6 +292,26 @@ public StatusMessage resetPassword() throws Exception {
...
+        Study study = studyService.getStudy(token.getStudyId());
+        CriteriaContext context = getCriteriaContext(study.getStudyIdentifier());
+        
+        UserSession session = null;
+        try {
+            session = authenticationService.oauthSignIn(context, token);
+        } catch(ConsentRequiredException e) {
+            setCookieAndRecordMetrics(e.getUserSession());
+            throw e;
+        }
```
DwayneJengSage: Similarly, we don't need to check consent.

src/main/resources/BridgeServer2.conf
```diff
@@ -176,3 +176,7 @@ api.developer.password = dummy-value
 
 shared.developer.email = dummy-value
 shared.developer.password = dummy-value
+
+synapse.oauth.url = https://repo-prod.prod.sagebase.org/auth/v1/oauth2/token
+synapse.oauth.client.id = dummy-value
+synapse.oauth.client.secret = dummy-value
```
DwayneJengSage, line 182: Where do we get these values so that I can (a) configure my local server and (b) so we can add these to the BridgeServer2-infra project?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/OAuthProviderService.java
alxdarksage: We pass in a vendor ID so in theory, this could be used with another 3rd party provider like Google. (At that point though, we'll have to lookup the configuration using the vendor ID.)

src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: A much better plan. Changed this to check for roles and no more consent check.

src/test/java/org/sagebionetworks/bridge/services/OAuthProviderServiceTest.java
alxdarksage: Changed

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: Removing. I think setCookieAndRecordMetrics should be moved into the request interceptor so that even when we throw RuntimeExceptions, we log the request correctly. We shouldn't have to catch like this all over the place. I will file a JIRA.

alxdarksage: https://sagebionetworks.jira.com/browse/BRIDGE-2632

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: Removed

src/main/resources/BridgeServer2.conf
alxdarksage: Explanation of these values added to the Bridge bootstrap document, the values are checked into LastPass, and I've created a PR to add the values for Elastic Beanstalk deployments.

src/main/java/org/sagebionetworks/bridge/services/OAuthProviderService.java
alxdarksage: Okay the method how returns an AccountId. In the future, a different vendor ID could be used to retrieve some other identifier besides the Synapse ID (email is a logical choice), and an AnccountId can be returned with those credentials. So the method is better encapsulated for further vendors, and I've updated the documentation to reflect this.

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -1433,5 +1428,59 @@ public void languagePreferencesArePersisted() {
...
+   @Test(expectedExceptions = EntityNotFoundException.class)
+   public void oauthSignInNotFoundNoToken() {
+       OAuthAuthorizationToken token = new OAuthAuthorizationToken(TEST_STUDY_IDENTIFIER, "vendorId",
+               "authToken", "callbackUrl");
+       
+       service.oauthSignIn(CONTEXT, token);
+   }
+   
+   @Test(expectedExceptions = UnauthorizedException.class)
+   public void oauthSignInUnconsented() {
```
DwayneJengSage: Minor: change test name

DwayneJengSage (approved):
One minor comment about test name. Otherwise looks good.

DwayneJengSage (changes requested):
I think we should hold off on checking this in until we resolve the OAuth API issue.

My concerns are 2-fold:

1. We've built a feature that's untestable, and given that we (Sage) own both Bridge and Synapse, there's no good reason for this to be untestable.
2. Simply integrating Bridge and Synapse at the web level doesn't meet our goals. We need to integrate them at the API level.

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: Changed.

DwayneJengSage (approved):

PR #124, opened 2019-11-15
Change study apis

alxdarksage:
Two new APIs for https://sagebionetworks.jira.com/browse/BRIDGE-2624:

- The cross-study admin API to switch studies without signing out has been expanded for all administrative users (it also continues to support cross-study admins, because given the information in the user session today, a client cannot know who is a cross-study admin once they change studies);

- There is a new API to return the studies that a user is a member of. As with the change study API, for cross-study admins this is all studies; for other users, it is every study that has the same Synapse user ID as the signed in user.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -10,7 +10,6 @@ import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -25,7 +24,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
 
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.BridgeUtils.SubstudyAssociations;
@@ -89,32 +87,27 @@ public class HibernateAccountDao implements AccountDao {
     }
     
     @Override
-    public Set<String> getStudyIdsForUser(Study study, String userId) {
-        AccountId accountId = AccountId.forId(study.getIdentifier(), userId);
+    public List<String> getStudyIdsForUser(StudyIdentifier studyId, String userId) {
+        AccountId accountId = AccountId.forId(studyId.getIdentifier(), userId);
         Account account = getAccount(accountId);
         if (account == null) {
             throw new EntityNotFoundException(Account.class);
         }
-        
-        QueryBuilder builder = new QueryBuilder();
-        builder.append("SELECT DISTINCT acct.studyId FROM HibernateAccount AS acct WHERE ");
-        
         boolean hasEmail = (account.getEmail() != null && TRUE.equals(account.getEmailVerified()));
         
+        QueryBuilder builder = new QueryBuilder();
+        builder.append("SELECT DISTINCT acct.studyId FROM HibernateAccount AS acct WHERE");
         if (hasEmail) {
             builder.append("email = :email", "email", account.getEmail());
         }
         if (hasEmail && account.getSynapseUserId() != null) {
-            builder.append(" OR ");
+            builder.append("OR");
         }
         if (account.getSynapseUserId() != null) {
             builder.append("synapseUserId = :synapseUserId", "synapseUserId", account.getSynapseUserId());
         }
-        
-        System.out.println(builder.getQuery());
-        
-        List<String> studyIds = hibernateHelper.queryGet(builder.getQuery(), builder.getParameters(), 0, 1000, String.class); 
-        return ImmutableSet.copyOf(studyIds);
+        System.out.println(builder.getParameters());
+        return hibernateHelper.queryGet(builder.getQuery(), builder.getParameters(), null, null, String.class); 
     }
     
     /**
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -10,6 +10,7 @@ import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -24,6 +25,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.BridgeUtils.SubstudyAssociations;
@@ -84,6 +86,35 @@ public class HibernateAccountDao implements AccountDao {
     // Provided to override in tests
     protected String generateGUID() {
         return BridgeUtils.generateGuid();
+    }
+    
+    @Override
+    public Set<String> getStudyIdsForUser(Study study, String userId) {
+        AccountId accountId = AccountId.forId(study.getIdentifier(), userId);
+        Account account = getAccount(accountId);
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
+        
+        QueryBuilder builder = new QueryBuilder();
+        builder.append("SELECT DISTINCT acct.studyId FROM HibernateAccount AS acct WHERE ");
+        
+        boolean hasEmail = (account.getEmail() != null && TRUE.equals(account.getEmailVerified()));
+        
+        if (hasEmail) {
+            builder.append("email = :email", "email", account.getEmail());
+        }
+        if (hasEmail && account.getSynapseUserId() != null) {
+            builder.append(" OR ");
+        }
+        if (account.getSynapseUserId() != null) {
+            builder.append("synapseUserId = :synapseUserId", "synapseUserId", account.getSynapseUserId());
+        }
+        
+        System.out.println(builder.getQuery());
+        
+        List<String> studyIds = hibernateHelper.queryGet(builder.getQuery(), builder.getParameters(), 0, 1000, String.class); 
+        return ImmutableSet.copyOf(studyIds);
     }
     
     /**
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
@@ -339,8 +339,8 @@ public class HibernateAccountDao implements AccountDao {
         
         AccountId unguarded = accountId.getUnguardedAccountId();
         if (unguarded.getId() != null) {
-            // Now that we allow study switching, we must enforce study membership here
-            // as well as in the queries below.
+            // Now that we allow study switching, we must enforce study membership here in the getById()
+            // method as well as in the queries below.
             Account account = hibernateHelper.getById(HibernateAccount.class, unguarded.getId());
             if (account != null && !account.getStudyId().equals(accountId.getStudyId())) {
                 return null;
--- a/src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
@@ -164,6 +164,9 @@ public class AuthenticationService {
         checkNotNull(context);
 
         Account account = accountDao.getAccount(context.getAccountId());
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
         return getSessionFromAccount(study, context, account);
     }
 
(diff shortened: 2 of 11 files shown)


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -86,6 +88,18 @@ protected String generateGUID() {
...
     
+    @Override
+    public List<String> getStudyIdsForUser(String synapseUserId) {
+        if (isBlank(synapseUserId)) {
+            return ImmutableList.of();
+        }
+        QueryBuilder query = new QueryBuilder();
+        query.append(
+            "SELECT DISTINCT acct.studyId FROM HibernateAccount AS acct WHERE synapseUserId = :synapseUserId",
+            "synapseUserId", synapseUserId);
```
DwayneJengSage, line 99: We don't have an index on synapseUserId. Should there be one?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -86,6 +88,18 @@ protected String generateGUID() {
         return BridgeUtils.generateGuid();
     }
     
+    @Override
+    public List<String> getStudyIdsForUser(String synapseUserId) {
+        if (isBlank(synapseUserId)) {
+            return ImmutableList.of();
+        }
```
DwayneJengSage, line 95: We don't normally have precondition checks in DAOs. Would it be better to have the callers verify that synapseUserId is specified?

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -1797,6 +1797,30 @@ public void editAccountFailsAcrossSubstudies() throws Exception {
         verify(mockHibernateHelper, never()).update(any(), eq(null));
         BridgeUtils.setRequestContext(null);
     }
+    
+    @Test
+    public void getStudyIdsForUser() throws Exception {
+        HibernateAccount hibernateAccount = makeValidHibernateAccount(false);
+        hibernateAccount.setSynapseUserId(SYNAPSE_USER_ID);
+        when(mockHibernateHelper.getById(HibernateAccount.class, ACCOUNT_ID)).thenReturn(hibernateAccount);
```
DwayneJengSage: Does this API call getById()?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -325,7 +339,13 @@ private Account getHibernateAccount(AccountId accountId) {
         
         AccountId unguarded = accountId.getUnguardedAccountId();
         if (unguarded.getId() != null) {
-            return hibernateHelper.getById(HibernateAccount.class, unguarded.getId());
+            // Now that we allow study switching, we must enforce study membership here in the getById()
+            // method as well as in the queries below.
+            Account account = hibernateHelper.getById(HibernateAccount.class, unguarded.getId());
+            if (account != null && !account.getStudyId().equals(accountId.getStudyId())) {
+                return null;
+            }
```
DwayneJengSage, line 347: Unit tests? (Null account case and mismatched studyId case.)

src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -170,6 +170,9 @@ public UserSession getSession(Study study, CriteriaContext context) {
         checkNotNull(context);
 
         Account account = accountDao.getAccount(context.getAccountId());
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
```
DwayneJengSage, line 175: Unit test?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -293,6 +299,59 @@ public StatusMessage resetPassword() throws Exception {
...
+            throw new BadRequestException("Account has not been assigned a Synapse user ID");
+        }
+        
+        // Retrieve the desired study
+        SignIn signIn = parseJson(SignIn.class);
+        String targetStudyId = signIn.getStudyId();
+        Study targetStudy = studyService.getStudy(targetStudyId);
+
+        // These are the accounts cross-study that are linked through the Synapse User ID
+        List<String> studyIds = accountDao.getStudyIdsForUser(participant.getSynapseUserId());
```
DwayneJengSage: I don't believe AuthenticationController has access to accountDao. Also, it probably shouldn't. We should probably go through the Service layer.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -1165,7 +1182,125 @@ public void unconsentedReauthSetsMetrics() throws Exception {
...
+    
+    @Test
+    public void changeStudySupportsCrossStudyAdmin() throws Exception {
+        mockRequestBody(mockRequest, new SignIn.Builder().withStudy("my-new-study").build());
+        userSession.setParticipant(new StudyParticipant.Builder().withSynapseUserId(SYNAPSE_USER_ID)
+                .withRoles(ImmutableSet.of(ADMIN)).build());
+        doReturn(userSession).when(controller).getAuthenticatedSession();
+        
+        when(mockAccountDao.getStudyIdsForUser(SYNAPSE_USER_ID))
+            .thenReturn(ImmutableList.of("my-new-study", TEST_STUDY_IDENTIFIER));
```
DwayneJengSage: In practice, admin accounts would only have TEST_STUDY_IDENTIFIER show up in their study list.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -1165,7 +1182,125 @@ public void unconsentedReauthSetsMetrics() throws Exception {
...
+        
+        Account account = Account.create();
+        AccountId accountId = AccountId.forSynapseUserId("my-new-study", SYNAPSE_USER_ID);
+        when(mockAccountDao.getAccount(accountId)).thenReturn(account);
```
DwayneJengSage: In practice, this would return null for admin accounts.

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -159,16 +165,37 @@ public String getAllStudies(@RequestParam(required = false) String format,
...
     }
+    
+    @GetMapping(path="/v3/studies/memberships", produces={APPLICATION_JSON_UTF8_VALUE})
+    public String getStudyMemberships() throws Exception {   
+        UserSession session = getAuthenticatedSession();
+        
+        if (session.getParticipant().getRoles().isEmpty()) {
+            throw new UnauthorizedException(STUDY_ACCESS_EXCEPTION_MSG);
+        }
+        List<String> studyIds = accountDao.getStudyIdsForUser(session.getParticipant().getSynapseUserId());
```
DwayneJengSage, line 185: Similarly, this should also go through the service.

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -159,16 +165,37 @@ public String getAllStudies(@RequestParam(required = false) String format,
...
+        Stream<Study> stream = studyService.getStudies().stream();
```
DwayneJengSage: We have 30-40 studies in production. Most users will only have access to one. At most, users might have access to a handful (for example, a Sage employee might have access to mPower, CRF, PKU, and Psoriasis). It seems inefficient to do a table scan of all studies just to filter out the majority of them.

It would be better to just iterate on the studyIds and call getStudy() instead. (And special case ADMIN as needed.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
DwayneJengSage: Is it necessary to get all study IDs for the user when we already know which study ID they want? Would it be easier to just get the account for the specific study ID and and synapse user ID? (And a second call to getAccount for the api study in the ADMIN special case?)

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyControllerTest.java
```diff
@@ -799,6 +803,88 @@ public void deleteStudy() throws Exception {
...
+                .withEmailVerified(true)
+                .withRoles(ImmutableSet.of(DEVELOPER))
+                .withSynapseUserId(SYNAPSE_USER_ID).build();
+        when(mockSession.getParticipant()).thenReturn(participant);
+        doReturn(mockSession).when(controller).getAuthenticatedSession();
+
+        Study studyA = createStudy("Study A", "studyA", false);
+        Study studyB = createStudy("Study B", "studyB", true);
+        Study studyC = createStudy("Study C", "studyC", true);
+        Study studyD = createStudy("Study D", "studyD", true);
```
DwayneJengSage: Maybe order the studies in a random order to verify that we are properly sorting the studies? (Here and in the next test as well.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/UserManagementController.java
```diff
@@ -71,6 +71,14 @@ public JsonNode signInForAdmin() throws Exception {
...
     }
     
+    /**
+     * This turned out to be useful... so useful we're opening it up to all administrative
+     * users.
+     * 
+     * @see org.sagebionetworks.bridge.spring.controllersAuthenticationController#changeStudy 
+     * @throws Exception
+     */
+    @Deprecated
```
DwayneJengSage, line 81: AuthenticationController.changeStudy requires synapseUserId. This API does not.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Yes. Added to changelog.sql.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Because of the way the Account classes evolved, we have no AccountService. So the AccountDao ends up doing some of the things a service class would do. It's on my long list of things to refactor. I could check this in the two controllers that call it, but then it would be duplicated in both controllers.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
alxdarksage: No, there's code here for an earlier implementation that wasn't removed. I've removed the lines that are no longer relevant.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: Added mismatched study, null case is handled by getByIdNotFound().

src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: There is no AccountService, so indeed AccountDao is currently referenced in BaseController.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
alxdarksage: This is true. I've removed the other ID which is a stronger test (we specifically don't want the list to match the target study).

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
alxdarksage: True, adjusted to return null.

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
alxdarksage: Filed a JIRA to refactor out a service class: https://sagebionetworks.jira.com/browse/BRIDGE-2636

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
alxdarksage: Yes this is my bad, I thought that call was cached but only the full getStudy() call is cached. Changed this.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: I've changed this to attempt loads of accounts (in API or the target study) to determine eligibility to switch studies.

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyControllerTest.java
alxdarksage: I shuffled these around to verify they are being sorted.

src/main/java/org/sagebionetworks/bridge/spring/controllers/UserManagementController.java
alxdarksage: That's a good point. I moved a couple of checks around and now it is not required for this role, and updated the relevant changeStudy test to verify.

DwayneJengSage (approved):

PR #125, opened 2019-11-18
Migrate Surveys to SQL

janaechan:
https://sagebionetworks.jira.com/browse/BRIDGE-1127


src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
```diff
@@ -0,0 +1,50 @@
...
+    void setType(String type);
+    
+    JsonNode getData();
+    void setData(JsonNode data);
+    
+    JsonNode getBeforeRules();
+    void setBeforeRules(JsonNode beforeRules);
+    
+    JsonNode getAfterRules();
+    void setAfterRules(JsonNode afterRules);
```
DwayneJengSage: Is there any reason these can't be SurveyRules?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.List;
+
+import org.sagebionetworks.bridge.hibernate.HibernateSurvey;
+import org.sagebionetworks.bridge.hibernate.HibernateSurveyElement;
+import org.sagebionetworks.bridge.json.JsonNodeToSurveyElementConverter;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+@JsonDeserialize(as=HibernateSurveyElement.class)
+public interface SurveyElementSQL extends BridgeEntity {
```
DwayneJengSage: This class looks identical to SurveyElement, except before/afterRules is a JsonNode instead of SurveyRules. Is there any reason why this needs to be a separate interface? Why not have HibernateSurveyElement and DynamoSurveyElement both share the same interface?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveySQL.java
```diff
@@ -0,0 +1,86 @@
...
+     * Gets the upload schema revision that corresponds to this survey. See
+     * {@link org.sagebionetworks.bridge.models.upload.UploadSchema#getRevision} for more details.
+     */
+    Integer getSchemaRevision();
+
+    /** @see #getSchemaRevision */
+    void setSchemaRevision(Integer schemaRevision);
+    
+    List<HibernateSurveyElement> getElements();
+    void setElements(List<HibernateSurveyElement> elements);
```
DwayneJengSage: Does this have to be HibernateSurveyElement? Why not make it SurveyElement? If for whatever reason, if the implementing class needs to say HibernateSurveyElement, why not make it `List<? extends SurveyElement>` instead?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveySQL.java
```diff
@@ -0,0 +1,86 @@
...
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolder;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.common.collect.ImmutableList;
+
+@BridgeTypeName("Survey")
+@JsonDeserialize(as=HibernateSurvey.class)
+public interface SurveySQL extends GuidCreatedOnVersionHolder, BridgeEntity  {
```
DwayneJengSage: Similarly, this class is also nearly identical to Survey. Why not use the same interface for both?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
```diff
@@ -0,0 +1,50 @@
...
+    String GUID_PROPERTY = "guid";
+    String IDENTIFIER_PROPERTY = "identifier";
+    String IMAGE_PROPERTY = "image";
+    String PROMPT_DETAIL_PROPERTY = "promptDetail";
+    String PROMPT_PROPERTY = "prompt";
+    String BEFORE_RULES_PROPERTY = "beforeRules";
+    String AFTER_RULES_PROPERTY = "afterRules";
+    String TITLE_PROPERTY = "title";
+    String TYPE_PROPERTY = "type";    
+    String UI_HINTS_PROPERTY = "uiHint";
```
DwayneJengSage: We should move this to a shared utility class instead of duplicating them. (Similarly in all its subclasses.)

src/main/java/org/sagebionetworks/bridge/dao/SurveyDaoSQL.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.List;
+import java.util.function.Consumer;
+
+import org.sagebionetworks.bridge.models.GuidCreatedOnVersionHolder;
+import org.sagebionetworks.bridge.models.PagedResourceList;
+import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
+import org.sagebionetworks.bridge.models.surveys.SurveySQL;
+
+public interface SurveyDaoSQL {
```
DwayneJengSage: This interface is identical to SurveyDao. Why not use the same interface for both?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurvey.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+    
+    /** @see #getElements */
+    public void setElements(List<HibernateSurveyElement> elements) {
+        this.elements = elements;
+    }
+    
+    public boolean keysEqual(GuidCreatedOnVersionHolder keys) {
+        return (keys != null && keys.getGuid().equals(guid) && keys.getCreatedOn() == createdOn);
+    }
```
DwayneJengSage: This is identical in both DynamoSurvey and HibernateSurvey. Why not move it to the interface as a default implementation?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyQuestion.java
```diff
@@ -0,0 +1,135 @@
...
+        setType(entry.getType());
+        setIdentifier(entry.getIdentifier());
+        setGuid(entry.getGuid());
+        setData(entry.getData());
+        setBeforeRules(entry.getBeforeRules());
+        setAfterRules(entry.getAfterRules());
+    }
+    
+    @Override
+    @DynamoDBIgnore
```
DwayneJengSage: Hibernate classes shouldn't have DynamoDB annotations.

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyQuestionSQL.java
```diff
@@ -0,0 +1,54 @@
...
+import org.sagebionetworks.bridge.hibernate.HibernateSurveyQuestion;
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+import org.sagebionetworks.bridge.json.JsonUtils;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(as = DynamoSurveyQuestion.class)
+@BridgeTypeName("SurveyQuestion")
+public interface SurveyQuestionSQL extends SurveyElementSQL {
```
DwayneJengSage: Similarly, why does this need to be its own interface?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
janaechan: The type is specified as mediumtext in the sql schema so I thought they have to be JsonNode?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveySQL.java
janaechan: I can't have two create methods to create both DynamoSurvey() and HibernateSurvey() so I thought I'd need a new interface for this? And also the interfaces deserialize as different classes

janaechan:
This way I wouldn't need the new interfaces right?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
DwayneJengSage: It means you store it in the database as JSON, but you can use the Hibernate converters to convert it into whatever class you need. See for example RequestInfos.

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveySQL.java
DwayneJengSage: When we call create(), it's because we don't care what the underlying implementation is, so it doesn't matter whether we return a DynamoSurvey or a HibernateSurvey. Similarly, when we deserialize from JSON.

If we get into a situation where we _do_ care (such as the DynamoSurveyDao needing to save a survey to Dynamo), we could use the copy constructors to convert them. (In particular, it looks like SurveyElementFactory.fromDynamoEntity() doesn't create things _from_ DynamoSurveyElement, but rather it converts things _to_ DynamoSurveyElement.)

src/main/java/org/sagebionetworks/bridge/dao/SurveyDaoSQL.java
DwayneJengSage: We should delete all of these extra interfaces.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurvey.java
```diff
@@ -0,0 +1,230 @@
...
+        this.copyrightNotice = survey.getCopyrightNotice();
+        this.moduleId = survey.getModuleId();
+        this.moduleVersion = survey.getModuleVersion();
+        this.version = survey.getVersion();
+        this.name = survey.getName();
+        this.identifier = survey.getIdentifier();
+        this.published = survey.isPublished();
+        this.deleted = survey.isDeleted();
+        this.schemaRevision = survey.getSchemaRevision();
+        this.elements = survey.getElements();
```
DwayneJengSage: The Survey might be a DynamoSurvey, in which case the survey elements would be DynamoSurveyElements. We'll need a converter.

Also, what about converting SurveyElements to SurveyQuestions and SurveyInfoScreens?

src/main/java/org/sagebionetworks/bridge/models/surveys/Survey.java
```diff
@@ -16,7 +18,11 @@
     static Survey create() {
         return new DynamoSurvey();
     }
-
+    
+    static Survey createHibernate() {
+        return new HibernateSurvey();
+    }
```
DwayneJengSage: Doesn't need this. If a class cares specifically about the Hibernate implementation (eg the HibernateSurveyDao), it can create a HibernateSurvey directly.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyInfoScreen.java
```diff
@@ -0,0 +1,96 @@
...
+        return prompt;
+    }
+
+    @Override
+    public void setPrompt(String prompt) {
+        this.prompt = prompt;
+    }
+
+    @Override
+    @DynamoDBIgnore
```
DwayneJengSage: Hibernate class doesn't need Dynamo annotations.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dao/SurveyDaoSQL.java
DwayneJengSage: We should delete this interface.

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementSQL.java
DwayneJengSage: We should delete this interface.

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementFactory.java
```diff
@@ -36,5 +42,28 @@ public static SurveyElement fromDynamoEntity(SurveyElement element) {
...
+    public static List<SurveyElement> fromElementList(List<SurveyElement> elements) {
+        ImmutableList.Builder<SurveyElement> builder = new ImmutableList.Builder<>();
+        for (SurveyElement element : elements) {
+            if (element.getType().equals(SURVEY_QUESTION_TYPE)) {
+                builder.add(new HibernateSurveyQuestion(element));
+            } else if (element.getType().equals(SURVEY_INFO_SCREEN_TYPE)) {
+                builder.add(new HibernateSurveyInfoScreen(element));
+            } else {
+                throw new InvalidEntityException("Survey element type '"+element.getType()+"' not recognized.");
+            }
```
DwayneJengSage: Can this call fromHibernateEntity()?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
```diff
@@ -0,0 +1,114 @@
...
+        return surveyCompoundKey;
+    }
+    
+    public void setSurveyCompoundKey(String surveyCompoundKey) {
+        this.surveyCompoundKey = surveyCompoundKey;
+    }
+    
+    public void setSurveyKeyComponents(String surveyGuid, long createdOn) {
+        this.surveyCompoundKey = surveyGuid + ":" + Long.toString(createdOn);
+    }
```
DwayneJengSage: It looks like these are only called by the Dynamo implementations. Would it be cleaner to just delete them from the interface?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
janaechan: Removing them would require casting to DynamoSurveyElements everywhere the these methods are called and adding them to SurveyQuestion and SurveyInfoScreen. Should I follow through with the refractor?

DwayneJengSage: The alternative would be to make sure get/setSurveyCompoundKey work as intended, which means that we can derive the surveyCompoundKey from the surveyGuid/createdOn and setting the surveyCompoundKey also sets the surveyGuid/createdOn.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
```diff
@@ -0,0 +1,114 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableList;
+
+@Entity
+@Table(name = "SurveyElements")
+public class HibernateSurveyElement implements SurveyElement {
+    @Id
+    private String guid;
+    private String surveyGuid;
```
DwayneJengSage, line 22: Which reminds me, SurveyElement needs to have createdOn in addition to surveyGuid.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
janaechan: Can we assume the given surveyCompoundKey when setting it is formatted correctly so that I can split on ":" and set surveyGuid/createdOn?

DwayneJengSage: I don't think we can.

janaechan: I'll add a check for that then. Also, this means setting the surveyGuid or createdOn also sets the surveyCompoundKey?

DwayneJengSage: Yes, I think it should.

That said, you might consider not storing surveyCompoundKey and just generating it on the fly whenever the getter is called.

janaechan: Should I throw some error if the surveyCompoundKey is formatted incorrectly?

DwayneJengSage:
Looks good so far. Let me know when you have the DAO and the tests and stuff.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+public class HibernateSurveyDao implements SurveyDao {
+    private static final String IDENTIFIER_PREFIX = "identifier:";
+    Integer OFFSET = 0;
+    Integer PAGESIZE = 1;
```
DwayneJengSage: It's probably not worth making constants for these, since they are very specific to getting a single element, and it actually makes the code less readable if I have to reference what OFFSET and PAGESIZE are.

Also, HibernateHelper.queryGet() takes in nulls for offset and limit, so you can pass in offset null instead of 0.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+    private static final String GET_ACTIVE_PUBLISH_WITH_IDENTIFIER = "FROM HibernateSurvey as survey " + 
+            "WHERE studyKey = :studyKey AND identifier = :identifier AND deleted = 0 AND published = :published ORDER BY createdOn DESC";
+    
+    private static final String GET_ACTIVE_WITH_IDENTIFIER = "FROM HibernateSurvey as survey " + 
+            "WHERE studyKey = :studyKey AND identifier = :identifier AND createdOn = :createdOn AND deleted = 0 ORDER BY createdOn DESC";
+    
+    private static final String GET_ELEMENTS = "FROM HibernateSurveyElement as surveyElement " + 
+            "WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn ORDER BY createdOn DESC";
+    
+    private static final String DELETE_SURVEY_ELEMENTS = "DELETE FROM HibernateSurveyElement WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn";
```
DwayneJengSage: A lot of these query fragments are used exactly once. At this point, it might be easier to just use hardcoded strings inside each method to make it easier to read.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+    }
+    
+    @Autowired
+    public final void setUploadSchemaService(UploadSchemaService uploadSchemaService) {
+        this.uploadSchemaService = uploadSchemaService;
+    }
+    
+    @Override
+    public Survey createSurvey(Survey survey) {
+        System.out.println("createSurvey");
```
DwayneJengSage: Minor: remove unused printlns.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            survey.setGuid(generateGuid());
+        }
+        long time = DateUtils.getCurrentMillisFromEpoch();
+        survey.setCreatedOn(time);
+        survey.setModifiedOn(time);
+        survey.setSchemaRevision(null);
+        survey.setPublished(false);
+        survey.setDeleted(false);
+        survey.setVersion(null);
+        hibernateHelper.create(survey, null);
```
DwayneJengSage: DynamoSurveyDao calls saveSurvey(), which handles a lot of logic for deleting orphan survey elements, setting the guid, survey guid, survey createdOn, and order on the survey elements, and reconciling survey rules. Is this something we need to replicate to the HibernateSurveyDao? Alternatively, move it to SurveyService?

Similarly for updateSurvey() and versionSurvey().

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+    public Survey updateSurvey(StudyIdentifier studyIdentifier, Survey survey) {
+        Survey existingSurvey = getSurvey(studyIdentifier, survey, true);
+        
+        if (existingSurvey != null) {
+            hibernateHelper.update(survey, null);
+            return survey;
+        } else {
+            return createSurvey(survey);
+        }
+    }
```
DwayneJengSage: In DynamoSurveyDao, updateSurvey() could only be used to update existing surveys (never to create surveys), and you could only update specific fields.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+
+    @Override
+    public Survey versionSurvey(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
+        Survey existingSurvey = getSurvey(studyIdentifier, keys, false);
+        System.out.println(existingSurvey);
+        HibernateSurvey copy = new HibernateSurvey(existingSurvey);
```
DwayneJengSage, line 100: The DynamoSurveyDao implementation also cleared the published and deleted flags and the version number, and also set the createdOn and modifiedOn timestamps.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        if (!survey.isPublished()) {
+            // update survey
+            survey.setPublished(true);
+            survey.setModifiedOn(DateUtils.getCurrentMillisFromEpoch());
+            // make schema from survey
+            if (!survey.getUnmodifiableQuestionList().isEmpty()) {
+                UploadSchema schema = uploadSchemaService.createUploadSchemaFromSurvey(studyIdentifier, survey, newSchemaRev);
+                survey.setSchemaRevision(schema.getRevision());
+            }
+            updateSurvey(studyIdentifier, survey);
```
DwayneJengSage: The DynamoSurveyDao implementation doesn't call updateSurvey(), because updateSurvey() can only update specific fields of the survey.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            List<HibernateSurvey> results = hibernateHelper.queryGet(
+                    getQuery, params, OFFSET, PAGESIZE, HibernateSurvey.class);
+            
+            if (!results.isEmpty()) {
+                return results.get(0);
+            }
+            return null;
+        }
+        SurveyId surveyId = new SurveyId(keys);
+        Survey ret = hibernateHelper.getById(HibernateSurvey.class, surveyId);
```
DwayneJengSage, line 175: Do we need to filter out deleted surveys?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        if (ret != null && includeElements) {
+            attachSurveyElements(ret);
+        }
```
DwayneJengSage: The reason we had the includeElements flag in DynamoSurveyDao is that querying the second table was very expensive. In SQL, it's just another query, and Hibernate handles it for us, so we no longer need that flag. Also, since we already fetched survey elements, it's probably unnecessary to query for survey elements a second time.

That said, the survey is probably just the raw survey elements, and it would be useful to have a helper method to convert those into the subclasses and to reconcile the survey rules.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        List<? extends HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, OFFSET, PAGESIZE, HibernateSurvey.class);
+        if (!results.isEmpty()) {
+            return results.get(0).getGuid();
+        }
+        return null;
+    }
+
+    @Override
+    public List<Survey> getSurveyAllVersions(StudyIdentifier studyIdentifier, String guid, boolean includeDeleted) {
```
DwayneJengSage, line 198: This API can also get by identifier.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        ImmutableMap<String, Object> params = ImmutableMap.of(
+                "studyKey", studyIdentifier.getIdentifier(),
+                "guid", guid);
+        
+        String getQuery = SELECT_TEMPLATE + ((!includeDeleted) ? GET_ACTIVE : GET_ALL);
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: The DynamoSurveyDao implementation didn't specify a page size, even though it probably should have. However, because it didn't, we don't have a pagination API for surveys, so we need to return all of them. HibernateHelper.queryGet() can take in null for the page size, meaning no page size.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        
+        return ImmutableList.copyOf(results);
+    }
+
+    @Override
+    public Survey getSurveyMostRecentVersion(StudyIdentifier studyIdentifier, String guid) {
```
DwayneJengSage, line 222: This API can also get by identifier.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        ImmutableMap<String, Object> params = ImmutableMap.of(
+                "studyKey", studyIdentifier.getIdentifier(),
+                "guid", guid);
+        
+        String getQuery = SELECT_TEMPLATE + GET_ACTIVE;
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: This API only gets 1 survey, so there's no need to ask Hibernate for 50 of them.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            "WHERE studyKey = :studyKey AND deleted = 0 AND published = 1 ORDER BY createdOn DESC";
+    
+    private static final String GET_ACTIVE = "FROM HibernateSurvey as survey " + 
+            "WHERE studyKey = :studyKey AND guid = :guid AND deleted = 0 ORDER BY createdOn DESC";
+    
+    private static final String GET_ACTIVE_PUBLISH = "FROM HibernateSurvey as survey " + 
+            "WHERE studyKey = :studyKey AND guid = :guid AND deleted = 0 AND published = :published ORDER BY createdOn DESC";
+    
+    private static final String GET_ACTIVE_PUBLISH_WITH_IDENTIFIER = "FROM HibernateSurvey as survey " + 
+            "WHERE studyKey = :studyKey AND identifier = :identifier AND deleted = 0 AND published = :published ORDER BY createdOn DESC";
```
DwayneJengSage: These two queries only ever ask for published = 1.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            }
+        } else {
+            ImmutableMap<String, Object> params = ImmutableMap.of(
+                    "studyKey", studyIdentifier.getIdentifier(),
+                    "guid", guid,
+                    "published", 1);
+            
+            String getQuery = SELECT_TEMPLATE + GET_ACTIVE_PUBLISH;
+            List<HibernateSurvey> results = hibernateHelper.queryGet(
+                    getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: This API only gets 1 survey, so there's no need to ask Hibernate for 50 of them.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            if (!results.isEmpty()) {
+                ImmutableMap<String, Object> elementParams = ImmutableMap.of(
+                        "surveyGuid", guid,
+                        "createdOn", results.get(0).getCreatedOn());
+                
+                String getElementQuery = SELECT_ELEMENT_TEMPLATE + GET_ELEMENTS;
+                
+                List<HibernateSurveyElement> elementResults = hibernateHelper.queryGet(
+                        getElementQuery, elementParams, OFFSET, 50, HibernateSurveyElement.class);
+//                results.get(0).setElements(elementResults);
```
DwayneJengSage: This should be consolidated into attachSurveyElements().

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            if (!results.isEmpty()) {
+                ImmutableMap<String, Object> elementParams = ImmutableMap.of(
+                        "surveyGuid", results.get(0).getGuid(),
+                        "createdOn", results.get(0).getCreatedOn());
+                
+                String getElementQuery = SELECT_ELEMENT_TEMPLATE + GET_ELEMENTS;
+                
+                List<HibernateSurveyElement> elementResults = hibernateHelper.queryGet(
+                        getElementQuery, elementParams, OFFSET, 50, HibernateSurveyElement.class);
+//                results.get(0).setElements(elementResults);
```
DwayneJengSage: This should be consolidated into attachSurveyElements().

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+            boolean includeElements) {
+        if (guid.toLowerCase().startsWith(IDENTIFIER_PREFIX)) {
+            ImmutableMap<String, Object> params = ImmutableMap.of(
+                    "studyKey", studyIdentifier.getIdentifier(),
+                    "identifier", guid.substring(IDENTIFIER_PREFIX.length()),
+                    "published", 1);
+            
+            String getQuery = SELECT_TEMPLATE + GET_ACTIVE_PUBLISH_WITH_IDENTIFIER;
+            List<HibernateSurvey> results = hibernateHelper.queryGet(
+                    getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: This API only gets 1 survey, so there's no need to ask Hibernate for 50 of them.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    @Override
+    public List<Survey> getAllSurveysMostRecentlyPublishedVersion(StudyIdentifier studyIdentifier,
+            boolean includeDeleted) {
+        ImmutableMap<String, Object> params = ImmutableMap.of(
+                "studyKey", studyIdentifier.getIdentifier());
+        
+//        String countQuery = SELECT_COUNT + GET_ACTIVE;
+        String getQuery = SELECT_TEMPLATE + GET_ALL_ACTIVE_PUBLISH;
```
DwayneJengSage: We need to check the includeDeleted flag.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
DwayneJengSage: For the queries that get one survey guid or identifier, we want to sort on createdOn desc.

For the queries that get all surveys in a study, we want to sort on survey guid (or identifier), and then sort on createdOn desc.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: Similarly, we don't want to ask for a max page size here.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+        Map<String, Survey> map = Maps.newLinkedHashMap();
+        for (Survey survey : surveys) {
+            Survey stored = map.get(survey.getGuid());
+            if (stored == null || survey.getCreatedOn() > stored.getCreatedOn()) {
+                map.put(survey.getGuid(), survey);
+            }
+        }
+        return ImmutableList.copyOf(map.values());
+    }
```
DwayneJengSage, line 323: It's possible to write a query to do this and let SQL do the work for us. That said, if it's easier to just do this, then doing this is fine.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        
+        return findMostRecentVersions(results);
+    }
+
+    @Override
+    public List<Survey> getAllSurveysMostRecentVersion(StudyIdentifier studyIdentifier, boolean includeDeleted) {
+        ImmutableMap<String, Object> params = ImmutableMap.of(
+                "studyKey", studyIdentifier.getIdentifier());
+        
+        String getQuery = SELECT_TEMPLATE + GET_ALL_SURVEYS;
```
DwayneJengSage: We need to check the includeDeleted flag.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, OFFSET, 50, HibernateSurvey.class);
```
DwayneJengSage: Similarly, we don't want to ask for a max page size here.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        survey.setDeleted(true);
+        hibernateHelper.update(survey, null);
+    }
+    
+    @Override
+    public void deleteSurveyPermanently(StudyIdentifier studyIdentifier, GuidCreatedOnVersionHolder keys) {
+        Survey existing = getSurvey(studyIdentifier, keys, false);
+        
+        if (existing != null) {
+            deleteAllElements(existing.getGuid(), existing.getCreatedOn());
```
DwayneJengSage: If we use "on cascade delete" for the foreign keys, is it really necessary to explicitly delete the survey elements before deleting the survey?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+    
+    private void attachSurveyElements(Survey survey) {
+        ImmutableMap<String, Object> elementParams = ImmutableMap.of(
+                "surveyGuid", survey.getGuid(),
+                "createdOn", survey.getCreatedOn());
+        
+        String getElementQuery = SELECT_ELEMENT_TEMPLATE + GET_ELEMENTS;
+        
+        List<HibernateSurveyElement> elementResults = hibernateHelper.queryGet(
+                getElementQuery, elementParams, OFFSET, 50, HibernateSurveyElement.class);
```
DwayneJengSage: In the off-chance there are more than 50 survey elements, we want to return all of them.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,365 @@
...
+        
+        List<SurveyElement> elements = Lists.newArrayList();
+        for (HibernateSurveyElement element : elementResults) {
+            SurveyElement surveyElement = SurveyElementFactory.fromHibernateEntity(element);
+//            reconcileRules(surveyElement);
```
DwayneJengSage: Need reconcileRules() implementation.

DwayneJengSage (changes requested):

DwayneJengSage:
Also, migration plan?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+        for (HibernateSurveyElement element : elementResults) {
+            SurveyElement surveyElement = SurveyElementFactory.fromHibernateEntity(element);
+            reconcileRules(surveyElement);
+            elements.add((SurveyElement) surveyElement);
+        }
+        survey.setElements(elements);
+    }
+    
+    private Survey saveSurvey(Survey survey) {
+        deleteAllElements(survey.getGuid(), survey.getCreatedOn());
```
DwayneJengSage, line 362: It might be worth testing to see how Hibernate handles if if you change the survey elements in the element collection. Does Hibernate automatically delete orphaned survey elements for you? If so, we probably don't need to manually delete all survey elements ourselves.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+            element.setOrder(i);
+            if (element.getGuid() == null) {
+                element.setGuid(generateGuid());
+            }
+            reconcileRules(element);
+            hibernateElements.add((HibernateSurveyElement) element);
+        }
+        
+//        List<FailedBatch> failures = surveyElementMapper.batchSave(dynamoElements);
+//        BridgeUtils.ifFailuresThrowException(failures);
```
DwayneJengSage: Minor: Delete commented code.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+            // that point they will either not be copied on top of element rules which exist, or both 
+            // element and constraint rules will be empty, so it makes no difference.
+            Constraints con = question.getConstraints();
+            if (BridgeUtils.isEmpty(question.getAfterRules())) {
+                question.setAfterRules( con.getRules() );
+            }
+            // question rules take precedence once they exist.
+            con.setRules(question.getAfterRules());
+        }
+    }
```
DwayneJengSage: This method is identical between DynamoSurveyDao and HibernateSurveyDao. I wonder if it's worth moving it to the interface, or even just turning it into a static utility method (possibly in a new SurveyUtil class).

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+        hibernateHelper.query(getQuery, params);
+    }
+    
+    private void attachSurveyElements(Survey survey) {
+        ImmutableMap<String, Object> elementParams = ImmutableMap.of(
+                "surveyGuid", survey.getGuid(),
+                "createdOn", survey.getCreatedOn());
+        
+        String getElementQuery = "SELECT surveyElement FROM HibernateSurveyElement as surveyElement" + 
+                "WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn ORDER BY createdOn DESC";
```
DwayneJengSage: This should be ordered by order asc, not createdOn.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+            getQuery += "AND identifier = :identifier ";
+        } else {
+            paramBuilder.put("guid", guid);
+            
+            getQuery += "AND guid = :guid ";
+        }
+        getQuery += ORDER_BY_CREATED_ON;
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, paramBuilder.build(), null, null, HibernateSurvey.class);
```
DwayneJengSage: This API only returns 1 survey. No need to ask Hibernate for all of them.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,405 @@
...
+        if (!results.isEmpty()) {
+            attachSurveyElements(results.get(0));
+        }
+        
+        return null;
```
DwayneJengSage: Should be a return statement inside the if block.

DwayneJengSage (changes requested):

DwayneJengSage:
Also, migration plan?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
janaechan: Figuring out how to test this

DwayneJengSage: Create survey with some survey elements, update the survey with different survey elements, see if you have orphaned survey elements left behind in the database.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,378 @@
...
+            SurveyElement surveyElement = SurveyElementFactory.fromHibernateEntity(element);
+            reconcileRules(surveyElement);
+            elements.add((SurveyElement) surveyElement);
+        }
+        survey.setElements(elements);
+        return survey;
+    }
+    
+    private Survey saveSurvey(Survey survey) {
+        deleteAllElements(survey.getGuid(), survey.getCreatedOn());
```
DwayneJengSage, line 362: Reminder note to test if Hibernate handles orphaned survey elements.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyTest.java
```diff
@@ -0,0 +1,43 @@
...
+    
+//    @Test
+//    public void canSerialize() {
+//        HibernateSurvey survey = new HibernateSurvey();
+//        survey.setStudyIdentifier(TEST_STUDY_IDENTIFIER);
+//        HibernateSurveyElement element = new HibernateSurveyElement();
+//        survey.getElements().add(element);
+//        
+//        
+//    }
```
DwayneJengSage, line 42: Reminder to fill in serialize tests. Also, copy constructor tests.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+    ArgumentCaptor<SurveyId> surveyIdCaptor;
+    
+    @BeforeMethod
+    public void beforeMethod() {
+        MockitoAnnotations.initMocks(this);
+        GuidCreatedOnVersionHolder holder = new GuidCreatedOnVersionHolderImpl(GUID, CREATED_ON);
+        SurveyId surveyId = new SurveyId(holder);
+        
+        doReturn(GUID).when(dao).generateGuid();
+        DateTimeUtils.setCurrentMillisFixed(TIMESTAMP);
```
DwayneJengSage, line 101: You should use an After annotation to call DateTimeUtils.setCurrentMillisSystem() to unmock the current time.

Consider using a BeforeClass and AfterClass so you only need to mock/unmock this once.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        Survey result = dao.createSurvey(survey);
+        assertSame(survey, result);
+        assertEquals(result.getGuid(), GUID);
+        assertEquals(result.getCreatedOn(), TIMESTAMP);
+        assertEquals(result.getModifiedOn(), TIMESTAMP);
+        assertNull(result.getSchemaRevision());
+        assertFalse(result.isPublished());
+        assertFalse(result.isDeleted());
+        assertNull(result.getVersion());
+        assertEquals(result.getElements().get(0).getGuid(), GUID);
```
DwayneJengSage, line 118: Should also check the survey guid and created on on the survey element, and also order.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+    
+    @Test
+    public void updateSurvey() {
+        HibernateSurvey survey = getHibernateSurvey();
+        // Some fields are copied over to the existing survey because they are mutable.
+        // verify this.
+        survey.setName("Updated Name");
+        survey.setCopyrightNotice("Updated Copyright Notice");
+        survey.setDeleted(true);
+        survey.setVersion(3L);
```
DwayneJengSage, line 137: Should we also include elements (and test that we automatically copy over the same attributes as create)?

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        assertFalse(result.isDeleted());
+        assertNull(result.getVersion());
+        assertEquals(result.getElements().get(0).getGuid(), GUID);
+        
+        verify(mockHelper).query(queryCaptor.capture(), paramsCaptor.capture());
+        assertEquals(queryCaptor.getValue(), "DELETE FROM HibernateSurveyElement WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn");
+        assertEquals(paramsCaptor.getValue().get("surveyGuid"), GUID);
+        assertEquals(paramsCaptor.getValue().get("createdOn"), TIMESTAMP);
+        
+        verify(mockHelper).update(surveyCaptor.capture(), any());
```
DwayneJengSage, line 125: Should be null to verify that we don't send an afterPersistConsumer. (Same with all the verify calls to mockHelper.update().)

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        HibernateSurvey survey = new HibernateSurvey();
+        // Mess up some things to verify they are changed by versioning
+        survey.setGuid(GUID);
+        survey.setPublished(true);
+        survey.setDeleted(true);
+        survey.setVersion(100L);
+        survey.setCreatedOn(0L);
+        survey.setModifiedOn(0L);
+        survey.setSchemaRevision(2);
+        survey.setElements(ImmutableList.of());
```
DwayneJengSage, line 177: Add some elements and verify they are copied over

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        assertEquals(updated.getModifiedOn(), TIMESTAMP);
+        assertNull(updated.getSchemaRevision());
+    }
+    
+    @Test
+    public void publishSurveySchemaRevision() {
+        HibernateSurvey survey = getHibernateSurvey();
+        
+        when(mockHelper.getById(eq(HibernateSurvey.class), any())).thenReturn(survey);
+        when(mockHelper.queryGet(any(), any(), any(), any(), eq(SurveyElement.class))).thenReturn(ELEMENTS);
```
DwayneJengSage, line 210: I don't think publish() calls mockHelper.queryGet()

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+    private static final String GUID = "oneGuid";
+    private static final long CREATED_ON = DateTime.now(DateTimeZone.UTC).getMillis();
+    private static final long TIMESTAMP = DateTime.now().getMillis();
+    private static final String IDENTIFIER = "surveyId";
+    private static final String IDENTIFIER_PREFIX = "identifier:";
+    private static final GuidCreatedOnVersionHolder KEYS = new GuidCreatedOnVersionHolderImpl(GUID, CREATED_ON);
+    static final GuidCreatedOnVersionHolder SURVEY_IDENTIFIER_KEYS = new GuidCreatedOnVersionHolderImpl(IDENTIFIER_PREFIX+IDENTIFIER, CREATED_ON);
+    private static final SurveyId surveyId = new SurveyId(KEYS);
+    private static final List<SurveyElement> ELEMENTS = ImmutableList.of(getSurveyQuestion(), getSurveyInfoScreen());
+    private static final List<HibernateSurveyElement> HIBERNATE_ELEMENTS = ImmutableList.of((HibernateSurveyQuestion)getSurveyQuestion(), (HibernateSurveyInfoScreen)getSurveyInfoScreen());
```
DwayneJengSage, line 70: The casts don't do what you think they do. getSurveyQuestion() and getSurveyInfoScreen() return HibernateSurveyQuestion and HibnernateSurveyInfoScreen, but mockHelper.queryGet() returns the parent class HibernateSurveyElement.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,378 @@
...
+    private Survey attachSurveyElements(Survey survey) {
+        ImmutableMap<String, Object> elementParams = ImmutableMap.of(
+                "surveyGuid", survey.getGuid(),
+                "createdOn", survey.getCreatedOn());
+        
+        String getElementQuery = "SELECT surveyElement FROM HibernateSurveyElement as surveyElement " + 
+                "WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn ORDER BY order ASC";
+        
+        List<HibernateSurveyElement> elementResults = hibernateHelper.queryGet(
+                getElementQuery, elementParams, null, null, HibernateSurveyElement.class);
```
DwayneJengSage, line 349: I think Hibernate automatically fills in survey.getElements() with survey elements, so we don't need to query them again. We just need to convert them.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        hibernateSurvey.setGuid(GUID);
+        hibernateSurvey.setStudyIdentifier(TEST_STUDY_IDENTIFIER);
+        hibernateSurvey.setIdentifier(IDENTIFIER);
+        hibernateSurvey.setCreatedOn(CREATED_ON);
+        hibernateSurvey.setElements(ELEMENTS);
+        hibernateSurvey.setPublished(false);
+        hibernateSurvey.setSchemaRevision(1);
+        hibernateSurvey.setVersion(1L);
+        
+        return hibernateSurvey;
```
DwayneJengSage, line 669: I think we need two different getHibernateSurveys() (or at least a boolean flag to control it). One for surveys returned by Hibernate (which will return HibernateSurveyElements) and one that's passed in by callers to our API (which will include HibernateSurveyQuestions and HibernateSurveyInfoScreens).

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        HibernateSurvey existing = getHibernateSurvey();
+        
+        when(mockHelper.getById(eq(HibernateSurvey.class), any())).thenReturn(existing);
+        when(mockHelper.queryGet(any(), any(), any(), any(), eq(HibernateSurveyElement.class)))
+            .thenReturn(HIBERNATE_ELEMENTS);
+        
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(GUID, CREATED_ON);
+        Survey survey = dao.getSurvey(TEST_STUDY, keys, true);
+        assertEquals(survey.getIdentifier(), IDENTIFIER);
+        assertEquals(survey.getElements().size(), 2);
```
DwayneJengSage, line 332: We should also verify that we convert the survey elements (same as every other API that calls attachSurveyElements())

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        when(mockHelper.queryGet(any(), any(), any(), any(), eq(HibernateSurvey.class))).thenReturn(ImmutableList.of(existing));
+        when(mockHelper.queryGet(any(), any(), any(), any(), eq(HibernateSurveyElement.class))).thenReturn(HIBERNATE_ELEMENTS);
+        
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(IDENTIFIER_PREFIX+IDENTIFIER, CREATED_ON);
+        Survey survey = dao.getSurvey(TEST_STUDY, keys, true);
+        assertEquals(survey.getIdentifier(), IDENTIFIER);
+        assertEquals(survey.getElements().size(), 2);
+        assertSame(existing, survey);
+        
+        verify(mockHelper).queryGet(queryCaptor.capture(),paramsCaptor.capture(), any(), any(), eq(HibernateSurvey.class));
```
DwayneJengSage, line 360: We pass in null for offset and 1 for page size. The test should reflect that. (Same with all the other calls to queryGet()

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,378 @@
...
+        SurveyId surveyId = new SurveyId(keys);
+        Survey ret = hibernateHelper.getById(HibernateSurvey.class, surveyId);
+        if (ret != null) {
+            return attachSurveyElements(ret);
+        }
+        return null;
+    }
+
+    @Override
+    public String getSurveyGuidForIdentifier(StudyIdentifier studyIdentifier, String surveyId) {
```
DwayneJengSage, line 183: Add unit test for getSurveyGuidForIdentifier()

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+    
+    @Test
+    public void getSurveyNotFoundWithIdentifier() {
+        GuidCreatedOnVersionHolder keys = new GuidCreatedOnVersionHolderImpl(IDENTIFIER_PREFIX+IDENTIFIER, CREATED_ON);
+        
+        assertNull(dao.getSurvey(TEST_STUDY, keys, true));
+    }
+    
+    @Test
+    public void getSurveyAllVersionsIncludeDeleted() {
```
DwayneJengSage, line 378: What about the exclude deleted version of this test? And the exclude deleted version of the test with identifier?

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDaoTest.java
```diff
@@ -0,0 +1,706 @@
...
+        String surveyElementQuery = queryCaptor.getAllValues().get(1);
+        
+        assertEquals(surveyQuery, "SELECT survey FROM HibernateSurvey as survey WHERE studyKey = :studyKey " + 
+                "AND deleted = 0 AND published = 1 AND guid = :guid ORDER BY createdOn DESC");
+        assertEquals(surveyElementQuery, "SELECT surveyElement FROM HibernateSurveyElement as surveyElement " + 
+                "WHERE surveyGuid = :surveyGuid AND createdOn = :createdOn ORDER BY order ASC");
+    }
+    
+    @Test
+    public void getAllSurveysMostRecentlyPublishedVersionIncludeDeleted() {
```
DwayneJengSage, line 463: What about the exclude deleted version of this test?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyDao.java
```diff
@@ -0,0 +1,378 @@
...
+                ((!includeDeleted) ? "AND deleted = 0 " : "") + ORDER_BY_GUID_CREATED_ON;
+        
+        List<HibernateSurvey> results = hibernateHelper.queryGet(
+                getQuery, params, null, null, HibernateSurvey.class);
+        
+        return findMostRecentVersions(results);
+    }
+
+    @Override
+    public List<Survey> getAllSurveysMostRecentVersion(StudyIdentifier studyIdentifier, boolean includeDeleted) {
```
DwayneJengSage, line 292: Unit test for this method?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
```diff
@@ -0,0 +1,122 @@
...
+    public void setSurveyCompoundKey(String surveyCompoundKey) {
+        if (surveyCompoundKey.contains(":")) {
+            setSurveyKeyComponents(surveyCompoundKey.split(":")[0], Long.valueOf(surveyCompoundKey.split(":")[1]));
+        }
+    }
+    
+    public void setSurveyKeyComponents(String surveyGuid, long createdOn) {
+        this.surveyGuid = surveyGuid;
+        this.createdOn = createdOn;
+    }
```
DwayneJengSage, line 121: Unit tests? In particular, test that setting the compound key sets the guid/createdOn, and setting the guid/createdOn sets the compound key.

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyElementFactory.java
```diff
@@ -36,5 +42,22 @@ public static SurveyElement fromDynamoEntity(SurveyElement element) {
...
+        }
+    }
+    
+    public static List<SurveyElement> fromElementList(List<SurveyElement> elements) {
+        ImmutableList.Builder<SurveyElement> builder = new ImmutableList.Builder<>();
+        for (SurveyElement element : elements) {
+            builder.add(fromHibernateEntity(element));
+        }
+        return builder.build();
+    }
```
DwayneJengSage, line 62: Unit tests for these 2 methods?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSurveyElement.java
```diff
@@ -0,0 +1,122 @@
...
+
+import org.sagebionetworks.bridge.models.surveys.SurveyElement;
+import org.sagebionetworks.bridge.models.surveys.SurveyRule;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableList;
+
+@Entity
+@Table(name = "SurveyElements")
+public class HibernateSurveyElement implements SurveyElement {
```
alxdarksage, line 19: You are using single table polymorphic mapping in hibernate (all subclasses of survey element are stored in the same table), but where do you tell Hibernate how to store this, or are you using defaults? If so, where do you tell Hibernate that the discriminator column's name is "type" and not "DTYPE" (the default)? See https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#entity-inheritance-single-table

DwayneJengSage:
Surveys are deprecated.

PR #128, opened 2019-11-22
Mark a survey option as exclusive (NOTA)

alxdarksage:
The survey engine will deselect all other options when an exclusive option is checked, and probably disable them as well. It is used for a "none of the above" option which is not the same as an "other" option.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOption.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOption.java
@@ -12,18 +12,21 @@ public final class SurveyQuestionOption {
     private final String detail;
     private final String value;
     private final Image image;
+    private final Boolean exclusive;
     
     @JsonCreator
     public SurveyQuestionOption(@JsonProperty("label") String label, @JsonProperty("detail") String detail,
-        @JsonProperty("value") String value, @JsonProperty("image") Image image) {
+            @JsonProperty("value") String value, @JsonProperty("image") Image image,
+            @JsonProperty("exclusive") Boolean exclusive) {
         this.label = label;
         this.detail = detail;
         this.value = value;
         this.image = image;
+        this.exclusive = exclusive;
     }
     
     public SurveyQuestionOption(String label) {
-        this(label, null, label, null);
+        this(label, null, label, null, null);
     }
     
     public String getLabel() {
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOption.java
```diff
@@ -12,18 +12,21 @@
     private final String detail;
     private final String value;
     private final Image image;
+    private final Boolean exclusive;
     
     @JsonCreator
     public SurveyQuestionOption(@JsonProperty("label") String label, @JsonProperty("detail") String detail,
-        @JsonProperty("value") String value, @JsonProperty("image") Image image) {
+            @JsonProperty("value") String value, @JsonProperty("image") Image image,
+            @JsonProperty("exclusive") Boolean exclusive) {
```
DwayneJengSage, line 20: I noticed there isn't a JSON serialization test in SurveyQuestionOptionTest. Should we add one?

src/test/java/org/sagebionetworks/bridge/TestSurvey.java
```diff
@@ -62,10 +62,11 @@ public static SurveyQuestion selectBy(Survey survey, DataType type) {
...
             List<SurveyQuestionOption> options = Lists.newArrayList(
-                    new SurveyQuestionOption("Terrible", null, "1", terrible),
-                    new SurveyQuestionOption("Poor", null, "2", poor), new SurveyQuestionOption("OK", null, "3", ok),
-                    new SurveyQuestionOption("Good", null, "4", good),
-                    new SurveyQuestionOption("Great", null, "5", great));
+                    new SurveyQuestionOption("Terrible", null, "1", terrible, null),
+                    new SurveyQuestionOption("Poor", null, "2", poor, null), 
+                    new SurveyQuestionOption("OK", null, "3", ok, null),
+                    new SurveyQuestionOption("Good", null, "4", good, null),
+                    new SurveyQuestionOption("Great", null, "5", great, null));
```
DwayneJengSage, line 69: Do we expect to add more features to SurveyQuestionOption over time? If so, this would cause a lot of churn as we'd have to update a lot of constructor calls. Is it worth converting to a Builder pattern at this time?

src/main/java/org/sagebionetworks/bridge/models/surveys/SurveyQuestionOption.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/TestSurvey.java
alxdarksage: This is the first addition in a few years and I don't have any further requests to add to this. I don't think it's worth creating a builder.

DwayneJengSage (approved):

PR #129, opened 2019-11-25
Introduce the Superadmin role separate from admin as a study-scoped admin

alxdarksage:


src/main/java/org/sagebionetworks/bridge/Roles.java
```diff
@@ -10,23 +10,25 @@
...
     public static final Map<Roles,EnumSet<Roles>> CAN_BE_EDITED_BY = new ImmutableMap.Builder<Roles, EnumSet<Roles>>()
-        .put(ADMIN, EnumSet.of(ADMIN))
-        .put(WORKER, EnumSet.of(ADMIN))
-        .put(RESEARCHER, EnumSet.of(ADMIN, RESEARCHER))
-        .put(DEVELOPER, EnumSet.of(ADMIN, RESEARCHER, DEVELOPER))
+        .put(SUPERADMIN, EnumSet.of(SUPERADMIN))
+        .put(ADMIN, EnumSet.of(SUPERADMIN, ADMIN))
+        .put(WORKER, EnumSet.of(SUPERADMIN, ADMIN))
+        .put(RESEARCHER, EnumSet.of(SUPERADMIN, ADMIN, RESEARCHER))
+        .put(DEVELOPER, EnumSet.of(SUPERADMIN, ADMIN, RESEARCHER, DEVELOPER))
```
DwayneJengSage, line 32: Do we want to apply the change mentioned in https://sagebionetworks.jira.com/browse/BRIDGE-2627? Or will that be a separate checkin?

src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
```diff
@@ -97,10 +97,12 @@ public boolean hasSignedMostRecentConsent() {
...
     }
     public boolean isInRole(Roles role) {
-        return (role != null && participant.getRoles().contains(role));
+        Set<Roles> proles = participant.getRoles();
+        return (role != null && (proles.contains(SUPERADMIN) || proles.contains(role)));
     }
     public boolean isInRole(Set<Roles> roleSet) {
-        return roleSet != null && !Collections.disjoint(participant.getRoles(), roleSet);
+        Set<Roles> proles = participant.getRoles();
+        return roleSet != null && (proles.contains(SUPERADMIN) || roleSet.stream().anyMatch(role -> isInRole(role)));
```
DwayneJengSage, line 105: Unit tests?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -533,7 +534,7 @@ public void updateParticipant(Study study, StudyParticipant participant) {
         // Allow admin and worker accounts to toggle status; in particular, to disable/enable accounts.
         if (participant.getStatus() != null) {
             Set<Roles> callerRoles = BridgeUtils.getRequestContext().getCallerRoles();
-            if (callerRoles.contains(ADMIN) || callerRoles.contains(WORKER)) {
+            if (callerRoles.contains(ADMIN) || callerRoles.contains(WORKER) || callerRoles.contains(SUPERADMIN)) {
```
DwayneJengSage, line 537: Unit test?

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -297,7 +299,7 @@ public Study createStudyAndUsers(StudyAndUsers studyAndUsers) throws SynapseExce
 
         // validate roles for each user
         for (StudyParticipant user: users) {
-            if (!Collections.disjoint(user.getRoles(), ImmutableSet.of(Roles.ADMIN, Roles.WORKER))) {
+            if (!Collections.disjoint(user.getRoles(), ImmutableSet.of(ADMIN, WORKER, SUPERADMIN))) {
```
DwayneJengSage, line 302: Unit test?

src/main/java/org/sagebionetworks/bridge/spring/controllers/SubstudyController.java
```diff
@@ -49,22 +50,22 @@ final void setSubstudyService(SubstudyService substudyService) {
     @PostMapping("/v3/substudies")
     @ResponseStatus(HttpStatus.CREATED)
     public VersionHolder createSubstudy() {
-        UserSession session = getAuthenticatedSession(ADMIN);
+        UserSession session = getAuthenticatedSession(SUPERADMIN);
```
DwayneJengSage, line 53: Previously, study admins could manage substudies. Is it intentional that only superadmins can manage substudies?

src/main/resources/db/changelog/changelog.sql
```diff
@@ -261,3 +261,9 @@ ADD UNIQUE KEY `Accounts-StudyId-SynapseUserId-Index` (`studyId`,`synapseUserId`
 -- changeset bridge:11
 
 CREATE INDEX `Accounts-SynapseUserId-Index` ON `Accounts`(`synapseUserId`);
+
+-- changeset bridge:12
+
+ALTER TABLE `AccountRoles`
+MODIFY COLUMN `role` enum('DEVELOPER','RESEARCHER','ADMIN','WORKER','SUPERADMIN') NOT NULL;
```
DwayneJengSage: A brief Google search suggests that some SQL implementations as ordinals (ie, the integer index into the enum values): http://www.mysqltutorial.org/mysql-enum/

If we're deleting tests_users and shifting worker over, there's a chance we might be accidentally elevating all worker accounts to superadmin accounts. (Which wouldn't be the end of the world, since all worker accounts are owned by us, but it's probably not desirable.)

Would it be easier to just keep test_users around?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/Roles.java
alxdarksage: It should be separate because we'll need to adjust the BridgeStudyManager and our integration tests might break as well (I doubt it but it's possible).

src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added (these tests will also all have to change when we change the security model).

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/SubstudyController.java
alxdarksage: Yes, these should be created by us (as super admins) as they represent organizations and their access to the users in the app.

src/main/resources/db/changelog/changelog.sql
alxdarksage: Easy enough, I've put it back. I have not verified that it will not break existing worker accounts in my local environment.

DwayneJengSage (approved):

PR #131, opened 2019-12-05
Remove creation of Synapse user accounts when creating a study

alxdarksage:
Part of switching to the user of Synapse for administrative accounts, is that we now ask for Synapse accounts up front and then we add them to the Bridge and Synapse models we create. So this PR removes the assumption that we're creating a Synapse account, and invites existing users into the new access team that we are creating.

Also I took the opportunity to refactor a bunch of BadRequestExceptions into a StudyAndUsersValidator.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
+++ b/src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
@@ -11,6 +11,8 @@ import java.util.List;
 import com.google.common.collect.ImmutableSet;
 
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
@@ -22,11 +24,13 @@ import org.sagebionetworks.client.SynapseClient;
 import org.sagebionetworks.client.exceptions.SynapseException;
 import org.sagebionetworks.client.exceptions.SynapseNotFoundException;
 
+@Component
 public class StudyAndUsersValidator implements Validator {
 
-    private final SynapseClient synapseClient;
+    private SynapseClient synapseClient;
 
-    public StudyAndUsersValidator(SynapseClient synapseClient) {
+    @Autowired
+    public final void setSynapseClient(SynapseClient synapseClient) {
         this.synapseClient = synapseClient;
     }
     
--- a/src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
+++ b/src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
@@ -1,0 +1,112 @@
+package org.sagebionetworks.bridge.validators;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.sagebionetworks.bridge.Roles.ADMIN;
+import static org.sagebionetworks.bridge.Roles.SUPERADMIN;
+import static org.sagebionetworks.bridge.Roles.WORKER;
+
+import java.util.Collections;
+import java.util.List;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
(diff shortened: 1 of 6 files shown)


src/test/java/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
```diff
@@ -940,25 +941,23 @@ public void createStudyAndUser() throws SynapseException {
 
         // stub
         when(mockParticipantService.createParticipant(any(), any(), anyBoolean())).thenReturn(mockIdentifierHolder);
-        doNothing().when(mockSynapseClient).newAccountEmailValidation(any(), any());
 
         // execute
         service.createStudyAndUsers(mockStudyAndUsers);
 
         // verify
-        verify(mockParticipantService, times(2)).createParticipant(any(), any(), anyBoolean());
```
DwayneJengSage, line 949: Don't we still call createParticipant()?

src/test/java/org/sagebionetworks/bridge/TestUtils.java
```diff
@@ -314,6 +314,7 @@ public static void assertValidatorMessage(Validator validator, Object object, St
             Validate.entityThrowingException(validator, object);
             fail("Should have thrown exception");
         } catch(InvalidEntityException e) {
+            System.out.println(e.getErrors());
```
DwayneJengSage: Minor: Remove print.

src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
```diff
@@ -0,0 +1,112 @@
...
+
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.models.studies.StudyAndUsers;
+import org.sagebionetworks.client.SynapseClient;
+import org.sagebionetworks.client.exceptions.SynapseException;
+import org.sagebionetworks.client.exceptions.SynapseNotFoundException;
+
+public class StudyAndUsersValidator implements Validator {
```
DwayneJengSage, line 28: I wonder if we can use Spring to create this as a singleton component and autowire it with the SynapseClient, instead of recreating it on the fly every time?

src/test/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.testng.annotations.Test;
+
+import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
+import org.sagebionetworks.bridge.models.studies.Study;
+import org.sagebionetworks.bridge.models.studies.StudyAndUsers;
+import org.sagebionetworks.client.SynapseClient;
+import org.sagebionetworks.client.exceptions.SynapseException;
+import org.sagebionetworks.client.exceptions.SynapseNotFoundException;
+
+public class StudyAndUsersValidatorTest extends Mockito {
```
DwayneJengSage, line 30: How about a test case with a valid StudyAndUsers?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/StudyServiceMockTest.java
alxdarksage: The next two lines verify the specific createParticipant calls, so this seems redundant.

src/test/java/org/sagebionetworks/bridge/TestUtils.java
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidator.java
alxdarksage: Yes, switched this.

src/test/java/org/sagebionetworks/bridge/validators/StudyAndUsersValidatorTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #132, opened 2019-12-09
Account dao refactor

alxdarksage:
In this first step, an AccountService is introduced that forwards everything from the existing AccountDao. This changes a bunch of references from the DAO to the service in a mechanical way that should make review easier... the actual refactor will come in a second PR.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -1,0 +1,147 @@
+package org.sagebionetworks.bridge.services;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.dao.AccountDao;
+import org.sagebionetworks.bridge.models.AccountSummarySearch;
+import org.sagebionetworks.bridge.models.PagedResourceList;
+import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
+import org.sagebionetworks.bridge.models.accounts.AccountSummary;
+import org.sagebionetworks.bridge.models.accounts.SignIn;
+import org.sagebionetworks.bridge.models.studies.Study;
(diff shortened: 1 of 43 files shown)


src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -0,0 +1,167 @@
...
+    @Test
+    public void getHealthCodeForAccount() {
+        Account account = Account.create();
+        account.setHealthCode(HEALTH_CODE);
+        when(mockAccountDao.getAccount(ACCOUNT_ID)).thenReturn(account);
+        
+        String healthCode = service.getHealthCodeForAccount(ACCOUNT_ID);
+        assertEquals(healthCode, HEALTH_CODE);
+        verify(mockAccountDao).getAccount(ACCOUNT_ID);
+    }
```
DwayneJengSage, line 167: We should also have a unit test for the null case.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -0,0 +1,147 @@
...
+     * For MailChimp, and other external systems, we need a way to get a healthCode for a given email.
+     */
+    public String getHealthCodeForAccount(AccountId accountId) {
+        Account account = getAccount(accountId);
+        if (account != null) {
+            return account.getHealthCode();
+        } else {
+            return null;
+        }
+    }
```
DwayneJengSage, line 146: Is it intentional that we're re-implementing getHealthCodeForAccount() instead of calling through to the DAO?

DwayneJengSage (commented):
Overall looks good. 2 comments about getHealthCodeForAccount().

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: No, removed from the DAO.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #133, opened 2019-12-10
[BRIDGE-2452] redirect http to https

zaro0508:
Attempt to setup http to https redirect. This is unlikely to work
since beanstalk does not provide a reference to the ALB.
This also moves all ALB configs from the cloudformation
template to ebextensions.


zaro0508:
replaced with https://github.com/Sage-Bionetworks/BridgeServer2-infra/pull/30 which does work

PR #134, opened 2019-12-12
Apply superadmin privileges to role checks on the server

alxdarksage:
Fix for https://sagebionetworks.jira.com/browse/BRIDGE-2649

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
@@ -1,11 +1,11 @@
 package org.sagebionetworks.bridge.models.accounts;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static org.sagebionetworks.bridge.Roles.SUPERADMIN;
 
 import java.util.Map;
 import java.util.Set;
 
+import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.Roles;
 import org.sagebionetworks.bridge.config.Environment;
 import org.sagebionetworks.bridge.models.studies.StudyIdentifier;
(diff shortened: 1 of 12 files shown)


src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
```diff
@@ -101,8 +101,7 @@ public boolean isInRole(Roles role) {
         return (role != null && (proles.contains(SUPERADMIN) || proles.contains(role)));
     }
     public boolean isInRole(Set<Roles> roleSet) {
-        Set<Roles> proles = participant.getRoles();
-        return roleSet != null && (proles.contains(SUPERADMIN) || roleSet.stream().anyMatch(role -> isInRole(role)));
+        return roleSet != null && roleSet.stream().anyMatch(role -> isInRole(role));
```
DwayneJengSage: It doesn't look like we have unit tests for SUPERADMIN matching everything in UserSessionTest. Is it worth adding those?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -634,14 +629,15 @@ private void updateAccountAndRoles(Study study, Account account, StudyParticipan
             String value = participant.getAttributes().get(attribute);
             account.getAttributes().put(attribute, value);
         }
-        if (callerIsAdmin(callerRoles)) {
-            updateRoles(callerRoles, participant, account);
+        RequestContext requestContext = getRequestContext();
+        if (requestContext.isAdministrator()) {
+            updateRoles(requestContext, participant, account);
```
DwayneJengSage, line 633: Should we add the superadmin equivalents to adminCanCreateAllRoles() and adminCanUpdateAllRoles() in unit tests?

DwayneJengSage (changes requested):
Overall looks good. Just a few comments about unit tests.

src/main/java/org/sagebionetworks/bridge/models/accounts/UserSession.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added superadminCanCreateEverybody() and superadminCanUpdateEverybody().

alxdarksage:
I also factored out the role tests into a common utility method so future changes are reflected everywhere.

DwayneJengSage (changes requested):
I'm not seeing superadminCanCreateEverybody() and superadminCanUpdateEverybody(). Did we miss a file?

alxdarksage:
Those particular tests were added when I changed the permissions for creating roles on new users (ie. admins can no longer create admins, etc.): https://github.com/Sage-Bionetworks/BridgeServer2/commit/9042501a07420a93f4343d5ea8d4338d0fcddfe9

DwayneJengSage (approved):

PR #136, opened 2019-12-12
Account dao refactor 2

alxdarksage:
- Moves functionality from the DAO to the service
- The getAccount method would return an account even if the caller's substudies did not match the account's substudies, which is a serious security error. This is fixed, however, it in turn led to 500 errors then the ReportTest integration test was run. That has been fixed on the server to return 404 now, but it also pointed to the fact that the ReportTest was using a substudy scoped developer for some calls when it should not have been (and was succeeding only because of this bug on the server). That's also fixed in an integration test PR.
- Removed code that was intended to prevent Hibernate errors by reloading account objects during a single HTTP request (we no longer need to do this now that the reauthentication token has been moved to a separate table);


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -94,30 +223,67 @@ public void createAccount(Study study, Account account, Consumer<Account> afterP
...
     public void editAccount(StudyIdentifier studyId, String healthCode, Consumer<Account> accountEdits) {
-        accountDao.editAccount(studyId, healthCode, accountEdits);
+        checkNotNull(studyId);
+        checkNotNull(healthCode);
+        
+        AccountId accountId = AccountId.forHealthCode(studyId.getIdentifier(), healthCode);
+        Account account = getAccount(accountId);
+        if (account != null) {
+            accountDao.updateAccount(account, accountEdits);
+        }
```
DwayneJengSage: This doesn't do what you want. The old code calls the Consumer, then calls updateAccount. This code calls updateAccount, which runs the Consumer _after_ updating the account.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
-        when(mockAccountDao.getAccount(ACCOUNT_ID)).thenReturn(account);
-        
+
         String healthCode = service.getHealthCodeForAccount(ACCOUNT_ID);
         assertEquals(healthCode, HEALTH_CODE);
         verify(mockAccountDao).getAccount(ACCOUNT_ID);
     }
+
+    @Test
+    public void verifyEmailUsingToken() {
```
DwayneJengSage, line 319: Should this test include a verify call to dao.update()?

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+        assertEquals(account.getStatus(), ENABLED);
+        assertEquals(account.getEmailVerified(), TRUE);
+        // modifiedOn is stored as a long, which loses the time zone of the original time stamp.
+        assertEquals(account.getModifiedOn().toString(), MOCK_DATETIME.withZone(UTC).toString());
+        assertEquals(account.getStatus(), ENABLED);
+        assertEquals(account.getEmailVerified(), TRUE);
+    }
+
+    @Test
+    public void verifyEmailUsingAccount() {
```
DwayneJengSage: This test was identical to verifyEmailUsingToken() even before the refactor and still is. I think we can just delete this test.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+        assertEquals(account.getPhoneVerified(), TRUE);
+        // modifiedOn is stored as a long, which loses the time zone of the original time stamp.
+        assertEquals(account.getModifiedOn().toString(), MOCK_DATETIME.withZone(UTC).toString());
+        assertEquals(account.getStatus(), ENABLED);
+        assertEquals(account.getPhoneVerified(), TRUE);
+        verify(mockAccountDao).updateAccount(account, null);
+    }
+
+    @Test
+    public void verifyPhoneUsingAccount() {
```
DwayneJengSage: Similarly, this test is identical to the previous and can be deleted.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+        verify(mockAccountDao).updateAccount(accountCaptor.capture(), eq(null));
+
+        Account updatedAccount = accountCaptor.getValue();
+        assertEquals(updatedAccount.getId(), USER_ID);
+        assertEquals(updatedAccount.getModifiedOn().getMillis(), MOCK_DATETIME.getMillis());
+        assertEquals(updatedAccount.getPasswordAlgorithm(), PasswordAlgorithm.DEFAULT_PASSWORD_ALGORITHM);
+        assertEquals(updatedAccount.getPasswordModifiedOn().getMillis(), MOCK_DATETIME.getMillis());
+        assertTrue(updatedAccount.getEmailVerified());
+        assertNull(updatedAccount.getPhoneVerified());
+        assertEquals(updatedAccount.getStatus(), ENABLED);
```
DwayneJengSage, line 420: I think I'm missing something here. The code in changePassword() only knows to flip the status to ENABLED if it's UNVERIFIED. However, we don't set status to UNVERIFIED here (or in the other changePassword tests), so how does AccountService know to flip it to ENABLED?

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+        // Execute. Identifiers not allows to change.
+        service.updateAccount(account, null);
+
+        verify(mockAccountDao).updateAccount(accountCaptor.capture(), eq(null));
+
+        Account updatedAccount = accountCaptor.getValue();
+        assertEquals(updatedAccount.getId(), USER_ID);
+        assertEquals(updatedAccount.getEmail(), OTHER_EMAIL);
+        assertEquals(updatedAccount.getPhone().getNationalFormat(), OTHER_PHONE.getNationalFormat());
+        assertEquals(updatedAccount.getModifiedOn().getMillis(), MOCK_DATETIME.getMillis());
```
DwayneJengSage: I think there's a bug in this test. mockGetAccountById() also sets the mockAccountDao to return this account on a get call. So the test gets an account and then uses itself to update itself.

What we want to do is have a different account on the get and update, so we can verify which parameters are being modified and which parameters are kept the same.

In particular, the old version of this test also verified study ID, email verified, phone verified, created on, and password modified on.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+        assertEquals(account.getVersion(), 1);
+
+        // No reauthentication token rotation occurs
+        verify(mockAccountDao, never()).updateAccount(any(), eq(null));
+        assertNull(account.getReauthToken());
+    }
+
+    @Test(expectedExceptions = EntityNotFoundException.class)
+    public void authenticateAccountNotFound() throws Exception {
+        when(mockAccountDao.getAccount(ACCOUNT_ID_WITH_EMAIL)).thenThrow(new EntityNotFoundException(Account.class));
```
DwayneJengSage: This should return an optional empty instead of throwing.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1,167 +1,954 @@
...
+    }
+
+    @Test
+    public void editAccount() throws Exception {
+        AccountId accountId = AccountId.forHealthCode(TEST_STUDY_IDENTIFIER, HEALTH_CODE);
+        Account account = mockGetAccountById(accountId, false);
+
+        service.editAccount(TEST_STUDY, HEALTH_CODE, mockConsumer);
+
+        verify(mockAccountDao).updateAccount(account, mockConsumer);
```
DwayneJengSage: Need to verify that the consumer is called before update.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: You're right. Changed.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Yes, and this call was not entirely completed. The persistedAccount variable isn't needed. Added verify to several verifyChannel  tests that did not have it.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Removed

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: deleted

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: verifyStatus behaves that way, but changePassword always sets the account to enabled if it's going to change the password. I added a value to the status field in these tests (rather than null) to verify this doesn't stop the account from being enabled.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: You're right. I rewrote this test so we doin't mock the account DAO twice and do check for all the fields that cannot be reset with the submitted account object.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Changed

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Added

DwayneJengSage (approved):

PR #137, opened 2019-12-12
Fixed BRIDGE-2471 and some NPE cases that would also return 500

alxdarksage:
Converted some required request parameters to be optional (when required, and not present, Spring is returning a 500 error). In the course of working on that, found a number of cases where accounts were being retrieved and used without checking for null and returning an appropriate 404, even though the userId is sometimes in the URL and can easily be wrong. Fixed all of those, added tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
@@ -327,7 +327,7 @@ public class AccountWorkflowService {
 
         Account account = accountService.getAccount(accountId);
         if (account == null) {
-            throw new EntityNotFoundException(Account.class);
+            return;
         }
         boolean verifiedEmail = account.getEmail() != null && Boolean.TRUE.equals(account.getEmailVerified());
         boolean verifiedPhone = account.getPhone() != null && Boolean.TRUE.equals(account.getPhoneVerified());
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
@@ -326,7 +326,9 @@ public class AccountWorkflowService {
         checkNotNull(accountId);
 
         Account account = accountService.getAccount(accountId);
-        
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
         boolean verifiedEmail = account.getEmail() != null && Boolean.TRUE.equals(account.getEmailVerified());
         boolean verifiedPhone = account.getPhone() != null && Boolean.TRUE.equals(account.getPhoneVerified());
         boolean sendEmail = study.isEmailVerificationEnabled() && !study.isAutoVerificationEmailSuppressed();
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
```diff
@@ -326,7 +326,9 @@ public void notifyAccountExists(Study study, AccountId accountId) {
         checkNotNull(accountId);
 
         Account account = accountService.getAccount(accountId);
-        
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
```
DwayneJengSage: This can only happen if the user tries to sign up with an identifier that already exists, and then while we're trying to notify the existing account, the account becomes deleted. In this case, it would be weird for sign-up to return a 404. I think a 500 Internal Server Error would be appropriate here, or possibly a 409 Concurrent Modification Error.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportControllerTest.java
```diff
@@ -476,6 +481,46 @@ public void nonAdminCannotDeleteParticipantIndex() {
...
+        controller.getParticipantReport(USER_ID, REPORT_ID, null, null);
+    }
+    
+    @Test(expectedExceptions = EntityNotFoundException.class, 
+            expectedExceptionsMessageRegExp=".*Account not found.*")
+    public void getParticipantReportForWorkerAccountNotFound() {
+        doReturn(session).when(controller).getAuthenticatedSession(WORKER);
+        reset(mockAccountService);
+        controller.getParticipantReportForWorker(TEST_STUDY_IDENTIFIER, USER_ID, REPORT_ID, null, null);
+    }
```
DwayneJengSage, line 498: Unit tests for the v4 versions of this API?

DwayneJengSage (changes requested):
Can we fix this at the Spring level? I'm concerned that if we only patch it at the API level, we'll end up playing whack-a-mole ad nauseam.

src/main/java/org/sagebionetworks/bridge/services/AccountWorkflowService.java
alxdarksage: At first I was going to with this to a 409 response which makes sense to me, but that would return to the caller who might then infer that the account has existed (which we're trying to avoid)... so I think the best response here may be to return and send nothing.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportControllerTest.java
alxdarksage: Added

alxdarksage:
"Can we fix this at the Spring level? I'm concerned that if we only patch it at the API level, we'll end up playing whack-a-mole ad nauseam." We might be able to handle that specific exception by returning a different error code through an @ExceptionHandler, but in this case a null is valid, and we want it to call the controller, because we use a default time range of two weeks before the request. So it's really not required. At best we could change the 500 to a 400 if we want this to be an error.

DwayneJengSage (approved):

PR #139, opened 2019-12-19
Fix for BRIDGE-2658

alxdarksage:
On investigation, the getScheduleWindowsBasedOnEvents(...) method was returning a start and end date time that are not in the timezone of the current request, which is then used to truncate scheduled activities in ways that are not in the user's timezone.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
@@ -51,14 +51,13 @@ public abstract class ActivityScheduler {
             // Start time is event time, with added delay if it's present.
             DateTime startTime = oneEventTime;
             if (schedule.getDelay() != null) {
-                startTime = oneEventTime.withZone(context.getRequestTimeZone()).plus(schedule.getDelay());
+                startTime = startTime.plus(schedule.getDelay());
             }
-
             // End time is only present if a sequence period is specified.
             // Sequence period is measured from the first timestamp plus the delay, to the end of the period.
             DateTime endTime = null;
             if (schedule.getSequencePeriod() != null) {
-                endTime = startTime.withZone(context.getRequestTimeZone()).plus(schedule.getSequencePeriod());
+                endTime = startTime.plus(schedule.getSequencePeriod());
             }
 
             // Construct the range tuple.
--- a/src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
@@ -51,14 +51,14 @@ public abstract class ActivityScheduler {
             // Start time is event time, with added delay if it's present.
             DateTime startTime = oneEventTime;
             if (schedule.getDelay() != null) {
-                startTime = oneEventTime.plus(schedule.getDelay());
+                startTime = oneEventTime.withZone(context.getRequestTimeZone()).plus(schedule.getDelay());
             }
 
             // End time is only present if a sequence period is specified.
             // Sequence period is measured from the first timestamp plus the delay, to the end of the period.
             DateTime endTime = null;
             if (schedule.getSequencePeriod() != null) {
-                endTime = startTime.plus(schedule.getSequencePeriod());
+                endTime = startTime.withZone(context.getRequestTimeZone()).plus(schedule.getSequencePeriod());
             }
 
             // Construct the range tuple.
@@ -162,7 +162,7 @@ public abstract class ActivityScheduler {
             Iterable<String> eventIds = Schedule.EVENT_ID_SPLITTER.split(eventIdsString.trim());
             for (String thisEventId : eventIds) {
                 if (context.getEvent(thisEventId) != null) {
-                    eventDateTimeList.add(context.getEvent(thisEventId));
+                    eventDateTimeList.add(context.getEvent(thisEventId).withZone(context.getRequestTimeZone()));
 
                     if (!getAll) {
                         // We only wanted one event, and we found it, so break.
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -51,14 +51,14 @@
             // Start time is event time, with added delay if it's present.
             DateTime startTime = oneEventTime;
             if (schedule.getDelay() != null) {
-                startTime = oneEventTime.plus(schedule.getDelay());
+                startTime = oneEventTime.withZone(context.getRequestTimeZone()).plus(schedule.getDelay());
             }
```
DwayneJengSage: If delay is null (which can be fairly common), then we don't set the time zone. It might be cleaner to put the withZone() on line 65 (scheduleWindowList.add(...)), which should guarantee that we never add something without ensuring its timezone.

src/test/java/org/sagebionetworks/bridge/models/schedules/ActivitySchedulerTest.java
```diff
@@ -454,6 +460,35 @@ public void greaterOfMinOrTimeBasedSchedulingReturned() {
...
+                .withEndsOn(DateTime.parse("2015-03-26T14:40:00+05:30"))
+                .withEvents(events).build();
+        
+        assertEquals(context.getRequestTimeZone(), IST);
+        
+        Schedule schedule = new Schedule();
+        schedule.setEventId("enrollment,enrollment2");
+        // Need a recurring schedule with a sequence to generate an end time and multiple windows
+        schedule.setSequencePeriod("P1W");
+        schedule.setScheduleType(RECURRING);
```
DwayneJengSage, line 482: We should also have a test that uses delay, just to make sure we set the time zone in both cases.

src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
```diff
@@ -162,7 +162,7 @@ private LocalDateTime getExpiresOn(LocalDate localDate, LocalTime localTime) {
             Iterable<String> eventIds = Schedule.EVENT_ID_SPLITTER.split(eventIdsString.trim());
             for (String thisEventId : eventIds) {
                 if (context.getEvent(thisEventId) != null) {
-                    eventDateTimeList.add(context.getEvent(thisEventId));
+                    eventDateTimeList.add(context.getEvent(thisEventId).withZone(context.getRequestTimeZone()));
```
DwayneJengSage, line 164: Is it worth adding a unit test that calls getEventDateTimes() directly, to ensure all datetimes returned by this method are in the correct time zone?

src/main/java/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
```diff
@@ -85,6 +85,10 @@ public DateTime getStartsOn() {
         return startsOn;
     }
     
+    public DateTimeZone getRequestTimeZone() { 
+        return startsOn.getZone();
+    }
+
```
DwayneJengSage, line 91: Do we also want to ensure startsOn and endsOn are in the same time zone?

src/test/java/org/sagebionetworks/bridge/models/schedules/IntervalActivitySchedulerTest.java
```diff
@@ -52,6 +52,36 @@ public void after() {
...
+            .withEndsOn(DateTime.parse("2019-12-19T00:00:00.000+05:30"))
+            .withEvents(events).build();
+        scheduledActivities = schedule.getScheduler().getScheduledActivities(plan, context);
+
+        assertEquals(scheduledActivities.size(), 2);
+        assertEquals(scheduledActivities.get(0).getScheduledOn().toString(), "2019-12-17T08:00:00.000+05:30");
+        assertEquals(scheduledActivities.get(0).getExpiresOn().toString(), "2019-12-18T08:00:00.000+05:30");
+        assertEquals(scheduledActivities.get(1).getScheduledOn().toString(), "2019-12-18T08:00:00.000+05:30");
+        assertEquals(scheduledActivities.get(1).getExpiresOn().toString(), "2019-12-19T08:00:00.000+05:30");
+        DateTimeUtils.setCurrentMillisSystem();
```
DwayneJengSage: Minor: This call to setCurrentMillisSystem() is redundant with after().

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: I've removed the .withZone() calls here because this is already being set when the event map of event times is being constructed, and to your point in the comment below, we should verify there that this is being set correctly.

src/main/java/org/sagebionetworks/bridge/models/schedules/ActivityScheduler.java
alxdarksage: Added a test that the time zone conversion happens in this method.

src/test/java/org/sagebionetworks/bridge/models/schedules/IntervalActivitySchedulerTest.java
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/models/schedules/ScheduleContext.java
alxdarksage: This is checked at ScheduledActivityController line 144. And it's a bug insofar as when time zones change, these can be different and then the query fails (BRIDGE-2126). Client developers have to force them to be the same.

DwayneJengSage (approved):

PR #142, opened 2020-01-14
MasterScheduleConfig scan should use consistent read

DwayneJengSage:
By default, scan is eventually consistent (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html). Make this use consistent read.


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoMasterSchedulerConfigDao.java
```diff
@@ -47,7 +47,7 @@ final void setMapper(DynamoDBMapper mapper) {
 
     @Override
     public List<MasterSchedulerConfig> getAllSchedulerConfig() {
-        DynamoDBScanExpression scan = new DynamoDBScanExpression();
+        DynamoDBScanExpression scan = new DynamoDBScanExpression().withConsistentRead(true);
```
alxdarksage, line 50: I thought we had to mark a table or specific index for consistent read before it would be possible to get a consistent read, I didn't know we could ask for it on a particular scan or query (but that is what it says in the docs).

alxdarksage (approved):

PR #143, opened 2020-01-15
Rollback attempt to fix scheduling behavior described by BRIDGE-2658

alxdarksage:
Rolls back the change in time zone causing us issues, along with the original tests.


DwayneJengSage:
I would prefer to do a 1-click revert on the original pull request https://github.com/Sage-Bionetworks/BridgeServer2/pull/139 This way, we can mostly guarantee that the revert is identical and we don't have to review extra code.

alxdarksage:
If you want to do that, it's okay with me, but note there's an accompanying reversion to the integration tests.

PR #150, opened 2020-03-11
Assessments resource links

alxdarksage:
JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-2788

Design doc: https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/914882561/Assessments+shared+assessments+and+their+resources+and+tags#Assessment-Resources


src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -669,5 +674,85 @@ public static boolean isInRole(Set<Roles> callerRoles, Set<Roles> requiredRoles)
...
+        if (parts.length != 2) {
+            LOG.error("Could not parse shared assessment ownerID, guid=" + guid + ", ownerId=" + ownerId);
+            throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+        }
+        String originAppId = parts[0];
+        String originOrgId = parts[1];
+        Set<String> callerSubstudies = BridgeUtils.getRequestContext().getCallerSubstudies();
+        boolean scopedUser = !callerSubstudies.isEmpty();
+        boolean orgMember = callerSubstudies.contains(originOrgId);
+        if (!callerAppId.equals(originAppId) || (scopedUser && !orgMember)) {
```
DwayneJengSage, line 727: Unit test for wrong app ID?

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -669,5 +674,85 @@ public static boolean isInRole(Set<Roles> callerRoles, Set<Roles> requiredRoles)
...
+        // This happens in tests, we expect it to never happens in production. So log if it does.
+        if (parts.length != 2) {
+            LOG.error("Could not parse shared assessment ownerID, guid=" + guid + ", ownerId=" + ownerId);
+            throw new UnauthorizedException();
+        }
+        String originAppId = parts[0];
+        String originOrgId = parts[1];
+        return new Tuple<>(originAppId, originOrgId);
+    }
+    */
```
DwayneJengSage: Delete commented code.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
```diff
@@ -134,20 +145,67 @@ public Assessment saveAssessment(String appId, Assessment assessment) {
...
+        List<HibernateAssessmentResource> hibernateDestResources = destResources.stream()
+                .map((resource) -> HibernateAssessmentResource.create(resource, SHARED_STUDY_ID_STRING, assessmentId))
+                .collect(Collectors.toList());
+
+        HibernateAssessment retValue = hibernateHelper.executeWithExceptionHandling(hibernateOrigin, (session) -> {
+            // logically delete any existing resources in destination app context 
+            NativeQuery<?> query = session.createNativeQuery(LOGICAL_DELETE_SQL);
+            query.setParameter("appId", SHARED_STUDY_ID_STRING);
+            query.setParameter("assessmentId", assessmentId);
+            query.executeUpdate();
```
DwayneJengSage: It seems really uninuitive to me that publishing a new version of an assessment overwrites all resources for that assessment identifier.

This will get particularly messy with the delete API. If the study developer creates assessment rev1 and adds docs, then creates assessment rev2 and adds docs, and then deletes assessment rev1, this also deletes the docs for assessment rev2.

If resources are tied to the assessment identifier, it makes more sense for their to be a separate API to update them. If we want to use publish/import/delete to update resources, it makes more sense to have resources be tied to the assessment revision.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
```diff
@@ -0,0 +1,438 @@
...
+        service.getResource(APP_ID, ASSESSMENT_ID, GUID);
+    }
+    
+    @Test
+    public void createResource() {
+        Assessment assessment = AssessmentTest.createAssessment();
+        when(mockAssessmentService.getLatestAssessment(APP_ID, ASSESSMENT_ID)).thenReturn(assessment);
+        
+        AssessmentResource resource = AssessmentResourceTest.createAssessmentResource();
+        resource.setDeleted(true);
```
DwayneJengSage: Also set the guid, createdOn, and modifiedOn to dummy values so we can verify they are actually being overwritten.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
```diff
@@ -0,0 +1,438 @@
...
+        when(mockDao.saveResource(eq(APP_ID), eq(ASSESSMENT_ID), any())).thenReturn(resource);
+        
+        AssessmentResource retValue = service.createResource(APP_ID, ASSESSMENT_ID, resource);
+        assertSame(retValue, resource);
+        assertEquals(resource.getGuid(), GUID);
+        assertEquals(resource.getCreatedOn(), CREATED_ON);
+        assertEquals(resource.getModifiedOn(), CREATED_ON);
+        assertFalse(resource.isDeleted());
+        assertEquals(resource.getCreatedAtRevision(), 5);
```
DwayneJengSage, line 192: Check upToDate()?

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
```diff
@@ -785,15 +795,19 @@ public void publishAssessment() {
         
         when(mockDao.getAssessment(APP_ID_VALUE, "oldGuid")).thenReturn(Optional.of(existing));
         
-        when(mockDao.publishAssessment(any(), any(), any())).thenReturn(ASSESSMENT);
+        when(mockResourceDao.getResources(APP_ID_VALUE, IDENTIFIER, 0, service.getPageSize(), null, null, null, false))
+                .thenReturn(new PagedResourceList<>(ImmutableList.of(), 0));
```
DwayneJengSage: Should this unit test load resources and passes it into the assessment dao?

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
```diff
@@ -876,21 +895,16 @@ public void importAssessment() {
...
         when(mockDao.getAssessmentRevisions(APP_ID_VALUE, IDENTIFIER, 0, 1, true))
-            .thenReturn(new PagedResourceList<>(ImmutableList.of(localAssessment), 1));
-        
-        when(mockDao.saveAssessment(APP_ID_VALUE, sharedAssessment)).thenReturn(sharedAssessment);
-        
-        when(mockDao.getAssessmentRevisions(APP_ID_VALUE, IDENTIFIER, 0, 1, false)).thenReturn(EMPTY_LIST);
+            .thenReturn(new PagedResourceList<>(ImmutableList.of(), 0));
+        when(mockResourceDao.getResources(SHARED_STUDY_ID_STRING, IDENTIFIER, 0, service.getPageSize(), null, null, null, false))
+            .thenReturn(new PagedResourceList<>(ImmutableList.of(), 0));
+        when(mockDao.importAssessment(eq(APP_ID_VALUE), eq(sharedAssessment), any())).thenReturn(sharedAssessment);
```
DwayneJengSage: Similarly, should this also load resources and pass them in to the dao?

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
```diff
@@ -422,6 +403,31 @@ private int nextRevisionNumber(String appId, String identifier) {
...
+                oneResource.setCreatedOn(timestamp);
+                oneResource.setModifiedOn(timestamp);
+                oneResource.setDeleted(false);
+                oneResource.setCreatedAtRevision(assessmentRev);
+                oneResource.setVersion(0);
+                resourcesToPublish.add(oneResource);
+                
+                offset += getPageSize();
+            }
+        } while(page.getItems().size() == getPageSize());
```
DwayneJengSage: If we're going to load all of the resources, I think we'd be better off just asking the database for all the resources instead of calling the paginated API in a loop.

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
```diff
@@ -422,6 +403,31 @@ private int nextRevisionNumber(String appId, String identifier) {
         return opt.isPresent() ? (opt.get().getRevision()+1) : 1;
     }
     
+    List<AssessmentResource> loadResourcesForAssessment(String appId, String assessmentId, int assessmentRev) {
```
DwayneJengSage: Minor: Add a comment explaining that this is for copying resources when we publish and import. Out of context, it's not immediately obvious that's what this method does.

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
```diff
@@ -1132,6 +1099,50 @@ public void updateSharedAssessmentChecksOwnershipWhenMissing() {
         service.updateSharedAssessment(APP_ID_VALUE, sharedAssessment);
     }
     
+    @SuppressWarnings("unchecked")
+    @Test
+    public void loadResourcesForAssessment() {
+        when(service.getPageSize()).thenReturn(2);
+        AssessmentResource res = AssessmentResourceTest.createAssessmentResource();
```
DwayneJengSage: Set guid, createdOn, and modifiedOn to dummy values so we can verify that we're overwriting them.

src/main/resources/db/changelog/changelog.sql
```diff
@@ -310,3 +310,35 @@ CREATE TABLE `AssessmentTags` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
 
 CREATE INDEX `AssessmentTags-TagValue` ON `AssessmentTags`(`tagValue`);
+
+-- changeset bridge:14
+
+CREATE TABLE `ExternalResources` (
+  `guid` varchar(255) NOT NULL,
+  `appId` varchar(255) NOT NULL,
+  `assessmentId` varchar(255),
```
DwayneJengSage, line 319: Can you create a resource without an assessment? Does a foreign key make sense here?

src/main/resources/db/changelog/changelog.sql
```diff
@@ -310,3 +310,35 @@ CREATE TABLE `AssessmentTags` (
...
+    'OTHER') NOT NULL,
+  `contributors` text,
+  `creators` text,
+  `publishers` text,
+  `minRevision` int(10),
+  `maxRevision` int(10),
+  `createdAtRevision` int(10) NOT NULL,
+  `createdOn` bigint(20) unsigned DEFAULT NULL,
+  `modifiedOn` bigint(20) unsigned DEFAULT NULL,
+  `deleted` tinyint(1) DEFAULT '0',
```
DwayneJengSage: Should deleted be non-null?

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
alxdarksage: Mocked this out further to return three resources, and verify that they are passed to the DAO.

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
alxdarksage: Added and verified the import call carries them over to the DAO.

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
alxdarksage: Changed this to a single call (although this may be going away given design discussions).

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/AssessmentServiceTest.java
alxdarksage: changed these to verify they are set correctly

src/main/resources/db/changelog/changelog.sql
alxdarksage: Eventually I think you could, by adding another FK column... however you cannot enforce the FK constraint once there are different FK columns to different tables. For example, we could add a "protocolGuid" column if they also had external resource links. I'm not 100% sure we'll need this, but I didn't want to create something that makes it impossible later.

src/main/resources/db/changelog/changelog.sql
alxdarksage: changed

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
alxdarksage: This has been changed as described yesterday in conversation, I will update internal design doc and external docs shortly. In short, resources are imported/published separately, though you must do the assessment first.

alxdarksage: This has been updated along with the SDK and integration tests.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
```diff
@@ -134,20 +139,37 @@ public Assessment saveAssessment(String appId, Assessment assessment) {
     public void deleteAssessment(String appId, Assessment assessment) {
         HibernateAssessment hibernateAssessment = HibernateAssessment.create(assessment, appId);
         hibernateHelper.executeWithExceptionHandling(hibernateAssessment, (session) -> {
+            String assessmentId = hibernateAssessment.getIdentifier();
+            NativeQuery<?> query = session.createNativeQuery(DELETE_RESOURCES_SQL);
+            query.setParameter("appId", appId);
+            query.setParameter("assessmentId", assessmentId);
+            query.executeUpdate();
```
DwayneJengSage: I don't think we should delete resources when we delete the assessment. If I'm understanding this correctly, this API deletes an assessment revision, but the resources are tied to the assessment ID, so deleting the revision deletes resources for all revisions.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
```diff
@@ -0,0 +1,698 @@
...
+        when(mockDao.getResource("targetId", "guid3")).thenReturn(Optional.empty());
+        
+        List<AssessmentResource> retValue = service.copyResources("originId", "targetId", assessment, guids);
+        assertSame(retValue, resources);
+        
+        verify(mockDao).saveResources(eq("targetId"), eq(IDENTIFIER), resourceListCaptor.capture());
+        
+        AssessmentResource ar1 = resourceListCaptor.getValue().get(0);
+        assertEquals(ar1.getCreatedOn().toString(), TIMESTAMP.toString());
+        assertEquals(ar1.getModifiedOn().toString(), TIMESTAMP.toString());
```
DwayneJengSage, line 470: TIMESTAMP and MODIFIED_ON appear to be the same timestamp, so this test won't verify that modifiedOn was changed.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
```diff
@@ -0,0 +1,698 @@
...
+        assertEquals(ar1.getGuid(), "guid1");
+        assertEquals(ar1.getCreatedAtRevision(), 100);
+        assertEquals(ar1.getVersion(), 0);
```
DwayneJengSage, line 473: Should we check the deleted flag?

src/main/java/org/sagebionetworks/bridge/spring/controllers/BaseController.java
```diff
@@ -308,6 +309,17 @@ CriteriaContext getCriteriaContext(UserSession session) {
...
+    protected @Nonnull <T> T parseJson(TypeReference<? extends T> clazz) {
+        try {
+            return MAPPER.readValue(request().getInputStream(), clazz);
+        } catch (Throwable ex) {
+            if (Throwables.getRootCause(ex) instanceof InvalidEntityException) {
+                throw (InvalidEntityException)Throwables.getRootCause(ex);
+            }
+            throw new InvalidEntityException("Error parsing JSON in request body: " + ex.getMessage());    
+        }
+    }
```
DwayneJengSage, line 322: In BaseController.Test, we have tests for parseJson(Class). Should we add unit tests for parseJson(TypeReference)

src/main/java/org/sagebionetworks/bridge/spring/controllers/SharedAssessmentResourceController.java
```diff
@@ -0,0 +1,107 @@
...
+        if ("true".equals(physical)) {
+            service.deleteResourcePermanently(SHARED_STUDY_ID_STRING, assessmentId, guid);
+        } else {
+            service.deleteResource(SHARED_STUDY_ID_STRING, assessmentId, guid);
+        }
+        return new StatusMessage("Assessment resource deleted.");        
+    }
+    
+    @PostMapping("/v1/sharedassessments/identifier:{assessmentId}/resources/import")
+    public ResourceList<AssessmentResource> importAssessmentResources(@PathVariable String assessmentId) {
```
DwayneJengSage, line 98: Missing unit tests for this.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
alxdarksage: This is a physical delete, as we do for tests. I think at the least, if we're deleting the last revision of this assessment, we should clean up the resources. Changed to check for that.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
alxdarksage: I've changed the constants file so these are different... that was an error in the constants. Tests pass.

src/test/java/org/sagebionetworks/bridge/services/AssessmentResourceServiceTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/BaseController.java
alxdarksage: Added tests

src/main/java/org/sagebionetworks/bridge/spring/controllers/SharedAssessmentResourceController.java
alxdarksage: Added

DwayneJengSage (approved):

PR #151, opened 2020-03-13
Tag API to clean up tags.

alxdarksage:


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTagDao.java
```diff
@@ -0,0 +1,75 @@
...
+        return API_MAXIMUM_PAGE_SIZE;
+    }
+    
+    @Override
+    public Map<String, List<String>> getTags() {
+        Map<String, List<String>> byGroup = new HashMap<>();
+        List<Tag> page = null;
+        int offset = 0;
+        do {
+            page = hibernateHelper.queryGet("from Tag", null, offset, getPageSize(), Tag.class);
```
DwayneJengSage: We'd be better off asking the database for all the tags instead of calling a paginated API in a loop.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTagDao.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+    
+    private static void addByNameSpace(Map<String, List<String>> byGroup, Tag tag) {
+        String[] elements = tag.getValue().split(":", 2);
+        String ns = (elements.length == 1) ? "default" : elements[0];
+        String tagValue = (elements.length == 1) ? elements[0] : elements[1];
+        
+        byGroup.putIfAbsent(ns, new ArrayList<>());
+        byGroup.get(ns).add(tagValue);
+    }
```
DwayneJengSage: Consider using [Guava multimaps](https://github.com/google/guava/wiki/NewCollectionTypesExplained#multimap) and [Multimaps.index()](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/Multimaps.html#index-java.lang.Iterable-com.google.common.base.Function-) to make this simpler and more readable.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTagDao.java
alxdarksage: Removed paging

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTagDao.java
alxdarksage: Switched to Multimap, which is much cleaner

DwayneJengSage (approved):

PR #162, opened 2020-04-01
BRIDGE-2806

alxdarksage:
JSON parsing errors that come from parsing request bodies should be InvalidEntityExceptions (these return the parsing error to users, but don't log sensitive information because they are 400-level exceptions). In most controller methods, we call the parseJson(...) methods, which correct transform Jackson deserialization errors to IEEs, however there are a handful of methods where we directly use the ObjectMapper, where this exception handling is missing. This PR:

1) finds and fixes all instances of this in the controllers, creating some new parseJson() methods to clean this up;

2) adds tests for all these cases to ensure they return an InvalidEntityException;

3) removes a lot of unnecessary "throws Exception" statements from the controllers where there were no checked exceptions being thrown. The only remaining "throws" statements come from places where we are serializing response values, using the ObjectMapper (I think these are OK to leave because they will translate to 500 errors, and they shouldn't happen and they would be a 500 if they occurred). Alternatively we could say that all checked exceptions should be handled where they occur, so they are semantically correct (e.g. Jackson errors in the exception handler can't be judged as user input errors vs. server coding errors), and no controllers should declare a "throws" clause. But that would be a different PR.


src/main/java/org/sagebionetworks/bridge/spring/controllers/BaseController.java
```diff
@@ -311,24 +317,25 @@ CriteriaContext getCriteriaContext(UserSession session) {
...
+    private @Nonnull <T> T parseWithExceptionConversion(ExceptionThrowingSupplier<T> supplier) {
         try {
-            return MAPPER.readValue(request().getInputStream(), clazz);
+            return supplier.get();
         } catch (Throwable ex) {
             if (Throwables.getRootCause(ex) instanceof InvalidEntityException) {
                 throw (InvalidEntityException)Throwables.getRootCause(ex);
             }
-            throw new InvalidEntityException("Error parsing JSON in request body: " + ex.getMessage());    
+            throw new InvalidEntityException("Error parsing JSON in request body" + ex.getMessage());
```
DwayneJengSage: This still leaks the phone number into the logs, just as an INFO instead of an ERROR.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -226,9 +226,9 @@ public StatusMessage signOutV4() throws Exception {
     
     @PostMapping({"/v3/auth/signUp", "/api/v1/auth/signUp"})
     @ResponseStatus(HttpStatus.CREATED)
-    public StatusMessage signUp() throws Exception {
+    public StatusMessage signUp() {
         JsonNode node = parseJson(JsonNode.class);
-        StudyParticipant participant = MAPPER.treeToValue(node, StudyParticipant.class);
+        StudyParticipant participant = parseJson(node, StudyParticipant.class);
```
DwayneJengSage, line 231: This block parses request as a JsonNode, then re-parses the JsonNode as a StudyParticipant. I think you only need one parseJson() here.

src/main/java/org/sagebionetworks/bridge/spring/controllers/IntentController.java
```diff
@@ -28,12 +28,12 @@ final void setIntentService(IntentService intentService) {
...
     @PostMapping("/v3/itp")
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public StatusMessage submitIntentToParticipate() throws Exception {
+    public StatusMessage submitIntentToParticipate() {
         // An early hack in the system was that sharing scope was added to the consent signature 
         // JSON even though it is not part of the signature. We need to move that value because 
         // the client API continues to treat sharing as part of the consent signature.
         JsonNode requestNode = parseJson(JsonNode.class);
-        IntentToParticipate intent = MAPPER.treeToValue(requestNode, IntentToParticipate.class);
+        IntentToParticipate intent = parseJson(requestNode, IntentToParticipate.class);
```
DwayneJengSage, line 36: Same here.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/BaseController.java
alxdarksage: I've created a method to change this message so it no longer leaks information from the raw exception message.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: The node is needed to retrieve the "study" property, which is not part of the StudyParticipant object. Parsing to a node, and then to StudyParticipant and a JsonUtils.asText(...) call, is needed.

src/main/java/org/sagebionetworks/bridge/spring/controllers/IntentController.java
alxdarksage: The intermediate parse to a node is needed to retrieve the sharingOption, which is not part of the ConsentSignature object.

src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -1033,6 +1040,34 @@ public void getIntegerOrDefaultThrowsBadRequest() {
...
+    public void convertParsingErrorMismatchedInputException() {
+        MismatchedInputException mie = Mockito.mock(MismatchedInputException.class);
+        Reference ref1 = new Reference(null, "password");
+        Reference ref2 = new Reference(null, "phone");
+        List<Reference> list = ImmutableList.of(ref1, ref2);
+        Mockito.when(mie.getPath()).thenReturn(list);
+        
+        InvalidEntityException e = BridgeUtils.convertParsingError(mie);
+        assertEquals(e.getMessage(), "Error parsing JSON in request body fields: password, phone");
+    }
```
DwayneJengSage: Let's take this one step further: Instead of mocking the exception and hoping we got the format right, let's make some malformed JSON and parse the JSON so that we have a real exception to work with.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: Now throwing a real exception. I also widened the test to the parent exception that includes the getPath() method (although I haven't been able to trigger the more obscure exceptions, like UnresolvedForwardReference, I worry they might also leak information). And added a comma to the error message because rereading it... it was confusing.

DwayneJengSage (approved):

PR #166, opened 2020-04-06
Remove StudyIdentifier object

alxdarksage:
In preparation of a large rename from study --> app, removing the StudyIdentifier object. We're not going to implement this for an app identifier so everywhere, this needs to be a String.

All unit and integration tests run. The one thing to focus on is that sessions in Redis will have a serialized version of StudyIdentifier and that needs to be deserialized correctly to a String by the CacheProvider. We also lave StudyIdentifier on the AccountSummary object because it is part of the public API, and we'll need to update SDK consumers to use an AccountSummary.appId property before it can be removed.


src/main/java/org/sagebionetworks/bridge/models/accounts/AccountSummary.java
```diff
@@ -21,16 +35,15 @@
...
     private final Set<String> substudyIds;
     
     @JsonCreator
     public AccountSummary(@JsonProperty("firstName") String firstName, @JsonProperty("lastName") String lastName,
             @JsonProperty("email") String email, @JsonProperty("synapseUserId") String synapseUserId,
             @JsonProperty("phone") Phone phone, @JsonProperty("externalIds") Map<String, String> externalIds,
             @JsonProperty("id") String id, @JsonProperty("createdOn") DateTime createdOn,
-            @JsonProperty("status") AccountStatus status,
-            @JsonProperty("studyIdentifier") StudyIdentifier studyIdentifier,
+            @JsonProperty("status") AccountStatus status, @JsonProperty("studyId") String studyId,
```
DwayneJengSage: Does it break JSON parsing if you change the property name from studyIdentifier to studyId?

DwayneJengSage (commented):
I only looked at the implementation, not the tests. One comment on AccountSummary and JSON parsing. Assuming that one issue is a non-issue, should be fine to merge.

src/main/java/org/sagebionetworks/bridge/models/accounts/AccountSummary.java
alxdarksage: We never receive this object through the API, and we never store it in Redis, so we don't use this deserialization code. We could remove It in fact... but I don't want to do anything like that during this refactor. I will restore it to what it was just so there's no confusion about this.

PR #167, opened 2020-04-10
Parse a new device name string that's found in the wild.

alxdarksage:
See BRIDGE 2822. Also converted some strings to patterns because this is faster to execute.


nategbrown9 (approved):
Looks good.

PR #170, opened 2020-04-13
App id refactor: rename app ID constants and organize them

alxdarksage:
Rename the app ID constants for the shared and API studies, and put them under BridgeConstants, because these are studies we create in every environment. Use a different string for tests, via the renamed TestConstants.TEST_APP_ID constant.

This is a large change, but it's very mechanical with no logic changes. All unit and integration tests pass, there are no changes to the API or SDK.


DwayneJengSage (approved):
Mostly just looked at code changes, didn't really look at test changes. Looks fine to me.

PR #185, opened 2020-04-22
Refactor more upload classes and user data download classes

alxdarksage:
The logic to read the environment variable `upload.dupe.study.whitelist` has been simplified and removed.


nategbrown9 (approved):
Looks good.

PR #188, opened 2020-04-23
App id little 13

alxdarksage:
Converting templates, OAuth classes, and discovered the very useful @JsonAlias annotation which allows us to deserialize "study" and "appId" to the same property. Adjusted AccountId.


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplate.java
```diff
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.hibernate;
 
+import javax.persistence.Column;
```
nategbrown9, line 3: Is this import used?

nategbrown9 (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateTemplate.java
alxdarksage: Yes, as an annotation on getAppId().

PR #190, opened 2020-04-24
Study --> App

alxdarksage:
Renaming the Study class to App, and changing every variable of type App from "study" (or some variant) to "app" (or some variant, e.g. newApp if it was originally newStudy). **And that's it.** This is already a huge number of changes so I have not changed comments or anything else so this is the only kind of change that has to be reviewed in this PR.


nategbrown9 (approved):
Looks good.

PR #201, opened 2020-04-30
The last server update

alxdarksage:
There may be a few more references to the old terminology here or there, but I've gone through almost every line of code at this point and this is probably good enough for now. These changes introduce the rest of the /v1/apps endpoints and HTTP parameters and similar changes to the API, while keeping all the old stuff for migration.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/ExportViaSqsService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ExportViaSqsService.java
@@ -25,7 +25,8 @@ public class ExportViaSqsService implements ExportService {
     // constants - these are package scoped so unit tests can access them
     static final String CONFIG_KEY_EXPORTER_SQS_QUEUE_URL = "exporter.request.sqs.queue.url";
     static final String REQUEST_KEY_END_DATE_TIME = "endDateTime";
-    static final String REQUEST_KEY_APP_WHITELIST = "studyWhitelist";
+    static final String REQUEST_KEY_STUDY_WHITELIST = "studyWhitelist";
+    static final String REQUEST_KEY_APP_WHITELIST = "appWhitelist";
     static final String REQUEST_KEY_TAG = "tag";
     static final String REQUEST_KEY_USE_LAST_EXPORT_TIME = "useLastExportTime";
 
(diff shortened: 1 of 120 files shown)


src/main/java/org/sagebionetworks/bridge/services/ExportViaSqsService.java
```diff
@@ -61,6 +62,7 @@ public void startOnDemandExport(String appId) throws JsonProcessingException {
         ObjectNode requestNode = JSON_OBJECT_MAPPER.createObjectNode();
         requestNode.put(REQUEST_KEY_END_DATE_TIME, endDateTimeStr);
         requestNode.set(REQUEST_KEY_APP_WHITELIST, appWhitelistNode);
+        requestNode.set(REQUEST_KEY_STUDY_WHITELIST, appWhitelistNode);
```
DwayneJengSage, line 65: Do we plan on changing the Exporter to take in the new attribute as well?

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
```diff
@@ -166,7 +166,7 @@ public ReportIndex getParticipantReportIndex(@PathVariable String identifier) {
     }
 
     /** Worker API to get reports for the given user in the given app by date-time. */
-    @GetMapping(path = {"/v1/apps/{appId}/participants/{userId}/reports/{reportId}",
+    @GetMapping(path = {"/v2/apps/{appId}/participants/{userId}/reports/{reportId}",
```
DwayneJengSage, line 169: Is there any reason we're bumping this to v2 when all the others are in v1?

DwayneJengSage (approved):
Looks good. Just a few questions.

src/main/java/org/sagebionetworks/bridge/services/ExportViaSqsService.java
alxdarksage: Yes. The next step is to change all the downstream consumers to use the new API.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
alxdarksage: We originally had a /v3/studies/* and /v4/studies/* version of these APIs that took a date and dateTime range, respectively. So I created these as /v1/apps/* and /v2/apps/*.

PR #203, opened 2020-05-08
Implementation of proposed API to integrate with CUIMC

alxdarksage:
Recent changes: downgraded to v3 of FHIR, switched to PUT with 200/201 status responses. Also supports Basic HTTP Authentication, state changes, writing reports.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -123,7 +123,7 @@ public class CRCController extends BaseController {
         // Force userId of the URL
         StudyParticipant.Builder builder = new StudyParticipant.Builder()
                 .copyOf(participant)
-                .withId(session.getId());
+                .withId(userId);
         
         Set<String> dataGroups = participant.getDataGroups();
         if (dataGroups.contains(AccountStates.SELECTED.name().toLowerCase())) {
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -3,11 +3,13 @@ package org.sagebionetworks.bridge.spring.controllers;
 import static com.google.common.net.HttpHeaders.AUTHORIZATION;
 import static java.lang.Boolean.TRUE;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.http.entity.ContentType.APPLICATION_JSON;
 import static org.sagebionetworks.bridge.BridgeConstants.API_APP_ID;
 import static org.sagebionetworks.bridge.Roles.RESEARCHER;
 import static org.sagebionetworks.bridge.spring.controllers.CRCController.AccountStates.TESTS_SCHEDULED;
 import static org.sagebionetworks.bridge.util.BridgeCollectors.toImmutableSet;
 
+import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -20,6 +22,8 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableMap;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.fluent.Request;
 import org.hl7.fhir.dstu3.model.Address;
 import org.hl7.fhir.dstu3.model.Appointment;
 import org.hl7.fhir.dstu3.model.ContactPoint;
@@ -33,6 +37,8 @@ import org.hl7.fhir.dstu3.model.ProcedureRequest;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.CrossOrigin;
@@ -44,6 +50,7 @@ import org.springframework.web.bind.annotation.RestController;
 import org.sagebionetworks.bridge.BridgeUtils;
 import org.sagebionetworks.bridge.RequestContext;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.exceptions.BridgeServiceException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.NotAuthenticatedException;
 import org.sagebionetworks.bridge.models.DateRangeResourceList;
@@ -64,6 +71,7 @@ import ca.uhn.fhir.parser.IParser;
 @CrossOrigin
 @RestController
 public class CRCController extends BaseController {
+    private static final Logger LOG = LoggerFactory.getLogger(CRCController.class);
     
     static final String TIMESTAMP_FIELD = "state_change_timestamp";
     static final String USER_ID_VALUE_NS = "https://ws.sagebridge.org/#userId";
@@ -192,11 +200,29 @@ public class CRCController extends BaseController {
     
     void createLabOrder(StudyParticipant participant) {
         Patient patient = createPatient(participant);
-        
         IParser parser = FHIR_CONTEXT.newJsonParser();
+        String json = parser.encodeResourceToString(patient);
         
         // Call external partner here and submit the patient record
         // this will trigger workflow at Columbia.
+        /*
+        try {
+            HttpResponse response = Request.Post("<url unknown>")
+                    .bodyString(json, APPLICATION_JSON)
+                    .execute()
+                    .returnResponse();
+            if (response.getStatusLine().getStatusCode() != 200 && 
+                response.getStatusLine().getStatusCode() != 201) {
+                LOG.error("Error submitting patient record to CUIMC for user " + participant.getId());
+                throw new BridgeServiceException("The server encountered an error");
+            }
+            // Anything to persist here?
+            LOG.info("Patient record submitted to CUIMC for user " + participant.getId());
+        } catch (IOException e) {
+            LOG.error("Error submitting patient record to CUIMC for user " + participant.getId());
+            throw new BridgeServiceException("The server encountered an error");
+        }
+        */
     }
 
     private String findUserId(List<Identifier> identifiers) {
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -1,0 +1,303 @@
+package org.sagebionetworks.bridge.spring.controllers;
+
+import static com.google.common.net.HttpHeaders.AUTHORIZATION;
+import static java.lang.Boolean.TRUE;
+import static org.sagebionetworks.bridge.Roles.RESEARCHER;
+import static org.sagebionetworks.bridge.util.BridgeCollectors.toImmutableSet;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.apache.commons.codec.binary.Base64;
+import org.hl7.fhir.r4.model.Address;
+import org.hl7.fhir.r4.model.ContactPoint;
+import org.hl7.fhir.r4.model.ContactPoint.ContactPointSystem;
+import org.hl7.fhir.r4.model.Enumerations.AdministrativeGender;
+import org.hl7.fhir.r4.model.HumanName;
+import org.hl7.fhir.r4.model.Patient;
+import org.joda.time.LocalDate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.CrossOrigin;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.RequestContext;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.exceptions.NotAuthenticatedException;
+import org.sagebionetworks.bridge.models.CriteriaContext;
+import org.sagebionetworks.bridge.models.StatusMessage;
+import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
+import org.sagebionetworks.bridge.models.accounts.SignIn;
+import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
+import org.sagebionetworks.bridge.models.accounts.UserSession;
+import org.sagebionetworks.bridge.models.accounts.UserSessionInfo;
+import org.sagebionetworks.bridge.models.apps.App;
+import org.sagebionetworks.bridge.models.reports.ReportData;
+import org.sagebionetworks.bridge.services.ParticipantService;
+import org.sagebionetworks.bridge.services.ReportService;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.parser.IParser;
+
+@CrossOrigin
+@RestController
+public class CRCController extends BaseController {
+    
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,367 @@
...
+        this.reportService = reportService;
+    }
+    
+    DateTime getTimestamp() {
+        return DateTime.now().withZone(DateTimeZone.UTC);
+    }
+    
+    @PostMapping("/v1/cuimc/participants/self")
+    public JsonNode updateParticipantSelf() {
+        UserSession session = getAuthenticatedSession(RESEARCHER);
```
DwayneJengSage, line 126: I think I'm missing something here. Is this really a researcher API? It looks like the researcher is modifying their own participant and ordering tests for themselves.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,587 @@
...
+        
+        JsonNode retValue = controller.updateParticipantSelf();
+        assertNotNull(retValue);
+        
+        verify(mockParticipantService).updateParticipant(eq(app), participantCaptor.capture());
+        verify(controller).createLabOrder(any()); // expand on this
+        verify(mockSessionUpdateService).updateParticipant(eq(session), any(), any(StudyParticipant.class));
+        
+        StudyParticipant captured = participantCaptor.getValue();
+        assertEquals(captured.getDataGroups(), makeSetOf(CRCController.AccountStates.TESTS_REQUESTED));
```
DwayneJengSage: Should also verify that we retain the user ID.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,587 @@
...
+    
+    @AfterMethod
+    public void afterMethod() {
+        BridgeUtils.setRequestContext(RequestContext.NULL_INSTANCE);
+    }
+    
+    @Test
+    public void updateParticipantCallsExternal() throws Exception {
+        StudyParticipant participant = new StudyParticipant.Builder()
+                .withDataGroups(makeSetOf(CRCController.AccountStates.SELECTED)).build();
```
DwayneJengSage: Is it worth adding data groups that aren't account states to verify that these are correctly preserved?

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,587 @@
...
+        session.setAppId(APP_ID);
+        session.setParticipant(new StudyParticipant.Builder().withId(USER_ID).build());
+        doReturn(session).when(controller).getAuthenticatedSession(RESEARCHER);
+
+        when(mockParticipantService.getParticipant(app, USER_ID, true))
+                .thenReturn(new StudyParticipant.Builder().build());
+        
+        controller.updateParticipantSelf();
+        
+        verify(controller, never()).createLabOrder(any());
```
DwayneJengSage, line 198: Should also verify that the account remains in the DECLINED state?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,367 @@
...
+        
+        CriteriaContext context = new CriteriaContext.Builder()
+                .withLanguages(session.getParticipant().getLanguages())
+                .withClientInfo(reqContext.getCallerClientInfo())
+                .withHealthCode(session.getHealthCode())
+                .withUserId(session.getId())
+                .withUserDataGroups(updatedParticipant.getDataGroups())
+                .withUserSubstudyIds(updatedParticipant.getSubstudyIds())
+                .withAppId(session.getAppId())
+                .build();
```
DwayneJengSage: I know this is replicated from ParticipantController, but would we be better off simplifying and removing the attributes that aren't used in CRC? This would allow us to keep the tests simple and require less maintenance when we have to change things.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,587 @@
...
+                .withDataGroups(makeSetOf(CRCController.AccountStates.SELECTED)).build();
+        mockRequestBody(mockRequest, participant);
+        
+        UserSession session = new UserSession();
+        session.setAppId(APP_ID);
+        session.setParticipant(new StudyParticipant.Builder().withId(USER_ID).build());
+        doReturn(session).when(controller).getAuthenticatedSession(RESEARCHER);
+
+        when(mockParticipantService.getParticipant(app, USER_ID, true))
+                .thenReturn(new StudyParticipant.Builder().build());
```
DwayneJengSage: Should we put request context attributes in this participant, and verify that they are being correctly propagated to the request context?

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,587 @@
...
+        controller.updateParticipantSelf();
+        
+        verify(controller, never()).createLabOrder(any());
+    }    
+    
+    @Test
+    public void postAppointment() throws Exception {
+        when(mockRequest.getHeader(AUTHORIZATION)).thenReturn(AUTHORIZATION_HEADER_VALUE);
+        when(mockAccountService.authenticate(any(), any())).thenReturn(account);
+        when(mockAccountService.getAccount(ACCOUNT_ID)).thenReturn(account);
```
DwayneJengSage, line 209: Should the account have other data groups other than AccountState to verify that they are properly preserved?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,381 @@
...
+        mutableGroups.add(state.name().toLowerCase());
+        account.setDataGroups(mutableGroups);
+
+        account.getAttributes().put(TIMESTAMP_FIELD, getTimestamp().toString());
+    }
+    
+    // This is bound to one specific account that we've given to Columbia. 
+    // No one else can authenticate for these calls, and the account itself
+    // has no administrative roles so it can't do anything else besides 
+    // calling these methods.
```
DwayneJengSage, line 291: How do we ensure this works when we test it in local, dev, staging, and prod? Do we need to hack up our database? Would it be better to have this be config driven?

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,665 @@
...
+        assertEquals(retValue.getStatusCodeValue(), 200);
+    }
+
+    @Test
+    public void authenticationWorks() {
+        when(mockRequest.getHeader(AUTHORIZATION)).thenReturn(AUTHORIZATION_HEADER_VALUE);
+        when(mockAppService.getApp(CRCController.APP_ID)).thenReturn(mockApp);
+        when(mockAccountService.authenticate(eq(mockApp), signInCaptor.capture())).thenReturn(account);
+        
+        controller.httpBasicAuthentication();
```
DwayneJengSage, line 381: Do we want to check that substudies are propagated properly?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,381 @@
...
+        return ResponseEntity.created(URI.create("/v1/cuimc/observations/" + userId))
+                .body(new StatusMessage("Observation created."));        
+    }
+    
+    void createLabOrder(StudyParticipant participant) {
+        Patient patient = createPatient(participant);
+        
+        IParser parser = FHIR_CONTEXT.newJsonParser();
+        
+        System.out.println(parser.encodeResourceToString(patient));
```
DwayneJengSage: Does this print statement leak PHI into the logs?

DwayneJengSage (changes requested):
As per stand-up, I'll leave it to your judgement which comments need to be fixed, and which comments can be deferred in favor of hitting our schedule promises.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: This call will be made from our own UI by someone on Wendy's team, not from our external partner, and is the same as an updateParticipant call, except we will call our external partner if the account is tagged with a specific tag, requesting that they contact the user to set up an appointment.

alxdarksage: This call should be a researcher call, I've adjusted the name of the method and the logic to follow ParticipantController.updateParticipant(), along with the special logic here to contact the lag.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Added the group "group1" and verification that it is not added/removed

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: This has been removed because this should be a researcher call that changes a participant account... it does not need to return an updated session.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: This has been removed as per comment above.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Yes, I have added "group1" and verify that it remains unchanged by these calls.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: Good question! I have changed this so there are two external IDs that are hard-coded to the real study, and to API for testing purposes. In the CRCTest in our integration tests, it creates this user in the api study with the test externalID, and is then able to make these calls. In production we have the production account that will only operate in the cxi-coronavirus study.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: I'm not sure what you mean here...

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: Definitely, this has to be replaced with a call to the external service... I will remove it, and this controller is not complete to deploy until we get information from them on how to post this patient record.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,372 @@
...
+        
+        // Verify the password
+        Account account = accountService.authenticate(app, signIn);
+
+        // This method of verification entirely sidesteps RequestContext 
+        // initialization. Set up anything that is needed in this controller.
+        BridgeUtils.setRequestContext(new RequestContext.Builder()
+                .withCallerAppId(appId)
+                .withCallerSubstudies(BridgeUtils.collectSubstudyIds(account))
+                .build());
```
DwayneJengSage, line 325: We have code here that sets the app ID and substudies into the request context, but this isn't tested anywhere. Should we add a test for this?

src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
```diff
@@ -53,6 +53,7 @@ void setLogger(Logger log) {
     }
 
     private void logException(HttpServletRequest request, Throwable throwable) {
+        throwable.printStackTrace();
```
DwayneJengSage: Remove this print statement.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -0,0 +1,372 @@
...
+    final void setReportService(ReportService reportService) {
+        this.reportService = reportService;
+    }
+    
+    DateTime getTimestamp() {
+        return DateTime.now().withZone(DateTimeZone.UTC);
+    }
+    
+    @PostMapping("/v1/cuimc/participants/{userId}")
+    public StatusMessage updateParticipant(@PathVariable String userId) {
```
DwayneJengSage, line 125: This method takes in the userId but never uses it.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,671 @@
...
+    
+    @Test
+    public void updateParticipantCallsExternal() throws Exception {
+        StudyParticipant participant = new StudyParticipant.Builder()
+                .withDataGroups(makeSetOf(CRCController.AccountStates.SELECTED, "group1")).build();
+        mockRequestBody(mockRequest, participant);
+        
+        UserSession session = new UserSession();
+        session.setAppId(APP_ID);
+        session.setParticipant(new StudyParticipant.Builder().withId(USER_ID).build());
```
DwayneJengSage, line 168: Should use a different USER_ID for the researcher for clarity. (For both versions of the test.)

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: I feel this is tested by the post* methods, which fail if the setting of RequestContext is removed (either of the two values will cause tests to fail for lack of that value in other methods).

src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
alxdarksage: Removed

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: This is definitely a bug. Will also add a test of this method in the integration tests (and fixed it here).

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: This is definitely a bug. Will also add a test of this method in the integration tests (and fixed it here).

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Now uses a different user id.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -0,0 +1,669 @@
...
+        doReturn(session).when(controller).getAuthenticatedSession(RESEARCHER);
+
+        when(mockParticipantService.getParticipant(app, USER_ID, true))
+                .thenReturn(new StudyParticipant.Builder().build());
+        
+        controller.updateParticipant(USER_ID);
```
DwayneJengSage: Should also use a different user ID in this test as well.

DwayneJengSage (approved):
Looks good. One comment about test changes.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Changed to use different user ID.

DwayneJengSage (approved):

PR #205, opened 2020-05-18
Fix a couple of things that findbugs complains about.

alxdarksage:


DwayneJengSage:
btw, you can run FindBugs locall with `mvn compile findbugs:check`. To run it with a friendly GUI, run `mvn compile findbugs:findbugs findbugs:gui`. It's also run as part of `mvn verify` which is the same build that's run in Travis. This will help you test everything quickly without waiting for a Travis build.

This is all documented in README.md.

PR #208, opened 2020-05-20
Changing the participant call so a Synapse process can trigger lab orders.

alxdarksage:
In addition, we are writing the JSON we get from Columbia as health data, using the synchronous health data service.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -3,25 +3,31 @@ package org.sagebionetworks.bridge.spring.controllers;
 import static com.google.common.net.HttpHeaders.AUTHORIZATION;
 import static java.lang.Boolean.TRUE;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.http.entity.ContentType.APPLICATION_JSON;
 import static org.sagebionetworks.bridge.BridgeConstants.API_APP_ID;
-import static org.sagebionetworks.bridge.Roles.RESEARCHER;
+import static org.sagebionetworks.bridge.BridgeConstants.TEST_USER_GROUP;
+import static org.sagebionetworks.bridge.BridgeUtils.parseAccountId;
 import static org.sagebionetworks.bridge.spring.controllers.CRCController.AccountStates.TESTS_SCHEDULED;
 import static org.sagebionetworks.bridge.util.BridgeCollectors.toImmutableSet;
 
+import java.io.IOException;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.google.common.collect.ImmutableList;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.net.HttpHeaders;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.fluent.Request;
 import org.hl7.fhir.dstu3.model.Address;
 import org.hl7.fhir.dstu3.model.Appointment;
 import org.hl7.fhir.dstu3.model.ContactPoint;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -189,47 +208,48 @@ public StatusMessage updateParticipant(@PathVariable String userId) {
...
-                    .bodyString(json, APPLICATION_JSON)
-                    .execute()
-                    .returnResponse();
+            HttpResponse response = post(json);
             if (response.getStatusLine().getStatusCode() != 200 && 
                 response.getStatusLine().getStatusCode() != 201) {
-                LOG.error("Error submitting patient record to CUIMC for user " + participant.getId());
-                throw new BridgeServiceException("The server encountered an error");
+                LOG.error(EXT_ERROR + account.getId());
+                throw new ServiceUnavailableException(UNAVAILABLE_ERROR);
```
DwayneJengSage: Would it make sense to throw an HTTP error that matches the back-end's response? ie, if they throw a 4XX, we throw a 400 Bad Request, if they throw a 5XX, we throw a 500 Internal Server error?

Similarly, if it's a 4XX, should we log a warning and if it's a 5XX, we should log an error. Or maybe just always log warnings. In our logs, errors should be actionable by the Bridge dev team.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -189,47 +208,48 @@ public StatusMessage updateParticipant(@PathVariable String userId) {
...
             }
-            // Anything to persist here?
-            LOG.info("Patient record submitted to CUIMC for user " + participant.getId());
+            LOG.info("Patient record submitted to CUIMC for user " + account.getId());
         } catch (IOException e) {
-            LOG.error("Error submitting patient record to CUIMC for user " + participant.getId());
-            throw new BridgeServiceException("The server encountered an error");
+            LOG.error(EXT_ERROR + account.getId());
```
DwayneJengSage: Similarly, we should log a different error (or warning) so we can differentiate between this error and the previous error.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -189,47 +208,48 @@ public StatusMessage updateParticipant(@PathVariable String userId) {
...
+            throw new ServiceUnavailableException(UNAVAILABLE_ERROR);
         }
-        */
+    }
+    
+    HttpResponse post(String bodyJson) throws IOException {
+        return Request.Post("<url unknown>")
+                .bodyString(bodyJson, APPLICATION_JSON)
+                .execute()
+                .returnResponse();
```
DwayneJengSage, line 264: Consider disabling this altogether until we have a real URL. This way, we won't get spurious errors when someone calls this API and forget to tag the participant with test_user.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -324,33 +350,39 @@ void httpBasicAuthentication() {
...
+        // This method of verification sidesteps RequestContext initialization
+        // through a session. Set up what is needed in the controller.
+        Set<String> substudies = BridgeUtils.collectSubstudyIds(account);
+        
+        RequestContext.Builder builder = new RequestContext.Builder().withCallerAppId(appId);
+        // Sage account can see all participants, others will only have access to their 
+        // substudy participants.
+        if (!substudies.contains("sage")) {
+            builder.withCallerSubstudies(substudies);
+        }
```
DwayneJengSage: This breaks our existing substudy model, and will cause other APIs to behave abnormally. Is this what we want?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: I don't have strong feelings about this because I haven't seen their interface. Hopefully we catch any issues that would generate a 4xx series of errors, but I'll update this.

alxdarksage: From the end user's perspective, there's nothing they should be able to do in our API that would cause their API to return a 3xx or 4xx error (if these happen, it is a programming error on the part of the Bridge server). 5xx errors are not distinguishable from the end user's perspective (whether it came from us or Columbia doesn't matter). I would suggest:

3xx and 4xx - we convert to 500 internal server error, and log it as error
5xx - we pass these through because 503 is temporary, but log it as error

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -189,46 +208,56 @@ public StatusMessage updateParticipant(@PathVariable String userId) {
...
-            LOG.error("Error submitting patient record to CUIMC for user " + participant.getId());
-            throw new BridgeServiceException("The server encountered an error");
+            LOG.error(INT_ERROR + account.getId());
+            throw new ServiceUnavailableException(UNAVAILABLE_ERROR);
         }
+        LOG.info("Patient record submitted to CUIMC for user " + account.getId());
+    }
+    
+    HttpResponse post(String bodyJson) throws IOException {
+        throw new UnsupportedOperationException();
```
DwayneJengSage: If our partners use this API but forget to tag the test user with test_user, this API will throw a 500 Internal Server Error and also log an error in our logs.

Do we want this error to be visible to our partners? If so, we should throw a 400 Bad Request (or another similar 4XX error). If we want this to be internal use only, we should log an info or a warning.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -324,33 +359,39 @@ void httpBasicAuthentication() {
...
+        // This method of verification sidesteps RequestContext initialization
+        // through a session. Set up what is needed in the controller.
+        Set<String> substudies = BridgeUtils.collectSubstudyIds(account);
+        
+        RequestContext.Builder builder = new RequestContext.Builder().withCallerAppId(appId);
+        // Sage account can see all participants, others will only have access to their 
+        // substudy participants.
+        if (!substudies.contains("sage")) {
+            builder.withCallerSubstudies(substudies);
+        }
```
DwayneJengSage: This breaks our existing substudy model, and will cause other APIs to behave abnormally. Is this what we want?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: We can't merge this until we have information to integrate with them... I'm just verifying at the moment that tests are not calling this method unmocked. I can remove this... the controller isn't complete until Columbia gets back to us with API information.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: Switching the Synapse and test accounts to use an email address, and removing this exception.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: Changed to use an email credential rather than an external ID, so Sage accounts can be unbound from a substudy. This gives these accounts access to Columbia's external ID accounts.

DwayneJengSage (approved):

PR #214, opened 2020-06-03
Assessment in app config

alxdarksage:
In addition:
- refactored how we reference BridgeConfig as a static in these reference objects so they can be mocked without PowerMock (once you're using PowerMock, it feels like your design is failing);
- added validation that references are not duplicates within the app config.


src/main/java/org/sagebionetworks/bridge/services/AppConfigService.java
```diff
@@ -206,6 +256,7 @@ public AppConfig createAppConfig(String appId, AppConfig appConfig) {
         newAppConfig.setSchemaReferences(appConfig.getSchemaReferences());
         newAppConfig.setConfigReferences(appConfig.getConfigReferences());
         newAppConfig.setFileReferences(appConfig.getFileReferences());
+        newAppConfig.setAssessmentReferences(appConfig.getAssessmentReferences());
```
DwayneJengSage, line 259: Unit test to make sure that assessement references are being marshalled when creating?

DwayneJengSage (approved):
Looks good. One comment about tests for createAppConfig().

src/main/java/org/sagebionetworks/bridge/services/AppConfigService.java
alxdarksage: Added

PR #222, opened 2020-06-19
Add request query parameters to the Metrics Interceptor logging

kss39:
https://sagebionetworks.jira.com/browse/BRIDGE-2349

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
@@ -3,6 +3,7 @@ package org.sagebionetworks.bridge.models;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
+import com.google.common.collect.Multimap;
 import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.time.DateUtils;
 
@@ -128,9 +129,9 @@ public class Metrics {
      *
      * @param paramsMap The query parameters.
      */
-    public void setQueryParams(Map<String, List<String>> paramsMap) {
+    public void setQueryParams(Multimap<String, String> paramsMap) {
         if (paramsMap != null && !paramsMap.isEmpty()) {
-            ObjectNode paramsJson = MAPPER.valueToTree(paramsMap);
+            ObjectNode paramsJson = MAPPER.valueToTree(paramsMap.asMap());
             json.set("query_params", paramsJson);
         }
     }
--- a/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
@@ -3,12 +3,15 @@ package org.sagebionetworks.bridge.models;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
+import org.apache.http.NameValuePair;
 import org.joda.time.DateTime;
 import org.sagebionetworks.bridge.time.DateUtils;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.sagebionetworks.bridge.json.JsonUtils;
+
+import java.util.List;
 
 /**
  * Request-scoped metrics.
@@ -118,6 +121,19 @@ public class Metrics {
 
     public void setUploadSize(long uploadSize) {
         json.put("upload_size", uploadSize);
+    }
+
+    /**
+     * Set the query params from the url request to json.
+     *
+     * @param params The query parameters.
+     */
+    public void setQueryParams(List<NameValuePair> params) {
+        ObjectNode paramNode = MAPPER.createObjectNode();
+        for (NameValuePair pair : params) {
+            paramNode.put(pair.getName(), pair.getValue());
+        }
+        json.set("query_params", paramNode);
     }
 
     private void put(final String field, final String value) {
--- a/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/Metrics.java
@@ -133,13 +133,15 @@ public class Metrics {
      * @param params The query parameters.
      */
     public void setQueryParams(List<NameValuePair> params) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/models/Metrics.java
```diff
@@ -120,6 +127,23 @@ public void setUploadSize(long uploadSize) {
...
+        if (params != null) {
+            ObjectNode paramNode = MAPPER.createObjectNode();
+            for (NameValuePair pair : params) {
+                if (ALLOW_LIST.contains(pair.getName())) {
+                    paramNode.put(pair.getName(), pair.getValue());
+                }
+            }
+            json.set("query_params", paramNode);
+        }
+    }
```
DwayneJengSage: I would recommend moving this to MetricsFilter and replacing this with a simple setter that takes in a Map and writes it to JSON.

The reason for this is two-fold:
1. Keeping data classes separate from functionality classes makes it easier to reason about and test software.
2. Keeping the HTTP stuff in MetricsFilter allows a useful abstraction layer that would (for example) allow us to use Metrics using a different framework.

src/test/java/org/sagebionetworks/bridge/models/MetricsTest.java
```diff
@@ -116,6 +122,34 @@ public void testSetSession() {
...
     }
 
+    @Test
+    public void testSetQueryParams() {
+        // Test empty params metrics.
+        String requestId = "12345";
+        Metrics metrics = new Metrics(requestId);
+        metrics.setQueryParams(null);
+        String json = metrics.toJsonString();
+        assertFalse(json.contains("\"query_params\":"));
```
DwayneJengSage: I recommend using BridgeObjectMapper.get().readTree() to convert this into a JsonNode instead of parsing strings. There are examples in this class that do this. (There are also examples in this class that don't. Ignore those. We should probably fix those at some point.)

src/test/java/org/sagebionetworks/bridge/models/MetricsTest.java
```diff
@@ -116,6 +122,34 @@ public void testSetSession() {
...
+
+        // Test parsed, but no query params metrics.
+        List<NameValuePair> testList = new ArrayList<>(URLEncodedUtils.parse("", StandardCharsets.UTF_8));
+        metrics.setQueryParams(testList);
```
DwayneJengSage: I recommend creating a new Metrics object for each test case. This will guarantee that you have a fresh Metrics object and ensure previous tests don't mess up subsequent tests.

src/test/java/org/sagebionetworks/bridge/spring/filters/MetricsFilterTest.java
```diff
@@ -4,8 +4,7 @@
 import static org.sagebionetworks.bridge.BridgeConstants.X_FORWARDED_FOR_HEADER;
 import static org.sagebionetworks.bridge.BridgeConstants.X_REQUEST_ID_HEADER;
 import static org.sagebionetworks.bridge.TestConstants.TIMESTAMP;
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.*;
```
DwayneJengSage: Our current coding convention is to always expand wildcards for imports. The reasons are two-fold:
1. Historically, importing a wild card would cause Java to pull in all of the imports, instead of just the ones we need. This is no longer true in Java 8 and probably also not true for static imports, but we do this for historical reasons.
2. More importantly, listing out imports makes it unambiguously clear which classes and which methods come from which imports. This is mostly for documentation, but it could also prevent compile errors or runtime errors in the event org.testng.Assert decided to add a new static function that conflicted with one of ours.

src/test/java/org/sagebionetworks/bridge/spring/filters/MetricsFilterTest.java
```diff
@@ -62,7 +61,8 @@ public void metricsTest() throws Exception {
         when(mockRequest.getRemoteAddr()).thenReturn(TestConstants.IP_ADDRESS);
         when(mockRequest.getHeader(USER_AGENT)).thenReturn("userAgent/1");
         when(mockResponse.getStatus()).thenReturn(201);
-        
+        when(mockRequest.getQueryString()).thenReturn("consents=true&email=should_not_leak@fake.com");
```
DwayneJengSage: Consider adding query params with %20 and + to make sure we're calling the right URL Decode method.

src/main/resources/BridgeServer2.conf
```diff
@@ -186,3 +186,7 @@ cuimc.prod.url = http://34.192.153.148:8080/nypcovi/Patient/${patientId}
 cuimc.prod.username =
 cuimc.prod.password =
 
+# The allowlist of URL query parameters.
+# Other parameters in the query will not show up in the server log,
+# in order to protect PII.
+query.param.allowlist = includeDeleted,physical,format,summary,startTime,endTime,pageSize,offsetKey,offsetBy,tag,catagory,minRevision,maxRevision,queryParam,createAccount,createdOnStart,createdOnEnd,consents,scheduledOnStart,scheduledOnEnd,startDate,endDate,deleteReauthToken,until,daysAhead,minimumPerSchedule,mostRecent,mostrecent,published,newSchemaRev,synchronous,redrive
```
DwayneJengSage: Typo: Should be "category", not "catagory".

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -48,7 +55,18 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
+        for (NameValuePair pair : params) {
+            String key = pair.getName();
+            String value = pair.getValue();
+            if (ALLOW_LIST.contains(key)) {
+                if (!paramsMap.containsKey(key)) {
+                    paramsMap.put(key, new LinkedList<>());
+                }
+                paramsMap.get(key).add(value);
+            }
+        }
```
DwayneJengSage: Consider using a Guava Multimap to simplify this code https://github.com/google/guava/wiki/NewCollectionTypesExplained#multimap

DwayneJengSage (changes requested):
Looks good. Just one last comment about data structures.

DwayneJengSage (approved):

PR #229, opened 2020-06-29
Assessment ownership now expressed by association to an organization.

alxdarksage:
Edits to assessments can either be done by a superadmin, or by a member of the organization that owns the assessment (the creator of the assessment sets the organization ownership). 

In addition, it became apparent in integration tests that failing to update a signed in user's session was a significant security hole. If an admin removes a person from the organization, they could still operate as a member of that organization until their session expired, which really does not seem right. Expecting the admin to delete their session is depending too much on user behavior to ensure security. So I am editing the session for the user if they have one. This is something that we may wish to do elsewhere.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -1,13 +1,22 @@
 package org.sagebionetworks.bridge;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.sagebionetworks.bridge.BridgeConstants.CALLER_NOT_MEMBER_ERROR;
+import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.SUPERADMIN;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
 
 public class AuthUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(AuthUtils.class);
     
     /**
-     * Unless you are a SUPERADMIN, you can only list the members of your own organization.
+     * Unless you are a superadmin, you can only list the members of your own organization.
      */
     public static boolean checkOrgMembership(String proposedOrgId) {
         RequestContext context = BridgeUtils.getRequestContext();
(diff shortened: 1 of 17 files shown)


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -21,4 +30,52 @@ public static void checkOrgMembershipAndThrow(String proposedOrgId) {
...
+     * If the caller is an app admin or superadmin, they can set any organization. Otherwise, the 
+     * organization must match the caller's organization. We do not need to validate the org ID since 
+     * it was validated when it was set as an organizational relationship on the account. 
+     */
+    public static void checkAssessmentOwnership(String appId, String ownerId) {
+        RequestContext rc = BridgeUtils.getRequestContext();
+        String callerOrgMembership = rc.getCallerOrgMembership();
+        if (isBlank(ownerId)) {
+            throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+        }
```
DwayneJengSage: What does it mean if an assessment has no owner ID? Is that a caller error or a server error?

If not having an owner ID is possible and not an error, should superadmins have access to assessments without owners?

Similarly for checkSharedAssessmentOwnership()

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -21,4 +30,52 @@ public static void checkOrgMembershipAndThrow(String proposedOrgId) {
...
+        if (ownerId == null) {
+            LOG.error("Owner ID is null, guid=" + guid);
+            throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+        }
+        String[] parts = ownerId.split(":", 2);
+        // This happens in tests, we expect it to never happens in production. So log if it does.
+        if (parts.length != 2) {
+            LOG.error("Could not parse shared assessment ownerID, guid=" + guid + ", ownerId=" + ownerId);
+            throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+        }
```
DwayneJengSage, line 71: If these can happen in tests, should superadmins have access to these?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: It should always have an ownerId, so this would be a server error. Removed and replaced with a checkNotNull().

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: Good point. Changed this so they are always accessible to superadmins.

DwayneJengSage (approved):

PR #230, opened 2020-07-01
Changes to reflect changes made by partner; don't process production accounts.

alxdarksage:


src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -349,29 +353,41 @@ private Request addAuthorizationHeader(Request request, Account account) {
...
+        }
+        throw new BadRequestException("Could not find Bridge user ID.");
+    }
 
-    private String findUserId(Reference ref) {
-        if (ref != null && ref.getReference() != null) {
-            String value = ref.getReference();
-            if (value.toLowerCase().startsWith("patient/")) {
-                return value.substring(8);
+    private String findUserId(Reference subject) {
```
nategbrown9, line 386: Is there a reason the Reference is called a subject here and is called an actor in the method above?

nategbrown9 (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: Only because the accessor in two of the cases is getSubject() and in the other case is getActor()... despite being labeled as something different, they're both Reference objects.

PR #233, opened 2020-07-01
The new search parameters should be returned in the request parameters of ResourceList.

alxdarksage:
As it turns out my problems were in the study manager, but it highlighted that I missed these and could not see how the server was interpreting what I was sending. Added.


nategbrown9 (approved):
Looks good.

PR #236, opened 2020-07-07
Substudy to study

alxdarksage:
Study v5. The change from substudy is not backwards compatible, but I have not identified any clients that will break when we make this change.


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoCriteria.java
```diff
@@ -112,20 +112,20 @@ public void setNoneOfGroups(Set<String> noneOfGroups) {
     @Override
     @DynamoDBAttribute
     @DynamoDBTypeConverted(converter=StringSetMarshaller.class)
-    public Set<String> getAllOfSubstudyIds() {
-        return allOfSubstudyIds;
+    public Set<String> getAllOfStudyIds() {
+        return allOfStudyIds;
```
DwayneJengSage: We have a lot of DynamoDB classes where we rename substudyID. Should we add the annotation to keep the old name? This would allow us to read old data without requiring migrating thousands of rows. (In some cases, renaming the field might not be possible, since this would cause conflict with app/study.)

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateEnrollment.java
```diff
@@ -19,7 +19,7 @@
     @Id
     private String appId;
     @Id
-    private String substudyId;
+    private String studyId;
```
DwayneJengSage: Similarly, would this also change the column name for Hibernate tables?

src/main/java/org/sagebionetworks/bridge/models/RequestInfo.java
```diff
@@ -92,8 +93,8 @@ public String getUserAgent() {
         return userDataGroups;
     }
 
-    public Set<String> getUserSubstudyIds() {
-        return userSubstudyIds;
+    public Set<String> getUserStudyIds() {
+        return userStudyIds;
     }
```
DwayneJengSage, line 99: Need a JsonAlias, or deserializing this from Redis will fail.

src/main/java/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
```diff
@@ -210,8 +210,8 @@ public DateTimeZone getTimeZone() {
     public JsonNode getClientData() {
         return clientData;
     }
-    public Set<String> getSubstudyIds() {
-        return substudyIds;
+    public Set<String> getStudyIds() {
+        return studyIds;
```
DwayneJengSage, line 215: UserSession is in Redis, and UserSession contains StudyParticipant. We'll need a JsonAlias so we can read the old SubstudyIds from Redis.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
```diff
@@ -320,15 +320,15 @@ public void setUserDataGroups(Set<String> userDataGroups) {
     
     /** {@inheritDoc} */
     @Override
-    public Map<String, String> getUserSubstudyMemberships() {
-        return userSubstudyMemberships;
+    public Map<String, String> getUserStudyMemberships() {
+        return userStudyMemberships;
```
DwayneJengSage: The Exporter still reads this field by this name.

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
```diff
@@ -263,8 +263,8 @@ public Assessment getLatestAssessment(String appId, String identifier) {
 
     /**
      * Takes a reference to an assessment in the caller's context (appId), and creates an 
-     * assessment in the shared context. Stores the full substudy identifier (appId + 
-     * substudyId) as the "ownerId" in shared context for authorization checks on 
+     * assessment in the shared context. Stores the full study identifier (appId + 
+     * studyId) as the "ownerId" in shared context for authorization checks on
```
DwayneJengSage: study or org?

src/main/java/org/sagebionetworks/bridge/validators/AssessmentValidator.java
```diff
@@ -75,8 +75,8 @@ public void validate(Object target, Errors errors) {
             }
         }
         
-        // ownerId == substudyId except in the shared assessments app, where it must include
-        // the app as a namespace prefix, e.g. "appId:substudyId". Assessments are always 
+        // ownerId == studyId except in the shared assessments app, where it must include
+        // the app as a namespace prefix, e.g. "appId:studyId". Assessments are always
```
DwayneJengSage: study or org?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoCriteria.java
alxdarksage: Yes, we should keep the old names in all cases... added here, and I'll double check all persistence classes to make sure this is true. (There are a few other places where this needed to be added on Dynamo-persisted classes.)

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateEnrollment.java
alxdarksage: This didn't cause problems running integration tests, but that's surprising. I've added @Column annotations to appId (which is studyId in table) and studyId (which is substudyId in table). At the least, it documents the mapping.

src/main/java/org/sagebionetworks/bridge/models/RequestInfo.java
alxdarksage: Added (with test)

src/main/java/org/sagebionetworks/bridge/models/accounts/StudyParticipant.java
alxdarksage: Added (with addition to oldJsonParsesCorrectly test).

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
alxdarksage: Added a @DynamoDBAttribute to keep the existing column name... but is there a dependency on this Java code, or is it through the table alone?

src/main/java/org/sagebionetworks/bridge/services/AssessmentService.java
alxdarksage: Organization (changed).

src/main/java/org/sagebionetworks/bridge/validators/AssessmentValidator.java
alxdarksage: org (documentation updated)

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataRecord.java
DwayneJengSage: The dependency is through the DynamoDB tables directly.

DwayneJengSage (approved):

PR #237, opened 2020-07-08
Geocode location for CRC appointment record.

alxdarksage:


src/main/resources/BridgeServer2.conf
```diff
@@ -178,6 +178,9 @@ synapse.oauth.url = https://repo-prod.prod.sagebase.org/auth/v1/oauth2/token
 synapse.oauth.client.id = dummy-value
 synapse.oauth.client.secret = dummy-value
 
+# To reverse geocode location of appointments in CRC controller
+crc.geocode.api.key = dummy-value
```
DwayneJengSage, line 182: Do we need a corresponding infra change to deploy this key to AWS?

DwayneJengSage (approved):

src/main/resources/BridgeServer2.conf
alxdarksage: Yes. https://github.com/Sage-Bionetworks/BridgeServer2-infra/pull/34 (I'm not 100% sure if this is correct, but I believe if I set the appropriate environment variable in all 3 environments, it'll work.)

PR #240, opened 2020-07-13
Move setCookieAndRecordMetrics functionality from controllers to filter

kss39:


src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -85,14 +85,18 @@ public JsonNode emailSignIn() {
...
-            setCookieAndRecordMetrics(e.getUserSession());
+        } catch (ConsentRequiredException e) {
+            session = e.getUserSession();
+            updateRequestInfoFromSession(session);
             throw e;
+        } finally {
+            request().setAttribute("CreatedUserSession", session);
         }
-        setCookieAndRecordMetrics(session);
+        updateRequestInfoFromSession(session);
```
DwayneJengSage: Consider moving the updateRequestInfoFromSession() into the finally block, so we only need to call it once. (Here and in all the other sign-in methods.)

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -766,17 +766,18 @@ public void localSignInSetsSessionCookie() throws Exception {
...
+//        verify(mockResponse).addCookie(cookieCaptor.capture());
+//
+//        Cookie cookie = cookieCaptor.getValue();
+//        assertEquals(cookie.getName(), BridgeConstants.SESSION_TOKEN_HEADER);
+//        assertEquals(cookie.getValue(), TEST_SESSION_TOKEN);
+//        assertEquals(cookie.getMaxAge(), BridgeConstants.BRIDGE_SESSION_EXPIRE_IN_SECONDS);
+//        assertEquals(cookie.getPath(), "/");
+//        assertEquals(cookie.getDomain(), DOMAIN);
+//        assertFalse(cookie.isHttpOnly());
+//        assertFalse(cookie.getSecure());
```
DwayneJengSage: Deleted code should be deleted, not commented. (Here and in all the other tests.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/BaseController.java
```diff
@@ -354,19 +354,11 @@ protected void writeSessionInfoToMetrics(UserSession session) {
     /** Combines metrics logging with the setting of the session token as a cookie in local
      * environments (useful for testing).
      */
-    protected void setCookieAndRecordMetrics(UserSession session) {
-        writeSessionInfoToMetrics(session);
```
DwayneJengSage, line 358: What about getSessionIfItExists()?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -85,14 +85,18 @@ public JsonNode emailSignIn() {
...
         try {
             session = authenticationService.emailSignIn(context, signInRequest);
-        } catch(ConsentRequiredException e) {
-            setCookieAndRecordMetrics(e.getUserSession());
+        } catch (ConsentRequiredException e) {
+            session = e.getUserSession();
+            updateRequestInfoFromSession(session);
             throw e;
+        } finally {
+            request().setAttribute("CreatedUserSession", session);
```
DwayneJengSage: If we always call request().setAttribute() at the same time as updateRequestInfoFromSession(), consider putting request().setAttribute() in updateRequestInfoFromSession().

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -61,10 +75,21 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
+        // Used for record metrics from the UserSession.
+        UserSession session = null;
+        // Log session info when a session is present
         try {
             chain.doFilter(req, res);
             metrics.setStatus(response.getStatus());
+            session = (UserSession) request.getAttribute("CreatedUserSession");
+        } catch (ConsentRequiredException e) {
+            session = e.getUserSession();
+            throw e;
```
DwayneJengSage: There's no need to catch the ConsentRequiredException here. We've already set the attribute at the Controllers level.

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -61,10 +75,21 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
 
         metrics.setQueryParams(paramsMap);
 
+        // Used for record metrics from the UserSession.
+        UserSession session = null;
+        // Log session info when a session is present
         try {
             chain.doFilter(req, res);
             metrics.setStatus(response.getStatus());
+            session = (UserSession) request.getAttribute("CreatedUserSession");
```
DwayneJengSage: This can be moved into the finally block.

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -77,6 +102,16 @@ private String header(HttpServletRequest request, String name, String defaultVal
...
 
+    /**
+     * Writes the user's account ID, internal session ID, and app ID to the metrics.
+     * Requires @NotNull for both metrics and session.
+     */
+    private void writeSessionInfoToMetrics(Metrics metrics, UserSession session) {
+        metrics.setSessionId(session.getInternalSessionToken());
+        metrics.setUserId(session.getId());
+        metrics.setAppId(session.getAppId());
+    }
```
DwayneJengSage: Do we need this in two places? If we need it in two places, you should create a utility function somewhere that's called in both places instead of copy-pasting the method.

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
```diff
@@ -38,6 +42,16 @@
     // Allow-list for query parameters metrics logging.
     private static final List<String> ALLOW_LIST =
             BridgeConfigFactory.getConfig().getList("query.param.allowlist");
+
+    /**
+     * Redis CacheProvider to give UserSession objects for Metrics.
+     */
+    CacheProvider cacheProvider;
```
DwayneJengSage: By convention, this is always private. (Same with bridgeConfig in RequestFilter.)

src/main/java/org/sagebionetworks/bridge/spring/filters/RequestFilter.java
```diff
@@ -95,11 +111,23 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
 
         req = new RequestIdWrapper(request, requestId);
+
+        // set local cookies when a session in present
+        UserSession session = null;
         try {
             chain.doFilter(req, res);
+            session = (UserSession) request.getAttribute("CreatedUserSession");
+        } catch (ConsentRequiredException e) {
+            session = e.getUserSession();
```
DwayneJengSage: Similarly here. We don't need to catch the ConsentRequiredException, and the request.getAttribute() can be moved into the finally block.

src/main/java/org/sagebionetworks/bridge/spring/filters/RequestFilter.java
```diff
@@ -113,6 +141,16 @@ protected void setRequestContext(RequestContext context) {
...
 
+    protected Cookie makeSessionCookie(String sessionToken, int expireInSeconds) {
+        Cookie cookie = new Cookie(SESSION_TOKEN_HEADER, sessionToken);
+        cookie.setMaxAge(expireInSeconds);
+        cookie.setPath("/");
+        cookie.setDomain("localhost");
+        cookie.setHttpOnly(false);
+        cookie.setSecure(false);
+        return cookie;
+    }
```
DwayneJengSage: Similarly here. If we need this in two places, we should make a shared utility function.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -1415,14 +1417,14 @@ private void mockJson(String json) throws Exception {
...
         verify(controller, atLeastOnce()).getMetrics();
-        
         verify(metrics, atLeastOnce()).setSessionId(TEST_INTERNAL_SESSION_ID);
         verify(metrics, atLeastOnce()).setUserId(TEST_ACCOUNT_ID);
         verify(metrics, atLeastOnce()).setAppId(TEST_APP_ID);
     }
     
     private void verifyCommonLoggingForSignIns() throws Exception {
-        verifyMetrics();
+        // verifyMetrics();
```
DwayneJengSage: Is verifyMetrics() still being used? If not, can we delete it?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
alxdarksage: In a similar vein, if we're grabbing the session here to update RequestInfo, can this be done along with other processing in a filter? It would be nice to only have to grab the session out of ConsentRequiredException in one place.

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
alxdarksage: Or the inverse, do it here but don't do it multiple places in the controllers.

src/main/java/org/sagebionetworks/bridge/spring/filters/RequestFilter.java
alxdarksage: Hopefully, you don't need the code you've moved into these filters anymore in the controllers, and the controllers can be simplified.

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
kss39: The problem is, `authService.emailSignIn` doesn't only throw `ConsentRequiredException`, it also throws `AuthenticationFailedException` and `EntityNotFoundException`. In the original code, the try/catch block only handles `CRE` instead of other exceptions, so it calls `setCookieAndRecordMetrics` twice rather than in finally block. (For other two Exceptions, no session info is recorded; so session could be `null` if `AFE` or `ENFE` is thrown)

kss39: With regard to other comments below, I'll integrate `updateRequestInfoFromSession` along with setting request attributes together in `BaseController`, so that we don't need to grab `e.getUserSession()` again in Filters.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
kss39: `verifyMetric()` is still being used in `signOut` tests. I will delete this comment line but leave the `verifyMetric` as it is.
(This message is outdated. Now the verifyMetrics() has been removed.)

src/main/java/org/sagebionetworks/bridge/spring/filters/MetricsFilter.java
kss39: Oops, I found that I actually didn't use the `cacheProvider`. So I will delete this part.

src/test/java/org/sagebionetworks/bridge/spring/controllers/UserManagementControllerTest.java
```diff
@@ -163,16 +163,17 @@ public void signInForSuperadmin() throws Exception {
...
+//        verify(mockResponse).addCookie(cookieCaptor.capture());
+//
+//        Cookie cookie = cookieCaptor.getValue();
+//        assertEquals(cookie.getName(), SESSION_TOKEN_HEADER);
+//        assertEquals(cookie.getValue(), session.getSessionToken());
+//        assertEquals(cookie.getMaxAge(), BRIDGE_SESSION_EXPIRE_IN_SECONDS);
+//        assertEquals(cookie.getPath(), "/");
+//        assertEquals(cookie.getDomain(), "localhost");
+//        assertFalse(cookie.isHttpOnly());
+//        assertFalse(cookie.getSecure());
```
DwayneJengSage: Delete these comments too.

src/main/java/org/sagebionetworks/bridge/spring/filters/RequestFilter.java
```diff
@@ -43,6 +51,13 @@
 
 @Component
 public class RequestFilter implements Filter {
+    BridgeConfig bridgeConfig;
```
DwayneJengSage: This should always be private.

src/test/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationControllerTest.java
```diff
@@ -1415,14 +1391,14 @@ private void mockJson(String json) throws Exception {
...
     private void verifyMetrics() {
         verify(controller, atLeastOnce()).getMetrics();
-        
         verify(metrics, atLeastOnce()).setSessionId(TEST_INTERNAL_SESSION_ID);
         verify(metrics, atLeastOnce()).setUserId(TEST_ACCOUNT_ID);
         verify(metrics, atLeastOnce()).setAppId(TEST_APP_ID);
     }
     
     private void verifyCommonLoggingForSignIns() throws Exception {
-        verifyMetrics();
```
DwayneJengSage, line 1425: Is verifyMetrics() still being used anywhere? Can we delete it?

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -749,4 +753,21 @@ public static InvalidEntityException convertParsingError(Throwable throwable) {
...
+     */
+    public static Cookie makeSessionCookie(String sessionToken, int expireInSeconds) {
+        Cookie cookie = new Cookie(SESSION_TOKEN_HEADER, sessionToken);
+        cookie.setMaxAge(expireInSeconds);
+        cookie.setPath("/");
+        cookie.setDomain("localhost");
+        cookie.setHttpOnly(false);
+        cookie.setSecure(false);
+        return cookie;
+    }
```
DwayneJengSage: Consider putting this in HttpUtil in the org.sagebionetworks.bridge.spring.util, so we can keep our Bridge stuff separate from our HTTP stuff.

DwayneJengSage (changes requested):

DwayneJengSage (approved):
Looks good to me.

Also, be sure to test with integration tests and also verify the logs to make sure metrics are still being logged.

PR #242, opened 2020-07-15
Sponsors

alxdarksage:
Add the ability for organizations to sponsor studies. This will be the basis for an expanded permissions model where members of the organization have access to studies through their organization.

A number of security checks that were for superadmins have been changed to admin. Superadmins can already do everything, and by setting permissions to superadmin, I was making the admin role less useful. I am also becoming clearer that the ADMIN role is an admin for an entire app. I do not know how we will indicate a study admin or organization admin without introducing a stronger permissions model, but ADMIN doesn't mean those things.


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyDao.java
```diff
@@ -76,6 +76,12 @@ public void deleteStudyPermanently(String appId, String id) {
         checkNotNull(appId);
         checkNotNull(id);
         
+        QueryBuilder builder = new QueryBuilder();
+        builder.append("DELETE FROM OrganizationsStudies WHERE appId = :appId AND studyId = :studyId");
+        builder.getParameters().put("appId", appId);
+        builder.getParameters().put("studyId", id);
+        hibernateHelper.nativeQueryUpdate(builder.getQuery(), builder.getParameters());
```
DwayneJengSage: If you specify ON DELETE CASCADE in the Foreign Key in SQL, then SQL should automatically delete the row from OrganizationsStudies when you delete the study (or the org, for that matter).

src/main/resources/db/changelog/changelog.sql
```diff
@@ -374,3 +374,23 @@ ADD COLUMN `orgMembership` varchar(255),
...
+-- changeset bridge:18
+
+-- Sponsors. Named consistent with other associative tables.
+CREATE TABLE IF NOT EXISTS `OrganizationsStudies` (
+  `appId` varchar(255) NOT NULL,
+  `studyId` varchar(255) NOT NULL,
+  `orgId` varchar(255) NOT NULL,
+  PRIMARY KEY (`appId`,`studyId`,`orgId`),
+  CONSTRAINT `fk_os_organization` FOREIGN KEY (`appId`, `orgId`) REFERENCES `Organizations` (`appId`, `identifier`),
+  CONSTRAINT `fk_os_study` FOREIGN KEY (`studyId`, `appId`) REFERENCES `Substudies` (`id`, `studyId`)
```
DwayneJengSage: ON DELETE CASCADE?

src/main/java/org/sagebionetworks/bridge/hibernate/PersistenceExceptionConverter.java
```diff
@@ -1,8 +1,29 @@
...
+import com.google.common.base.Throwables;
+
 public interface PersistenceExceptionConverter {
+    
+    /**
+     * With Hibernate, it appears that PersistenceException wraps HQL queries, but not native queries. 
+     * This leads to inconsistent exception unwrapping in the converters, which this method eliminates. 
+     * If a DAO mixes the two types of queries, the converter should work for both.
+     */
+    public static org.hibernate.exception.ConstraintViolationException getConstraintViolation(PersistenceException exception) {
```
DwayneJengSage: Does this need unit tests?

src/test/java/org/sagebionetworks/bridge/hibernate/SponsorPersistenceExceptionConverterTest.java
```diff
@@ -0,0 +1,85 @@
...
+        this.converter = new SponsorPersistenceExceptionConverter();
+    }
+    
+    @Test
+    public void repackagesConstraintViolation() { 
+        org.hibernate.exception.ConstraintViolationException cve = new org.hibernate.exception.ConstraintViolationException("Some error", null, "");
+        PersistenceException pe = new PersistenceException(cve);
+        
+        RuntimeException retValue = converter.convert(pe, Organization.create());
+        assertEquals(retValue.getClass().getName(), "org.sagebionetworks.bridge.exceptions.ConstraintViolationException");
```
DwayneJengSage: Minor: Could you check that the classes are equal so you don't have to do string equality? (I think you can even do == instead of equals().)

src/test/java/org/sagebionetworks/bridge/hibernate/SponsorPersistenceExceptionConverterTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    
+    private void constraintTests(Class<?> clazz, String sqlError, String finalMessage) {
+        SQLIntegrityConstraintViolationException sqle = new SQLIntegrityConstraintViolationException(sqlError);
+        org.hibernate.exception.ConstraintViolationException cve = new org.hibernate.exception.ConstraintViolationException("", sqle, "");
+        PersistenceException pe = new PersistenceException(cve);
+        
+        RuntimeException retValue = converter.convert(pe, null);
+        
+        assertEquals(retValue.getClass().getName(), clazz.getCanonicalName());
```
DwayneJengSage: Similarly, can you just check class equality here?

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -200,10 +200,6 @@ public void removeMember(String appId, String identifier, AccountId accountId) {
         
         AuthUtils.checkOrgMembershipAndThrow(identifier);
 
-        // Indicate if caller is trying to remove someone from an org they don't belong to
-        if (account.getOrgMembership() == null || !account.getOrgMembership().equals(identifier)) {
-            throw new BadRequestException("Account is not a member of organization " + identifier);
-        }
```
DwayneJengSage: Are we sure we want to remove this? Removing this will cause us to remove the participant from ANY organization, instead of the specified organization.

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -151,6 +158,7 @@ public void createStudy() {
         study.setModifiedOn(timestamp);
 
         when(studyDao.createStudy(any())).thenReturn(VERSION_HOLDER);
+        when(organizationDao.getOrganization(TEST_APP_ID, TEST_ORG_ID)).thenReturn(Optional.of(Organization.create()));
```
DwayneJengSage: This line doesn't appear to do anything. Did we forget a code change somewhere?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyDao.java
alxdarksage: Yes, works great. Changed.

src/main/resources/db/changelog/changelog.sql
alxdarksage: added

src/test/java/org/sagebionetworks/bridge/hibernate/SponsorPersistenceExceptionConverterTest.java
alxdarksage: Switched this to compare the classes

src/test/java/org/sagebionetworks/bridge/hibernate/SponsorPersistenceExceptionConverterTest.java
alxdarksage: Updated

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
alxdarksage: Yes this is different than the security check above, it's just checking that the id supplied matches the organization the account is in. Restoring this (the tests weren't deleted and they fail without this...!)

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
alxdarksage: Yes, I had toyed with forcing an orgId to be provided when creating a study, so it would start with one sponsor organization. But it seemed unnecessary since it's easy to associate/disassociate things after the fact if you need to. Removed.

src/main/java/org/sagebionetworks/bridge/hibernate/PersistenceExceptionConverter.java
DwayneJengSage: Does this need unit tests?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/PersistenceExceptionConverter.java
alxdarksage: Yes, added

alxdarksage: Yes, added

DwayneJengSage (approved):

PR #243, opened 2020-07-16
Allow developers access to a number of APIs if they are operating on their own accounts (so they can work without being made researchers).

alxdarksage:


src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -299,9 +296,15 @@ public String getParticipant(@PathVariable String userId, @RequestParam(defaultV
...
+        if (session.getId().equals(userId)) {
+            CriteriaContext context = getCriteriaContext(session);
+            participant = participantService.getSelfParticipant(app, context, consents);
+        } else {
+            participant = participantService.getParticipant(app, userId, consents);
+        }
+        
+        ObjectWriter writer = (app.isHealthCodeExportEnabled() || session.isInRole(ADMIN)) ?
                 StudyParticipant.API_WITH_HEALTH_CODE_WRITER :
                 StudyParticipant.API_NO_HEALTH_CODE_WRITER;
```
DwayneJengSage, line 309: There's a minor change here, if a participant calls getSelfParticipant in an app with healthCodeExportEnabled, they will be able to see their own HealthCode. Is this okay?

DwayneJengSage (approved):
Looks good. One question regarding getSelfParticipant()

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
alxdarksage: I think so, as they also have to be an administrative user, and so presumably looking at their account for development purposes.

DwayneJengSage: Can we add an integ test where a non-privileged user asks for self participant and we make sure we don't give out the healthCode?

alxdarksage: I will add this right now, I missed your note.

PR #252, opened 2020-08-11
Verify we work with real observation reports...

alxdarksage:
...taking care to detect invalid reports if they are submitted (because they might be, apparently, Columbia won't guarantee that they won't be).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -278,7 +278,7 @@ public class CRCController extends BaseController {
         String valueString = getObservationValue(observation);
         if (code == null || valueString == null || !SERUM_TEST_CODES.contains(code) || !SERUM_TEST_STATES.contains(valueString)) {
             state = TESTS_AVAILABLE_TYPE_UNKNOWN;
-            LOG.error("CRC observation in unknown format: code=" + code + ", valueString=" + valueString);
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -259,16 +269,56 @@ public StatusMessage updateParticipant(@PathVariable String userId) {
...
+        
+        AccountStates state = TESTS_AVAILABLE;
+        // There are two conditions under which the type of report is considered to be unknown. Either the code 
+        // is for a test other than the serum test, or the result value is not recognized by our client application,
+        // and that's also treated as an unknown type of report.
+        String code = getObservationCoding(observation);
+        String valueString = getObservationValue(observation);
+        if (code == null || valueString == null || !SERUM_TEST_CODES.contains(code) || !SERUM_TEST_STATES.contains(valueString)) {
+            state = TESTS_AVAILABLE_TYPE_UNKNOWN;
+            LOG.error("CRC observation in unknown format: code=" + code + ", valueString=" + valueString);
```
DwayneJengSage: Should this be logged as an error or as a warning? An error implies that there's something wrong with the server, and that it's actionable by the dev team. A warning implies that if there are a few of them that's fine, but if there are a lot of them, that's an indication of a widespread issue.

DwayneJengSage (approved):
Looks good. One comment about logging.

PR #253, opened 2020-08-12
Exporter 3 DynamoDB

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2603

Includes DynamoDB models and an API intended to be used by Integ Tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/dao/HealthDataEx3Dao.java
+++ b/src/main/java/org/sagebionetworks/bridge/dao/HealthDataEx3Dao.java
@@ -1,8 +1,8 @@
 package org.sagebionetworks.bridge.dao;
 
-import java.util.List;
 import java.util.Optional;
 
+import org.sagebionetworks.bridge.models.ForwardCursorPagedResourceList;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordEx3;
 
 /**
@@ -20,12 +20,14 @@ public interface HealthDataEx3Dao {
     Optional<HealthDataRecordEx3> getRecord(String id);
 
     /** Retrieves all records for the given healthcode and time range. */
-    List<HealthDataRecordEx3> getRecordsForHealthCode(String healthCode, long createdOnStart, long createdOnEnd);
+    ForwardCursorPagedResourceList<HealthDataRecordEx3> getRecordsForHealthCode(String healthCode, long createdOnStart,
+            long createdOnEnd, int pageSize, String offsetKey);
 
     /** Retrives all records for the given app and time range. */
-    List<HealthDataRecordEx3> getRecordsForApp(String appId, long createdOnStart, long createdOnEnd);
+    ForwardCursorPagedResourceList<HealthDataRecordEx3> getRecordsForApp(String appId, long createdOnStart,
+            long createdOnEnd, int pageSize, String offsetKey);
 
     /** Retrives all records for the given app, study, and time range. */
-    List<HealthDataRecordEx3> getRecordsForAppAndStudy(String appId, String studyId, long createdOnStart,
-            long createdOnEnd);
+    ForwardCursorPagedResourceList<HealthDataRecordEx3> getRecordsForAppAndStudy(String appId, String studyId,
+            long createdOnStart, long createdOnEnd, int pageSize, String offsetKey);
 }
--- a/src/main/java/org/sagebionetworks/bridge/dao/HealthDataEx3Dao.java
+++ b/src/main/java/org/sagebionetworks/bridge/dao/HealthDataEx3Dao.java
@@ -1,0 +1,31 @@
+package org.sagebionetworks.bridge.dao;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordEx3;
+
+/**
+ * Exporter 3 version of the Health Data DAO. Methods in this interface only apply to the EX3 implementation of Health
+ * Data Records, not to the original implementation of Health Data records.
+ */
+public interface HealthDataEx3Dao {
+    /** Create or update health data record. Returns the created or updated record. */
+    HealthDataRecordEx3 createOrUpdateRecord(HealthDataRecordEx3 record);
+
+    /** Deletes all health data records for the given health code. */
+    void deleteRecordsForHealthCode(String healthCode);
+
+    /** Retrieves the record for the given ID. */
+    Optional<HealthDataRecordEx3> getRecord(String id);
+
+    /** Retrieves all records for the given healthcode and time range. */
+    List<HealthDataRecordEx3> getRecordsForHealthCode(String healthCode, long createdOnStart, long createdOnEnd);
+
+    /** Retrives all records for the given app and time range. */
+    List<HealthDataRecordEx3> getRecordsForApp(String appId, long createdOnStart, long createdOnEnd);
+
+    /** Retrives all records for the given app, study, and time range. */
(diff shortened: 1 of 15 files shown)


src/main/java/org/sagebionetworks/bridge/dao/HealthDataEx3Dao.java
```diff
@@ -0,0 +1,31 @@
...
+    void deleteRecordsForHealthCode(String healthCode);
+
+    /** Retrieves the record for the given ID. */
+    Optional<HealthDataRecordEx3> getRecord(String id);
+
+    /** Retrieves all records for the given healthcode and time range. */
+    List<HealthDataRecordEx3> getRecordsForHealthCode(String healthCode, long createdOnStart, long createdOnEnd);
+
+    /** Retrives all records for the given app and time range. */
+    List<HealthDataRecordEx3> getRecordsForApp(String appId, long createdOnStart, long createdOnEnd);
```
alxdarksage: This doesn't have to be paged? (Same for the other ones, I'm assuming your limiting the records with the time range, but if we had a super-busy app, could this get overwhelmed?)

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3Dao.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public void deleteRecordsForHealthCode(String healthCode) {
+        // First, query the records we need to delete.
+        DynamoHealthDataRecordEx3 key = new DynamoHealthDataRecordEx3();
+        key.setHealthCode(healthCode);
+
+        DynamoDBQueryExpression<DynamoHealthDataRecordEx3> query = new DynamoDBQueryExpression<DynamoHealthDataRecordEx3>()
+                .withConsistentRead(false).withIndexName(DynamoHealthDataRecordEx3.HEALTHCODE_CREATEDON_INDEX)
+                .withHashKeyValues(key);
+        List<DynamoHealthDataRecordEx3> recordsToDelete = queryHelper(query);
```
alxdarksage, line 54: Doesn't the query "time out" at 1MB of data returned...what happens if the user has more data than this?

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3Dao.java
```diff
@@ -0,0 +1,133 @@
...
+    public List<HealthDataRecordEx3> getRecordsForHealthCode(String healthCode, long createdOnStart,
+            long createdOnEnd) {
+        // Hash key.
+        DynamoHealthDataRecordEx3 key = new DynamoHealthDataRecordEx3();
+        key.setHealthCode(healthCode);
+
+        // Range key.
+        Condition rangeKeyCondition = new Condition().withComparisonOperator(ComparisonOperator.BETWEEN)
+                .withAttributeValueList(new AttributeValue().withN(String.valueOf(createdOnStart)),
+                        new AttributeValue().withN(String.valueOf(createdOnEnd)));
```
alxdarksage: Is time range limited in service?

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
```diff
@@ -0,0 +1,151 @@
...
+
+        return record;
+    }
+
+    /** Retrieves all records for the given user and time range. */
+    @GetMapping(path="/v1/admin/exporter3/participants/{userId}/healthdata")
+    public ResourceList<HealthDataRecordEx3> getRecordsForUser(@PathVariable String userId,
+            @RequestParam(required = false) String createdOnStart,
+            @RequestParam(required = false) String createdOnEnd) {
+        UserSession session = getAuthenticatedSession(SUPERADMIN);
```
alxdarksage, line 95: This won't be accessible to developers, or worker accounts?

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
```diff
@@ -0,0 +1,151 @@
...
+                .withRequestParam(ResourceList.START_TIME, createdOnStart)
+                .withRequestParam(ResourceList.END_TIME, createdOnEnd);
+    }
+
+    /** Retrieves all records in the current app for the given time range. */
+    @GetMapping(path="/v1/admin/exporter3/apps/self/healthdata")
+    public ResourceList<HealthDataRecordEx3> getRecordsForCurrentApp(
+            @RequestParam(required = false) String createdOnStart,
+            @RequestParam(required = false) String createdOnEnd) {
+        UserSession session = getAuthenticatedSession(SUPERADMIN);
```
alxdarksage, line 119: Not worker accounts? The endpoint URL also encodes who can access the endpoint in /admin/ rather than just naming the resource, so you would have a harder time changing this later. Are there any other kinds of records than health data records? Is this a set of  /v1/healthdata/ endpoints?

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
```diff
@@ -0,0 +1,151 @@
...
+                createdOnStartDateTime, createdOnEndDateTime);
+
+        return new DateTimeRangeResourceList<>(recordList)
+                .withRequestParam("appId", session.getAppId())
+                .withRequestParam(ResourceList.START_TIME, createdOnStart)
+                .withRequestParam(ResourceList.END_TIME, createdOnEnd);
+    }
+
+    /** Retrieves all records in the current app for the given study and time range. */
+    @GetMapping(path="/v1/admin/exporter3/apps/self/study/{studyId}/healthdata")
```
alxdarksage: .../self/studies/{studyId}/... ?

alxdarksage (changes requested):
I'm curious about the 1MB limit on returned data from DDB, and some comments about the URLs and permissions of the controller.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3Dao.java
DwayneJengSage: The 1MB limitation is on queryPage(), which gets one page of the data. The max page size is 1MB. query() automatically handles pagination for us.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3Dao.java
DwayneJengSage: Yes, the service validates the time ranges.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
DwayneJengSage: That is correct. The purpose of these APIs is specifically to facilitate integration tests. We have no plans to expose these particular APIs outside of this scope.

When in the future, when we build participant and researcher-facing APIs for the Exporter 3 stuff, we'll be creating a new set of APIs, with different names and URLs.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
DwayneJengSage: These APIs are only exposed for integration tests, so it should be pretty easy to change these or remove these later on.

alxdarksage (approved):

PR #257, opened 2020-08-13
Give "admin" specified in config SUPERADMIN permission

kss39:
Giving SUPERADMIN instead of ADMIN permission for the StudyParticipant specified by admin.email in config file, so that we don't need to promote admin manually for the integration tests.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/DefaultStudyBootstrapperTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/DefaultStudyBootstrapperTest.java
@@ -123,7 +123,7 @@ public class DefaultStudyBootstrapperTest extends Mockito {
         assertEquals(SHARED_SUBPOP, subpopCaptor.getAllValues().get(2));
         
         StudyParticipant admin = participantCaptor.getAllValues().get(0);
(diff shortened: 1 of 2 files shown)


DwayneJengSage (approved):

src/test/java/org/sagebionetworks/bridge/DefaultStudyBootstrapperTest.java
```diff
@@ -123,7 +123,7 @@ public void createsDefaultStudyWhenMissing() {
         assertEquals(SHARED_SUBPOP, subpopCaptor.getAllValues().get(2));
         
         StudyParticipant admin = participantCaptor.getAllValues().get(0);
-        assertEquals(admin.getRoles(), ImmutableSet.of(ADMIN, RESEARCHER));
+        assertEquals(admin.getRoles(), ImmutableSet.of(SUPERADMIN, RESEARCHER));
```
alxdarksage: Minor: SUPERADMIN can do everything that RESEARCHER can do, so you only need to add SUPERADMIN.

alxdarksage (changes requested):

PR #258, opened 2020-08-13
For new APIs, remove deprecated ResourceList properties

alxdarksage:
I'm pulling this out of the sponsors pull request to reduce it's size and ensure everything is getting tested. Looking at the APIs I realized there's a lot of slop in the deprecated properties that I would like to remove in new APIs. So there's a flag which can be set by new APIs to remove deprecated properties from the JSON. Doesn't break existing integration tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/DateRangeResourceList.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/DateRangeResourceList.java
@@ -9,22 +9,29 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class DateRangeResourceList<T> extends ResourceList<T> {
     
+    public DateRangeResourceList(List<T> items, boolean suppressDeprecated) {
+        super(items, suppressDeprecated);
+    }
+    
     @JsonCreator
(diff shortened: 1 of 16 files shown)


src/main/java/org/sagebionetworks/bridge/models/DateRangeResourceList.java
```diff
@@ -9,22 +9,29 @@
 
 public class DateRangeResourceList<T> extends ResourceList<T> {
     
+    public DateRangeResourceList(List<T> items, boolean suppressDeprecated) {
+        super(items, suppressDeprecated);
+    }
```
DwayneJengSage, line 14: Did we forget to include a unit test for this class?

DwayneJengSage (approved):
One question about missing unit tests. Otherwise, looks good.

src/main/java/org/sagebionetworks/bridge/models/DateRangeResourceList.java
alxdarksage: Yes, added it.

PR #259, opened 2020-08-14
In anticipating of APIs that will page studies, convert existing API to use paging

alxdarksage:
The sponsors APIs may have cases where one organization (Sage) could have hundreds of studies, so switching these to be paged everywhere so we don't have multiple StudyList types in the SDK. Since these were recently substudies, I'm confident no one is using these APIs right now other than the BSM.

Also renaming "id" to "identifier" after examining our models and deciding that we slightly favor "identifier" (I feel this should be consistent in our APIs).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
@@ -1,5 +1,6 @@
 package org.sagebionetworks.bridge.hibernate;
 
+import javax.persistence.Column;
 import javax.persistence.Convert;
 import javax.persistence.Entity;
 import javax.persistence.Id;
@@ -21,6 +22,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 @BridgeTypeName("Study")
 public class HibernateStudy implements Study {
     @Id
+    @Column(name = "id")
     private String identifier;
     @Id
     private String appId;
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
@@ -12,6 +12,7 @@ import org.sagebionetworks.bridge.json.BridgeTypeName;
 import org.sagebionetworks.bridge.models.studies.Study;
 import org.sagebionetworks.bridge.models.studies.StudyId;
 
+import com.fasterxml.jackson.annotation.JsonAlias;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 @Entity
@@ -20,7 +21,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 @BridgeTypeName("Study")
 public class HibernateStudy implements Study {
     @Id
-    private String id;
+    private String identifier;
     @Id
     private String appId;
     private String name;
@@ -33,13 +34,14 @@ public class HibernateStudy implements Study {
     private Long version;
     
     @Override
-    public String getId() {
-        return id;
+    @JsonAlias("id")
+    public String getIdentifier() {
+        return identifier;
     }
 
     @Override
-    public void setId(String id) {
-        this.id = id;
+    public void setIdentifier(String identifier) {
+        this.identifier = identifier;
     }
 
     @JsonIgnore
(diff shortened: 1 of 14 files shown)


DwayneJengSage:
If we’re standardizing on “id” vs “identifier”, I would prefer “id”.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
```diff
@@ -20,7 +21,7 @@
 @BridgeTypeName("Study")
 public class HibernateStudy implements Study {
     @Id
-    private String id;
+    private String identifier;
```
DwayneJengSage: If the attribute is still called "id" in the database, how does Hibernate know to look for "id" instead of "identifier"?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyDao.java
```diff
@@ -27,16 +28,21 @@ final void setHibernateHelper(HibernateHelper hibernateHelper) {
...
         if (!includeDeleted) {
             query += " and deleted != 1";
         }
-        return ImmutableList.copyOf(hibernateHelper.queryGet(query, parameters, 
-                null, null, HibernateStudy.class));
+        int total = hibernateHelper.queryCount("select count(*) " + query, parameters);
+
+        List<HibernateStudy> hibStudies = hibernateHelper.queryGet(query, parameters, 
+                offsetBy, pageSize, HibernateStudy.class);
+        List<Study> studies = hibStudies.stream().map(s -> (Study)s).collect(toList());
```
DwayneJengSage: I prefer ImmutableList.copyOf(). I think both are probably roughly equally efficient, but copyOf() looks cleaner.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -46,14 +56,32 @@ public Study getStudy(String appId, String studyId, boolean throwsException) {
...
-                .map(Study::getId).collect(BridgeCollectors.toImmutableSet());
+        int offset = 0;
+        Set<String> identifiers = new HashSet<>();
+        PagedResourceList<Study> results = null;
+        
+        do {
+            results = getStudies(appId, offset, API_MAXIMUM_PAGE_SIZE, false);
+            identifiers.addAll(results.getItems().stream().map(Study::getIdentifier).collect(toSet()));
+            offset += API_MAXIMUM_PAGE_SIZE;
+        } while(identifiers.size() < results.getTotal());
```
DwayneJengSage: There's not really any reason to use pagination here. We're requesting the same amount of data from the database no matter what, except with the extra overhead of making a bunch of paginated calls, and the extra complexity of adding server logic for something the database already handles natively.

alxdarksage:
Okay I can undo it here. We're almost split on this right now:

getIdentifier()
  Activity
  App
  Assessment
  ExternalIdentifier (& ExternalIdentifierInfo)
  Organization
  ReportDataKey/ReportDataIndex
  Survey
  SurveyElement
  SurveyReference
  TaskReference

getId()
  Account
  AccountSummary
  AppConfigElement
  AssessmentReference
  ConfigReference
  SchemaReference
  HealthDataRecord
  SharedModuleMetadata
  Study
  StudyParticipant
  UploadSession
  UserSession

However assessment and study (nee substudy) are still changeable and can be converted to use id. Then the majority of classes would use ID including many of the newer objects.

Erin-Mounts:
FWIW “id” is an Objective-C keyword and makes JSON serialization/deserialization a bit trickier on iOS, which iirc had something to do with the historical preference for “identifier”.

I don’t think this will be an issue forever, as we’re working to move that serialization/deserialization code into Kotlin so it can be maintained in one place for multiple platforms, but for now “identifier” would be safer for client-facing objects.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
alxdarksage: I don't know how, but I can verify it does (maybe because table DDL defines the primary key and Hibernate can figure it out). For clarity, I'll add a @Column annotation to HibernateStudy.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyDao.java
alxdarksage: I didn't know that would cast to the interface, very nice, changed.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Removed (no paging indicated to SQL)

DwayneJengSage (approved):

PR #263, opened 2020-08-20
Change error reporting when location lookup fails, and update to the new (as of this morning) location API provided by external partner

alxdarksage:
I've tested this manually using the integration tests and providing valid and invalid location identifiers. I won't add valid location identifiers to the integration tests though, as this would call out to Columbia and might be fragile for our tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
@@ -329,19 +329,21 @@ public class CRCController extends BaseController {
             
             int statusCode = response.getStatusLine().getStatusCode();
             if (statusCode != 200 && statusCode != 201) {
-                LOG.warn("Error retrieving location, id = " + locationId + ", status = " + statusCode
-                        + ", response body = " + resBody);
+                logWarningMessage(locationId, statusCode, resBody);
                 return;
             }
             JsonNode bundleJson = BridgeObjectMapper.get().readTree(resBody);
             if (!bundleJson.has("entry") || ((ArrayNode)bundleJson.get("entry")).size() == 0) {
-                LOG.warn("Error retrieving location, id = " + locationId + ", status = " + statusCode
-                        + ", response body = " + resBody);
+                logWarningMessage(locationId, statusCode, resBody);
                 return;
             }
-            JsonNode locationJson = bundleJson.get("entry").get(0).get("resource");
-            JsonNode telecom = locationJson.get("telecom");
-            JsonNode address = locationJson.get("address");
+            JsonNode resJson = bundleJson.get("entry").get(0).get("resource");
+            if (resJson == null) {
+                logWarningMessage(locationId, statusCode, resBody);
+                return;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -315,41 +311,35 @@ String getObservationValue(Observation observation) {
...
+            }
+            JsonNode bundleJson = BridgeObjectMapper.get().readTree(resBody);
+            if (!bundleJson.has("entry") || ((ArrayNode)bundleJson.get("entry")).size() == 0) {
+                LOG.warn("Error retrieving location, id = " + locationId + ", status = " + statusCode
+                        + ", response body = " + resBody);
+                return;
+            }
+            JsonNode locationJson = bundleJson.get("entry").get(0).get("resource");
             JsonNode telecom = locationJson.get("telecom");
             JsonNode address = locationJson.get("address");
```
DwayneJengSage: Should we also check that "resource" exists? Otherwise, we'd get a NullPointerException if they return JSON that doesn't contain "resource".

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -637,27 +659,42 @@ public void postAppointmentFailsWhenLocationCallFails() throws Exception {
         when(mockAccountService.authenticate(any(), any())).thenReturn(account);
         when(mockAccountService.getAccount(ACCOUNT_ID)).thenReturn(account);
         
-        HttpResponse mockResponse = mock(HttpResponse.class);
-        doReturn(mockResponse).when(controller).get(any(), any());
+//        mockGetGeocoding();
```
DwayneJengSage, line 665: Commented out test code?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: It appears to be all or nothing. If there's an entry, its got a resource.

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
alxdarksage: Geocoding is temporarily disabled because it has been reprioritized for launch, but we may restore it so I'm just commenting it out for the moment.

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: I can add NPE checks though, no harm.

alxdarksage: Updated

DwayneJengSage (approved):

PR #265, opened 2020-08-21
Participant File Service API

kss39:
Still finishing the integration tests, to make sure that this actually works.


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Resource(name = "participantFileDdbMapper")
+    public final void setMapper(DynamoDBMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    @Override
+    public ForwardCursorPagedResourceList<ParticipantFile> getParticipantFiles(String userId, String offsetKey, int pageSize) {
+        checkNotNull(userId);
+        checkArgument(isNotBlank(userId));
```
alxdarksage, line 36: This check covers the checkNotNull check

src/main/java/org/sagebionetworks/bridge/dao/ParticipantFileDao.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Get a ForwardCursorPagedResourceList of ParticipantFiles from the given userId, with nextPageOffsetKey set.
+     * If nextPageOffsetKey is null, then the list reached the end and there does not exist next page.
+     *
+     * @param userId the id of the StudyParticipant
+     * @param offsetKey the exclusive starting offset of the query, if null, then query from the start
+     * @param pageSize the number of items in the result page
+     * @return a ForwardCursorPagedResourceList of ParticipantFiles
+     * @throws org.sagebionetworks.bridge.exceptions.BadRequestException if pageSize is less than 1 or greater
+     *         than API_MAXIMUM_PAGE_SIZE
```
alxdarksage: The minimum you specify is actually 5, though there's no reason the constant couldn't be set to 1.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+
+        if (pageSize < API_MINIMUM_PAGE_SIZE || pageSize > API_MAXIMUM_PAGE_SIZE) {
+            throw new BadRequestException(PAGE_SIZE_ERROR);
+        }
+        ParticipantFile key = new DynamoParticipantFile(userId, null);
```
alxdarksage: While this appears to be scoped, it could be possible while authenticated in app A, to request the files for a user ID in app B. You'll need to check the appId the caller's session is bound to, and we've done this a few different ways. One is to create a compound key of appId + userId, so that it is impossible to retrieve files outside of your bound app context.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+        checkArgument(isNotBlank(file.getFileId()));
+        mapper.save(file);
+    }
+
+    @Override
+    public void deleteParticipantFile(String userId, String fileId) {
+        checkNotNull(fileId);
+        checkNotNull(userId);
+        checkArgument(isNotBlank(fileId));
+        checkArgument(isNotBlank(userId));
```
alxdarksage, line 86: checkArgument(isNotBlank(...)) also checks for null (as noted above)

src/main/java/org/sagebionetworks/bridge/models/files/ParticipantFile.java
```diff
@@ -0,0 +1,22 @@
+package org.sagebionetworks.bridge.models.files;
+
+import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+public interface ParticipantFile extends BridgeEntity {
+    String getFileId();
```
alxdarksage, line 17: We conventionally add a static create() method that will return an implementation class (DynamoParticipantFile).

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+     *                  (the exclusive starting offset of the query, if null, then query from the start)
+     * @param pageSize the number of items in the result page
+     * @return a ForwardCursorPagedResourceList of ParticipantFiles
+     */
+    public ForwardCursorPagedResourceList<ParticipantFile> getParticipantFiles(String userId, String offsetKey, int pageSize) {
+        checkNotNull(userId);
+        checkArgument(isNotBlank(userId));
+
+        // pageSize validation is checked in DAO
+        return participantFileDao.getParticipantFiles(userId, offsetKey, pageSize);
```
alxdarksage, line 83: I would move any and all business logic to the service. The service is reusable if we switch persistence mechanism, let's say from Dynamo to a relational database. If we do that, we need to completely rewrite the DAO...so we wouldn't want paging checks in the DAO as that would have to be copied over. In the service it remains reusable and unchanged. Although Java package visibility rules don't allow us to enforce it with the compiler, your DAO should only be called by the service, which "protects" it from these kind of user submitted errors.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @DeleteMapping("/v3/participants/self/files/{fileId}")
+    public StatusMessage deleteParticipantFile(@PathVariable String fileId) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+
+        try {
+            fileService.deleteParticipantFile(userId, fileId);
+        } catch (EntityNotFoundException e) {
+            return DELETE_FAIL_MSG;
```
alxdarksage: Just allow this to throw through the method, the ExceptionHandler will return the correct 404.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,83 @@
...
+    @GetMapping("v3/participants/self/files/{fileId}")
+    @ResponseStatus(HttpStatus.FOUND)
+    public ParticipantFile getParticipantFile(@PathVariable String fileId) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+
+        return fileService.getParticipantFile(userId, fileId);
+    }
+
+    @PostMapping("v3/participants/self/files/{fileId}")
```
alxdarksage, line 58: Should return 201, not 200

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,83 @@
...
+            @RequestParam(required = false) String offsetKey, @RequestParam(required = false) String pageSize) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+        int pageSizeInt = BridgeUtils.getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+        return fileService.getParticipantFiles(userId, offsetKey, pageSizeInt);
+    }
+
+    @GetMapping("v3/participants/self/files/{fileId}")
+    @ResponseStatus(HttpStatus.FOUND)
+    public ParticipantFile getParticipantFile(@PathVariable String fileId) {
```
alxdarksage, line 49: A 302 response is returned with a Location header, but we're not sending a redirect to the client, we're returning the entity (which has a download URL in it). Therefore, the response here should be 200.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,83 @@
...
+        return fileService.getParticipantFile(userId, fileId);
+    }
+
+    @PostMapping("v3/participants/self/files/{fileId}")
+    public ParticipantFile createParticipantFile(@PathVariable String fileId) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+
+        ParticipantFile file = parseJson(DynamoParticipantFile.class);
+        file.setUserId(userId);
```
alxdarksage: set the appId from the user's session. It should not be possible to create the file in another app.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+
+        DynamoParticipantFile deleteTarget = new DynamoParticipantFile(fileId, userId);
+        if (mapper.load(deleteTarget) != null) {
```
alxdarksage, line 89: When this load ensures the file is in the caller's appId, this call will not be usable to delete files in a another app (though the userid is set from the session by the controller).

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
```diff
@@ -273,6 +274,13 @@ public AmazonS3 fileUploadS3Client(BasicAWSCredentials awsCredentials) {
         return AmazonS3ClientBuilder.standard().withPathStyleAccessEnabled(true).withRegion(US_EAST_1)
                 .withCredentials(new AWSStaticCredentialsProvider(awsCredentials)).build();
     }
+
+    @Bean(name = "participantFileS3Client")
+    @Resource(name = "awsCredentials")
+    public AmazonS3 participantFileS3Client(BasicAWSCredentials awsCredentials) {
+        return AmazonS3ClientBuilder.standard().withRegion(US_EAST_1)
+                .withCredentials(new AWSStaticCredentialsProvider(awsCredentials)).build();
+    }
```
DwayneJengSage: We don't need to create a new S3 client. This is a holdover from back in the old days when we had several different AWS Roles and several different S3 clients, but that's not sustainable. Instead, you should just use s3Client.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantFileDao.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return the ParticipantFile
+     */
+    Optional<ParticipantFile> getParticipantFile(String userId, String fileId);
+
+    /**
+     * Upload the ParticipantFile to the database.
+     *
+     * @param file the file to be uploaded
+     */
+    // TODO: Not decided what will happen if such file already exists.
```
DwayneJengSage: If the file already exists, we should throw an EntityAlreadyExistsException.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public DynamoParticipantFile(String userId, String fileId) {
+        this.userId = userId;
+        this.fileId = fileId;
+    }
+
+    public void setFileId(String fileId) {
+        this.fileId = fileId;
+    }
```
DwayneJengSage, line 46: Coding convention is that we put the getters and the setters for the same attribute next to each other.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
```diff
@@ -0,0 +1,122 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.models.files.ParticipantFile;
+
+@DynamoDBTable(tableName = "ParticipantFiles")
+public class DynamoParticipantFile implements ParticipantFile {
+    /**
+     * the file Id for this ParticipantFile.
+     * Unique in the scope of the user (StudyParticipant).
+     */
```
DwayneJengSage: Coding convention is that we put javadoc comments on the getters, not on the member variables. This is because javadoc (and therefore the IDE) looks at docs for public members, so the comments on the private member variables will never be seen.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * The App ID of this file.
+     */
+    private String appId;
+
+    @JsonIgnore
+    private String uploadUrl;
+    @JsonIgnore
+    private String downloadUrl;
```
DwayneJengSage: The URLs will be constantly changing, so there's not really a point to store them in DDB. Instead of URLs, we should store S3 keys.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
```diff
@@ -0,0 +1,122 @@
...
+    @DynamoDBAttribute(attributeName = "mimeType")
+    public String getMimeType() {
+        return this.mimeType;
+    }
+
+    @Override
+    @DynamoDBAttribute(attributeName = "createdOn")
+    public DateTime getCreatedOn() {
+        return this.createdOn;
+    }
```
DwayneJengSage: Dynamo can't store DateTimes. You'll need to use Long or String and use the `@JsonSerialize(using = DateTimeToLongSerializer.class)` on the getter and `@JsonDeserialize(using = DateTimeToLongDeserializer.class)` on the setter. (Or DateTimeSerializer and DateTimeDeserializer if using String.)

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
+public class DynamoParticipantFileTest {
+    private static final BridgeObjectMapper MAPPER = BridgeObjectMapper.get();
+
+    @Test
+    public void canSerialize() throws Exception {
```
DwayneJengSage, line 15: Should also test deserializing the JSON back into a Java object.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public ForwardCursorPagedResourceList<ParticipantFile> getParticipantFiles(String userId, String offsetKey, int pageSize) {
+        checkNotNull(userId);
+        checkArgument(isNotBlank(userId));
+
+        if (pageSize < API_MINIMUM_PAGE_SIZE || pageSize > API_MAXIMUM_PAGE_SIZE) {
+            throw new BadRequestException(PAGE_SIZE_ERROR);
+        }
```
DwayneJengSage: We normally put input validation in the service, not in the DAO. This way, if they DAO needs to be migrated, there's less code to migrate.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+    DynamoDBMapper mapper;
+
+    @Mock
+    PaginatedQueryList<ParticipantFile> paginatedQueryList;
+
+    @Captor
+    ArgumentCaptor<ParticipantFile> fileCaptor;
+
+    @InjectMocks
+    @Spy
```
DwayneJengSage: You don't need to spy if you aren't spying anything.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+    DynamoParticipantFileDao dao;
+
+    @BeforeMethod
+    public void beforeMethod() {
+        MockitoAnnotations.initMocks(this);
+        RESULT.setCreatedOn(TestConstants.TIMESTAMP);
+        RESULT.setAppId("api");
+        RESULT.setMimeType("dummy-type");
+        RESULT.setDownloadUrl("fake_url");
```
DwayneJengSage: Instead of re-instatiating the RESULT in the beforeMethod (which runs before every test), I'd recommend using a static block to statically initiate this. (Or make it a non-final instance and use a static `@BeforeClass`.)

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
DwayneJengSage: Is this even possible? Users can only get their own files, so it would be impossible for them to get a file outside of their app.

The only way a user can get a file outside their app is if two users get the same user ID, which would break a lot more than just Participant File API.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public void getParticipantFiles() {
```
DwayneJengSage, line 85: We need a version of this test with offsetKey.

We also need a version of this test that returns the nextOffsetKey.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+        when(paginatedQueryList.size()).thenReturn(1);
+        when(paginatedQueryList.get(0)).thenReturn(RESULT);
+        when(mapper.query(eq(ParticipantFile.class), any())).thenReturn(paginatedQueryList);
+
+        ForwardCursorPagedResourceList<ParticipantFile> result = dao.getParticipantFiles(KEY.getUserId(), null, 5);
+        assertNotNull(result);
```
DwayneJengSage, line 90: Check that the result contains request params for offsetKey and pageSize.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+        dao.getParticipantFiles("test_user", "dummy-key", 1);
+    }
+
+    @Test(expectedExceptions = BadRequestException.class)
+    public void getParticipantFilesPageSizeTooLarge() {
+        dao.getParticipantFiles("test_user", "dummy-key", 5000);
+    }
+
+    @Test
+    public void getParticipantFile() {
```
DwayneJengSage, line 191: Need a version of this test that returns an empty result.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,122 @@
...
+    public void uploadParticipantFile() {
+        dao.uploadParticipantFile(KEY);
+        verify(mapper).save(KEY);
+    }
+
+    @Test
+    public void deleteParticipantFile() {
+        when(mapper.load(any())).thenReturn(RESULT);
+        dao.deleteParticipantFile(KEY.getUserId(), KEY.getFileId());
+        verify(mapper).delete(any());
```
DwayneJengSage: We can use an ArgumentCaptor to verify we pass in the keys correctly into the delete().

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.joda.time.DateTimeZone.UTC;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.sagebionetworks.bridge.config.Environment.LOCAL;
+import static org.sagebionetworks.bridge.validators.ParticipantFileValidator.INSTANCE;
+
+@Component
+public class ParticipantFileService {
+
+    static final int EXPIRATION_IN_MINUTES = 10;
```
DwayneJengSage: We should set the expiration to 24 hours, since short expiration times have historically had problems with apps.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+    private String bucketName;
+
+    @Autowired
+    final void setParticipantFileDao(ParticipantFileDao dao) {
+        this.participantFileDao = dao;
+    }
+
+    @Autowired
+    final void setConfig(BridgeConfig config) {
+        bucketName = config.getHostnameWithPostfix("participant-file");
```
DwayneJengSage: This doesn't do what you want it to do. Use config.getString() instead.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+     * Returns the download URL for the given ParticipantFile.
+     * This method does not check whether this file actually exists.
+     *
+     * @param file the file whose URL is returned
+     * @return the download URL for the given ParticipantFile.
+     */
+    protected String getDownloadUrl(ParticipantFile file) {
+        String protocol = (env == LOCAL) ? "http" : "https";
+        return protocol + "://" + bucketName + "/" + getFilePath(file);
+    }
```
DwayneJengSage: This doesn't work, because we don't want to open the S3 bucket publicly. We'll need to use an S3 presigned URL instead.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Returns the url path of the given file.
+     * @param file the file
+     * @return the url path of the given file
+     */
+    protected String getFilePath(ParticipantFile file) {
```
DwayneJengSage: Does this need to be protexted instead of private?

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+
+    @Autowired
+    final void setParticipantFileDao(ParticipantFileDao dao) {
+        this.participantFileDao = dao;
+    }
+
+    @Autowired
+    final void setConfig(BridgeConfig config) {
+        bucketName = config.getHostnameWithPostfix("participant-file");
+        env = config.getEnvironment();
```
DwayneJengSage: What is env used for?

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+import static org.mockito.Mockito.when;
+import static org.sagebionetworks.bridge.config.Environment.PROD;
+import static org.testng.Assert.assertEquals;
+
+public class ParticipantFileServiceTest {
+
+    private static final String UPLOAD_BUCKET = "participant-file.sagebridge.org";
+    // private static final String UPLOAD_BUCKET_STAGING = "participant-file-staging.sagebridge.org";
+    private static final String DOWNLOAD_URL_1 = "https://participant-file.sagebridge.org/test_user/file_id";
+    // private static final String DOWNLOAD_URL_2 = "https://participant-file.sagebridge.org/oneGuid.1422311912486";
```
DwayneJengSage: I know this is just a test string, but the real bucket won't look like a URL.

Also, delete commented out strings.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+    ParticipantFileDao mockFileDao;
+
+    @Mock
+    BridgeConfig mockConfig;
+
+    @Mock
+    AmazonS3 mockS3Client;
+
+    @InjectMocks
+    @Spy
```
DwayneJengSage: You don't need to spy if you aren't spying anything.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+        verify(mockFileDao).getParticipantFiles("test_user", null, 20);
+    }
+
+    @Test
+    public void getParticipantFile() {
+        ParticipantFile file = new DynamoParticipantFile("test_user", "file_id");
+        file.setAppId("api");
+        file.setMimeType("dummy-type");
+        file.setCreatedOn(TestConstants.TIMESTAMP);
+        file.setDownloadUrl(DOWNLOAD_URL_1);
```
DwayneJengSage: Don't set the download URL in the file you return from the mock dao. This way, we can test that the service sets the download URL.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Test(expectedExceptions = EntityNotFoundException.class)
+    public void getParticipantFileNoSuchFile() {
+        when(mockFileDao.getParticipantFile(any(), any())).thenReturn(Optional.empty());
+
+        service.getParticipantFile("test_user", "file_id");
+    }
+
+    @Test
+    public void createParticipantFile() throws Exception {
```
DwayneJengSage: Add unit test for invalid file.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Logs the metadata in the database, and then returns the passed-in ParticipantFile with pre-signed URL
+     * for S3 file upload. If the file or the file metadata already exists, throws EntityAlreadyExistsException.
+     *
+     * @param file the file metadata to be upload
+     * @return the ParticipantFile with pre-signed S3 URL for file upload.
+     * @throws org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException
+     *         if the file already exists.
+     */
+    public ParticipantFile createParticipantFile(ParticipantFile file) {
```
DwayneJengSage: Need to take in the user's user ID and set the file with that user's userID. Otherwise, users can create files for other users.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
DwayneJengSage: As per the client team and as per the design doc, we _want_ to send a 302 with a redirect. If this controller doesn't do that, we should fix it so it does.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,83 @@
...
+        file.setCreatedOn(new DateTime().withZone(UTC));
```
DwayneJengSage: It is neither necessary nor desirable to flatten all timestamps to UTC.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+        Date expiration = new DateTime().withZone(UTC).plusMinutes(EXPIRATION_IN_MINUTES).toDate();
+        GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(bucketName, getFilePath(file), PUT);
+        request.setExpiration(expiration);
+        request.setContentType(file.getMimeType());
+        ResponseHeaderOverrides headers = new ResponseHeaderOverrides()
+                .withContentDisposition("attachment; filename=\""+getFilePath(file)+"\"");
+        request.setResponseHeaders(headers);
+
+        URL uploadUrl = s3Client.generatePresignedUrl(request);
+        file.setUploadUrl(uploadUrl.toExternalForm());
```
DwayneJengSage: This won't do anything, because uploadURL is set with JsonIgnore.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+        URL url = new URL("https://" + UPLOAD_BUCKET);
+        when(mockS3Client.generatePresignedUrl(any())).thenReturn(url);
```
DwayneJengSage: We should verify the args to generatePresignedUrl with an ArgumentCaptor.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+                    throw new EntityAlreadyExistsException(ParticipantFile.class, entityKey);
+                }
+        );
+
+        Date expiration = new DateTime().withZone(UTC).plusMinutes(EXPIRATION_IN_MINUTES).toDate();
+        GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(bucketName, getFilePath(file), PUT);
+        request.setExpiration(expiration);
+        request.setContentType(file.getMimeType());
+        ResponseHeaderOverrides headers = new ResponseHeaderOverrides()
+                .withContentDisposition("attachment; filename=\""+getFilePath(file)+"\"");
```
DwayneJengSage: This isn't going to be used in a browser, so we don't need the Content-Disposition header.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+
+        participantFileDao.getParticipantFile(file.getUserId(), file.getFileId()).ifPresent(
+                oldFile -> {
+                    Map<String, Object> entityKey = new HashMap<>();
+                    entityKey.put("oldFile", oldFile);
+                    throw new EntityAlreadyExistsException(ParticipantFile.class, entityKey);
+                }
+        );
+
+        Date expiration = new DateTime().withZone(UTC).plusMinutes(EXPIRATION_IN_MINUTES).toDate();
```
DwayneJengSage: It is neither necessary nor desirable to flatten timestamps to UTC.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,172 @@
...
+        GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(bucketName, getFilePath(file), PUT);
```
DwayneJengSage: We should specify server-side encryption by specifying `request.addRequestParameter(Headers.SERVER_SIDE_ENCRYPTION, ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);`

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,154 @@
...
+        ParticipantFile file = new DynamoParticipantFile("test_user", "file_id");
+        file.setAppId("api");
+        file.setMimeType("dummy-type");
+        file.setCreatedOn(TestConstants.TIMESTAMP);
+        file.setDownloadUrl(DOWNLOAD_URL_1);
+
+        when(mockFileDao.getParticipantFile(eq("test_user"), eq("file_id"))).thenReturn(Optional.of(file));
+        service.deleteParticipantFile("test_user", "file_id");
+
+        verify(mockFileDao).deleteParticipantFile(eq("test_user"), eq("file_id"));
```
DwayneJengSage, line 212: Also verify we delete the file from S3.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        assertSame(result, persisted);
+        verify(mockFileService).getParticipantFile("test_user", "file_id");
+    }
+
+    @Test
+    public void createParticipantFile() throws Exception {
+        ParticipantFile newFile = new DynamoParticipantFile("test_user", "file_id");
+        newFile.setAppId("api");
+        newFile.setMimeType("dummy-type");
+        mockRequestBody(mockRequest, persisted);
```
DwayneJengSage: You should mock newFile, not persisted.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        when(mockFileService.createParticipantFile(any())).thenReturn(persisted);
+        ParticipantFile result = controller.createParticipantFile("file_id");
+
+        assertEquals(result.getFileId(), "file_id");
+        assertEquals(result.getUserId(), "test_user");
+        assertEquals(result.getAppId(), "api");
+        assertEquals(result.getMimeType(), "dummy-type");
```
DwayneJengSage: Need to check createdOn.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        verify(mockFileService).getParticipantFile("test_user", "file_id");
+    }
+
+    @Test
+    public void createParticipantFile() throws Exception {
+        ParticipantFile newFile = new DynamoParticipantFile("test_user", "file_id");
+        newFile.setAppId("api");
+        newFile.setMimeType("dummy-type");
+        mockRequestBody(mockRequest, persisted);
+        when(mockFileService.createParticipantFile(any())).thenReturn(persisted);
```
DwayneJengSage: Should return newFile, not persisted.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,134 @@
...
+        assertEquals(result.getFileId(), "file_id");
+        assertEquals(result.getUserId(), "test_user");
+        assertEquals(result.getAppId(), "api");
+        assertEquals(result.getMimeType(), "dummy-type");
+    }
+
+    @Test
+    public void deleteParticipantFile() {
+        StatusMessage message = controller.deleteParticipantFile("file_id");
+        assertEquals(message.getMessage(), DELETE_MSG.getMessage());
```
DwayneJengSage, line 139: Need to verify call to service.deleteParticipantFile()

src/main/java/org/sagebionetworks/bridge/validators/ParticipantFileValidator.java
```diff
@@ -0,0 +1,34 @@
...
+
+        if (StringUtils.isBlank(file.getUserId())) {
+            errors.rejectValue("userId", "is required");
+        }
+        if (StringUtils.isBlank(file.getFileId())) {
+            errors.rejectValue("fileId", "is required");
+        }
+        if (StringUtils.isBlank(file.getMimeType())) {
+            errors.rejectValue("mimeType", "is required");
+        }
```
DwayneJengSage: Does Mime Type need to be required?

src/test/java/org/sagebionetworks/bridge/validators/ParticipantFileValidatorTest.java
```diff
@@ -0,0 +1,28 @@
...
+    public void validates() {
+        ParticipantFile file = new DynamoParticipantFile("user", "file");
+        file.setMimeType("dummy");
+        file.setAppId("api");
+
+        Validate.entityThrowingException(INSTANCE, file);
+    }
+
+    @Test
+    public void fileIdRequired() {
```
DwayneJengSage, line 23: Need test cases for userId and appId.

src/test/java/org/sagebionetworks/bridge/validators/ParticipantFileValidatorTest.java
```diff
@@ -0,0 +1,28 @@
+package org.sagebionetworks.bridge.validators;
+
+import static org.sagebionetworks.bridge.TestUtils.assertValidatorMessage;
+import static org.sagebionetworks.bridge.validators.ParticipantFileValidator.INSTANCE;
+
+import org.sagebionetworks.bridge.dynamodb.DynamoParticipantFile;
```
DwayneJengSage: We want to decouple this from our Dynamo implementation. Use ParticipantFile.create() instead. (Here and in a bunch of other files as well.)

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dao/ParticipantFileDao.java
kss39: My bad, this is checked in the ParticipantFileService, but I forgot to delete this TODO. If the file already exists, the file service will throw EntityAlreadExistsException.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
kss39: Sorry, I don't quite understand what should I change here...?

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
kss39: I believe that the userID field is set by the Controller from the current active session, so the user cannot have access to other users' files...?

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,210 @@
...
+    @InjectMocks
+    DynamoParticipantFileDao dao;
+
+    @BeforeMethod
+    public void beforeMethod() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void getParticipantFiles() {
```
DwayneJengSage, line 85: The tests for getParticipantFiles should validate the query expression.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,166 @@
...
+import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.BridgeConstants.PAGE_SIZE_ERROR;
+import static org.sagebionetworks.bridge.validators.ParticipantFileValidator.INSTANCE;
+
+@Component
+public class ParticipantFileService {
+
+    static final int EXPIRATION_IN_MINUTES = 1440;
+
+    static final String PARTICIPANT_FILE_BUCKET = "participant-file.bucket";
```
DwayneJengSage, line 42: You also need an entry in src/main/resources/BridgeServer2.conf. (The value can be a dummy value, or a reference to a manually created org-sagebridge-participantfile-local bucket.) This creates the entry in the config, so the config system knows to look for the env var to overwrite it with.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
DwayneJengSage: We should move the logic from the Controller to the Service. This is because (a) we should put all the business logic in the Service, so that if we need to migrate Controllers, we can do so easily and (b) if another service or controller wants to call ParticipantFileService, the logic is already in the Service, so they don't need to replicate it.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,82 @@
...
+    public ParticipantFile createParticipantFile(@PathVariable String fileId) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+        String appId = session.getAppId();
+
+        ParticipantFile file = parseJson(DynamoParticipantFile.class);
+        file.setUserId(userId);
+        file.setCreatedOn(new DateTime());
+        file.setFileId(fileId);
+        file.setAppId(appId);
```
DwayneJengSage: We should move the logic from the Controller to the Service. This is because (a) we should put all the business logic in the Service, so that if we need to migrate Controllers, we can do so easily and (b) if another service or controller wants to call ParticipantFileService, the logic is already in the Service, so they don't need to replicate it.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -0,0 +1,166 @@
...
+
+        participantFileDao.getParticipantFile(file.getUserId(), file.getFileId()).ifPresent(
+                oldFile -> {
+                    Map<String, Object> entityKey = new HashMap<>();
+                    entityKey.put("oldFile", oldFile);
+                    throw new EntityAlreadyExistsException(ParticipantFile.class, entityKey);
+                }
+        );
+
+        file.setUploadUrl(generatePresignedRequest(file, PUT).toExternalForm());
```
DwayneJengSage, line 135: This won't do anything, because uploadURL is set with JsonIgnore.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Test
+    public void createParticipantFile() throws Exception {
+        ParticipantFile newFile = ParticipantFile.create();
+        newFile.setUserId("test_user");
+        newFile.setFileId("file_id");
+        newFile.setAppId("api");
+        newFile.setMimeType("dummy-type");
+        mockRequestBody(mockRequest, newFile);
+        when(mockFileService.createParticipantFile(any())).thenReturn(persisted);
```
DwayneJengSage: Should return newFile, not persisted.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileController.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @PostMapping("v3/participants/self/files/{fileId}")
+    @ResponseStatus(HttpStatus.CREATED)
+    public ParticipantFile createParticipantFile(@PathVariable String fileId) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+        String userId = session.getParticipant().getId();
+        String appId = session.getAppId();
+
+        ParticipantFile file = parseJson(DynamoParticipantFile.class);
```
DwayneJengSage: This should be ParticipantFile, not DynamoParticipantFile.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,178 @@
...
+        service.getParticipantFiles("test_user", null, 1);
+    }
+
+    @Test(expectedExceptions = BadRequestException.class)
+    public void getParticipantFilesPageSizeTooLarge() {
+        service.getParticipantFiles("test_user", null, 5000);
+    }
+
+    @Test
+    public void getParticipantFile() {
```
DwayneJengSage, line 102: This test should verify arguments into generatePresignedUrl().

Same with createParticipantFile().

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+        when(mockFileService.getParticipantFiles(eq("test_user"), isNull(), anyInt())).thenReturn(page);
+
+        ForwardCursorPagedResourceList<ParticipantFile> result = controller.getParticipantFiles(null, "5");
+        assertSame(page, result);
+
+        verify(mockFileService).getParticipantFiles(eq("test_user"), isNull(), eq(5));
+    }
+
+    @Test
+    public void getParticipantFile() {
```
DwayneJengSage, line 109: Should verify the Location header on the response()

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertSame(result, persisted);
+        verify(mockFileService).getParticipantFile("test_user", "file_id");
+    }
+
+    @Test
+    public void createParticipantFile() throws Exception {
+        ParticipantFile newFile = ParticipantFile.create();
+        newFile.setUserId("test_user");
+        newFile.setFileId("file_id");
+        newFile.setAppId("api");
```
DwayneJengSage: Don't set the app here, so we can verify that the controller is setting the app.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,51 @@
...
+        assertEquals(node.get("uploadUrl").textValue(), "dummy.upload");
+        assertEquals(node.size(), 8);
+    }
+
+    @Test
+    public void canDeserialize() throws Exception {
+
+        String json ="{\"fileId\":\"fileId\",\"userId\":\"userId\",\"createdOn\":\"2015-01-27T00:38:32.486Z\"," +
+                "\"mimeType\":\"image/jpeg\",\"appId\":\"api_test\",\"uploadUrl\":\"dummy.upload\"," +
+                "\"downloadUrl\":\"dummy.download\",\"type\":\"DynamoParticipantFile\"}";
```
DwayneJengSage: Type should be ParticipantFile.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,51 @@
...
+        JsonNode node = MAPPER.readTree(json);
+
+        assertEquals(node.get("userId").textValue(), "userId");
+        assertEquals(node.get("fileId").textValue(), "fileId");
+        assertEquals(node.get("createdOn").textValue(), TestConstants.TIMESTAMP.toString());
+        assertEquals(node.get("mimeType").textValue(), "image/jpeg");
+        assertEquals(node.get("appId").textValue(), "api_test");
+        assertEquals(node.get("downloadUrl").textValue(), "dummy.download");
+        assertEquals(node.get("uploadUrl").textValue(), "dummy.upload");
+        assertEquals(node.size(), 8);
```
DwayneJengSage, line 34: Verify the type is ParticipantFile.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Test
+    public void canDeserialize() throws Exception {
+
+        String json ="{\"fileId\":\"fileId\",\"userId\":\"userId\",\"createdOn\":\"2015-01-27T00:38:32.486Z\"," +
+                "\"mimeType\":\"image/jpeg\",\"appId\":\"api_test\",\"uploadUrl\":\"dummy.upload\"," +
+                "\"downloadUrl\":\"dummy.download\",\"type\":\"DynamoParticipantFile\"}";
+        DynamoParticipantFile file = MAPPER.readValue(json, DynamoParticipantFile.class);
```
DwayneJengSage: Use ParticipantFile.class

src/main/java/org/sagebionetworks/bridge/models/files/ParticipantFile.java
```diff
@@ -0,0 +1,27 @@
+package org.sagebionetworks.bridge.models.files;
+
+import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.dynamodb.DynamoParticipantFile;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+
+public interface ParticipantFile extends BridgeEntity {
```
DwayneJengSage, line 12: You need `@BridgeTypeName("ParticipantFile")` and `@JsonDeserialize(as = DynamoParticipantFile.class)`

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,231 @@
...
+        assertNotNull(result);
+        List<ParticipantFile> resultList = result.getItems();
+        assertEquals(resultList.size(), 1);
+        ParticipantFile resultFile = resultList.get(0);
+        assertEquals(resultFile, RESULT);
+
+        verify(mapper).query(any(), expressionCaptor.capture());
+        DynamoDBQueryExpression<ParticipantFile> expression = expressionCaptor.getValue();
+        assertEquals(expression.getLimit().intValue(), 5);
+        assertNull(expression.getExclusiveStartKey());
```
DwayneJengSage, line 99: We should also verify hashKeyValue and consistentRead.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,146 @@
...
+        newFile.setUserId("test_user");
+        newFile.setFileId("file_id");
+        newFile.setMimeType("dummy-type");
+        mockRequestBody(mockRequest, newFile);
+        when(mockFileService.createParticipantFile(any(), any(), any())).thenAnswer(i ->
+        {
+            newFile.setAppId(i.getArgument(0));
+            newFile.setCreatedOn(TestConstants.TIMESTAMP);
+            return newFile;
+        });
```
DwayneJengSage: It's not necessary to mock these behaviors. We want to test the behavior of the Controller, not the Service.

Also, you need to verify the inputs to the service.createParticipantFile() call.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+        assertEquals(result.getDownloadUrl(), downloadUrl);
+        assertEquals(result.getAppId(), "api");
+
+        verify(mockS3Client).generatePresignedUrl(requestCaptor.capture());
+        GeneratePresignedUrlRequest request = requestCaptor.getValue();
+        assertEquals(request.getMethod(), HttpMethod.GET);
+        assertEquals(request.getContentType(), file.getMimeType());
+        assertEquals(request.getKey(), "test_user/file_id");
+        assertEquals(request.getRequestParameters().get(Headers.SERVER_SIDE_ENCRYPTION),
+                ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
```
DwayneJengSage: Should also validate bucket and expiration. (Same with createParticipantFile().)

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,63 @@
...
+        JsonNode node = MAPPER.readTree(json);
+
+        assertEquals(node.get("userId").textValue(), "userId");
+        assertEquals(node.get("fileId").textValue(), "fileId");
+        assertEquals(node.get("createdOn").textValue(), TestConstants.TIMESTAMP.toString());
+        assertEquals(node.get("mimeType").textValue(), "image/jpeg");
+        assertEquals(node.get("appId").textValue(), "api_test");
+        assertEquals(node.get("downloadUrl").textValue(), "dummy.download");
+        assertEquals(node.get("uploadUrl").textValue(), "dummy.upload");
+        assertEquals(node.size(), 8);
```
DwayneJengSage, line 34: Verify the type is ParticipantFile.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        when(mockFileDao.getParticipantFile(any(), any())).thenReturn(Optional.empty());
+
+        // UserId and AppId should not depend on file, it should be manually set by the Service.
+        ParticipantFile file = ParticipantFile.create();
+        file.setFileId("file_id");
+        file.setUserId("wrong_user");
+        file.setMimeType("dummy-type");
+        file.setAppId("wrong_api");
+        file.setCreatedOn(TestConstants.TIMESTAMP);
```
DwayneJengSage: Don't set the timestamp here, so we can verify that the service sets the timestamp.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,206 @@
...
+    ParticipantFileDao mockFileDao;
+
+    @Mock
+    BridgeConfig mockConfig;
+
+    @Mock
+    AmazonS3 mockS3Client;
+
+    @InjectMocks
+    @Spy
```
DwayneJengSage: You don't need to spy if you aren't spying anything.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertEquals(deser.getDownloadUrl(), "dummy.download");
+        assertEquals(deser.getUploadUrl(), "dummy.upload");
+    }
+
+    @Test
+    public void canDeserialize() throws Exception {
+
+        String json ="{\"fileId\":\"fileId\",\"userId\":\"userId\",\"createdOn\":\"2015-01-27T00:38:32.486Z\"," +
+                "\"mimeType\":\"image/jpeg\",\"appId\":\"api_test\",\"uploadUrl\":\"dummy.upload\"," +
+                "\"downloadUrl\":\"dummy.download\",\"type\":\"DynamoParticipantFile\"}";
```
DwayneJengSage: Type should be ParticipantFile.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDao.java
```diff
@@ -0,0 +1,97 @@
...
+                .withLimit(pageSize)
+                .setConsistentRead(true);
+        if (offsetKey != null) {
+            HashMap<String, AttributeValue> startKey = new HashMap<>();
+            startKey.put("userId", new AttributeValue().withS(userId));
+            startKey.put("fileId", new AttributeValue().withS(offsetKey));
+            queryExpression.withExclusiveStartKey(startKey);
+        }
+
+        QueryResultPage<DynamoParticipantFile> results = mapper.queryPage(DynamoParticipantFile.class, queryExpression);
```
DwayneJengSage: We want query() not queryPage(). queryPage() paginates by DynamoDB paging, which doesn't necessarily line up with Bridge paging.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,221 @@
...
+        assertEquals(result.getAppId(), "api");
+
+        verify(mockS3Client).generatePresignedUrl(requestCaptor.capture());
+        GeneratePresignedUrlRequest request = requestCaptor.getValue();
+        assertEquals(request.getBucketName(), UPLOAD_BUCKET);
+        assertEquals(request.getMethod(), HttpMethod.GET);
+        assertEquals(request.getContentType(), file.getMimeType());
+        assertEquals(request.getKey(), "test_user/file_id");
+        assertEquals(request.getRequestParameters().get(Headers.SERVER_SIDE_ENCRYPTION),
+                ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
```
DwayneJengSage: Need to verify expiration.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,221 @@
...
+        assertEquals(result.getAppId(), "api");
+        assertEquals(TestConstants.TIMESTAMP.compareTo(result.getCreatedOn()), 0);
+
+        verify(mockS3Client).generatePresignedUrl(requestCaptor.capture());
+        GeneratePresignedUrlRequest request = requestCaptor.getValue();
+        assertEquals(request.getMethod(), HttpMethod.PUT);
+        assertEquals(request.getContentType(), file.getMimeType());
+        assertEquals(request.getKey(), "test_user/file_id");
+        assertEquals(request.getRequestParameters().get(Headers.SERVER_SIDE_ENCRYPTION),
+                ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
```
DwayneJengSage: Need to verify bucket and expiration.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+        verify(mockFileDao).getParticipantFile(any(), any());
+        verify(mockFileDao).uploadParticipantFile(any());
```
DwayneJengSage: We should verify the inputs into the dao.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantFileControllerTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Test
+    public void createParticipantFile() throws Exception {
+        ParticipantFile newFile = ParticipantFile.create();
+        newFile.setUserId("test_user");
+        newFile.setFileId("file_id");
+        newFile.setMimeType("dummy-type");
+        mockRequestBody(mockRequest, newFile);
+        when(mockFileService.createParticipantFile(any(), any(), any())).thenReturn(newFile);
+        ParticipantFile result = controller.createParticipantFile("file_id");
```
DwayneJengSage, line 126: We should use a different file ID so we can verify that the controller sets the file ID from the input parameters.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFileDaoTest.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Test
+    public void getParticipantFilesPageSize() {
+        when(resultPage.stream()).thenReturn(RESULT_LIST.stream());
+        when(mapper.query(eq(DynamoParticipantFile.class), any())).thenReturn(resultPage);
+//        when(resultPage.getLastEvaluatedKey())
+//                .thenReturn(Collections.singletonMap("fileId", new AttributeValue("file4")));
+//        when(resultPage.getResults()).thenReturn(RESULT_LIST);
+//        when(mapper.queryPage(eq(DynamoParticipantFile.class), any())).thenReturn(resultPage);
```
DwayneJengSage: Minor: Delete commented out code.

DwayneJengSage (approved):
Looks good. One minor comment about commented out code.

DwayneJengSage (approved):

PR #267, opened 2020-08-24
Missing apis

alxdarksage:
/v3/files/{guid}/revisions/{createdOn} - get an individual file revision, so we can show the metadata in the UI.

/v1/organizations/nonmembers - a developer API to retrieve administrative accounts that are not yet assigned to an organization (so we don't have to open up the /v3/participants APIs to developers). This might change when we add an organizational administrator, *that* role would then be needed for this API, although we'll probably do the association when creating the account as well.


src/main/java/org/sagebionetworks/bridge/services/FileService.java
```diff
@@ -158,7 +158,11 @@ public void deleteAllAppFiles(String appId) {
         checkNotNull(guid);
         checkNotNull(createdOn);
 
-        return fileRevisionDao.getFileRevision(guid, createdOn);
+        Optional<FileRevision> opt = fileRevisionDao.getFileRevision(guid, createdOn);
+        if (opt.isPresent()) {
+            opt.get().setDownloadURL(getDownloadURL(opt.get()));
+        }
```
DwayneJengSage, line 164: Do we have a unit test for this test case?

DwayneJengSage (changes requested):
Looks good. One question about a missing unit test?

src/main/java/org/sagebionetworks/bridge/services/FileService.java
alxdarksage: Added case where optional is empty.

DwayneJengSage (approved):

PR #270, opened 2020-09-01
enrollment API

alxdarksage:
An API to directly enroll users in studies (represented by the Enrollments or AccountsSubstudies table). 

- adds some extra metadata about the enrollment or withdrawal event;
- allows caller to register an external ID during enrollment without adding it to the external IDs table (which I will remove after migration)
- provides an API to show, for a given study, the people who are enrolled in and/or withdrawn from that study.

For further discussion of the API itself, see: https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/886440163/Bridge+v2+Implementation+Design#Enrollments

Eventually these APIs, which are scoped to particular organizations, will functionally replace the global participant APIs.

This is not articulated with our existing consent system, because once we force assignment to a study (which you can already optionally do when a user consents), our existing permissions system will cause those accounts to be hidden in some cases from existing administrative users (who we want to remove from the AccountSubstudies table). So future check-ins will further prepare to migrate all accounts (that can be migrated) to be members of an organization, or enrolled in a study, without losing the visibility of accounts through the API.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
@@ -1,0 +1,194 @@
+package org.sagebionetworks.bridge.services;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.AuthUtils.checkSelfAdminOrSponsorAndThrow;
+import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.BridgeConstants.NEGATIVE_OFFSET_ERROR;
+import static org.sagebionetworks.bridge.BridgeConstants.PAGE_SIZE_ERROR;
+import static org.sagebionetworks.bridge.models.ResourceList.ENROLLMENT_FILTER;
+import static org.sagebionetworks.bridge.models.ResourceList.OFFSET_BY;
+import static org.sagebionetworks.bridge.models.ResourceList.PAGE_SIZE;
+import static org.sagebionetworks.bridge.validators.EnrollmentValidator.INSTANCE;
+
+import com.google.common.collect.ImmutableMap;
+
+import org.joda.time.DateTime;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.BridgeUtils;
+import org.sagebionetworks.bridge.dao.EnrollmentDao;
+import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.models.PagedResourceList;
+import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
+import org.sagebionetworks.bridge.models.studies.Enrollment;
+import org.sagebionetworks.bridge.models.studies.EnrollmentFilter;
+import org.sagebionetworks.bridge.validators.Validate;
+
+@Component
+public class EnrollmentService {
+    
+    private AccountService accountService;
+    
+    private SponsorService sponsorService;
+    
+    private EnrollmentDao enrollmentDao;
+    
+    @Autowired
+    public final void setAccountService(AccountService accountService) {
+        this.accountService = accountService;
+    }
+    
+    @Autowired
+    public final void setSponsorService(SponsorService sponsorService) {
+        this.sponsorService = sponsorService;
+    }
+    
+    @Autowired
+    public final void setEnrollmentDao(EnrollmentDao enrollmentDao) {
+        this.enrollmentDao = enrollmentDao;
+    }
+    
+    protected DateTime getEnrollmentDateTime() {
+        return DateTime.now();
+    }
+
+    protected DateTime getWithdrawalDateTime() {
+        return DateTime.now();
+    }
+    
+    /**
+     * Get enrollments in a study. This API will be expanded to retrieve and sort the data for 
+     * common reporting requirements (e.g. how many people have withdrawn from the study).
+     */
+    public PagedResourceList<Enrollment> getEnrollmentsForStudy(String appId, String studyId, 
+            EnrollmentFilter filter, Integer offsetBy, Integer pageSize) {
+        checkNotNull(appId);
+        checkNotNull(studyId);
+        
+        checkSelfAdminOrSponsorAndThrow(sponsorService, studyId, null);
+
+        if (offsetBy != null && offsetBy < 0) {
+            throw new BadRequestException(NEGATIVE_OFFSET_ERROR);
+        }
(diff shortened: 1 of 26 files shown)


src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+    
+    /**
+     * For methods that are going to save the account, this method handles enrollment but does not persist it.
+     */
+    public Enrollment enroll(Account account, Enrollment enrollment) {
+        checkNotNull(account);
+        checkNotNull(enrollment);
+        
+        Validate.entityThrowingException(INSTANCE, enrollment);
```
DwayneJengSage, line 94: Do we have a unit test for if we call enroll() with an invalid enrollment?

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -0,0 +1,194 @@
...
+        enrollment.setExternalId(enrollment.getExternalId());
+        if (enrollment.getEnrolledOn() == null) {
+            enrollment.setEnrolledOn(getEnrollmentDateTime());
+        }
+        String callerUserId = BridgeUtils.getRequestContext().getCallerUserId();
+        if (!account.getId().equals(callerUserId)) {
+            enrollment.setEnrolledBy(callerUserId);
+        }
+
+        account.getEnrollments().add(enrollment);
```
DwayneJengSage, line 123: Unit tests that verify we added the enrollment?

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+    
+    /**
+     * For methods that are going to save the account, this method withdraws the account but does not persist it.
+     */
+    public Enrollment unenroll(Account account, Enrollment enrollment) {
+        checkNotNull(account);
+        checkNotNull(enrollment);
+        
+        Validate.entityThrowingException(INSTANCE, enrollment);
```
DwayneJengSage, line 150: Unit test for invalid enrollment for unenroll()?

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
```diff
@@ -0,0 +1,471 @@
...
+        Enrollment existing = Enrollment.create(TEST_APP_ID, TEST_STUDY_ID, USER_ID);
+        Enrollment otherStudy = Enrollment.create(TEST_APP_ID, "otherStudy", USER_ID);
+        
+        Account account = Account.create();
+        account.setId(USER_ID);
+        account.setEnrollments(Sets.newHashSet(otherStudy, existing));
+        when(mockAccountService.getAccount(ACCOUNT_ID)).thenReturn(account);
+        
+        Enrollment enrollment = Enrollment.create(TEST_APP_ID, TEST_STUDY_ID, USER_ID);
+        enrollment.setWithdrawnOn(MODIFIED_ON);
```
DwayneJengSage: Should this timestamp be different from the default, so we can verify that we read and set this from the inputted enrollment.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -0,0 +1,194 @@
...
+        
+        checkSelfAdminOrSponsorAndThrow(sponsorService, enrollment.getStudyId(), enrollment.getAccountId());
+        
+        // If supplied, this value should be the same timestamp as the withdrewOn
+        // value in the signature. Otherwise just set it here. 
+        if (enrollment.getWithdrawnOn() == null) {
+            enrollment.setWithdrawnOn(getWithdrawalDateTime());
+        }
```
DwayneJengSage, line 158: Unit test for when withdrawnOn is not specified?

src/test/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentControllerTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Enrollment retValue = controller.enroll(TEST_STUDY_ID);
+        assertSame(retValue, completed);
+        
+        verify(mockService).enroll(enrollmentCaptor.capture());
+        
+        Enrollment value = enrollmentCaptor.getValue();
+        assertEquals(value.getEnrolledOn(), CREATED_ON);
+        assertEquals(value.getExternalId(), "anExternalId");
+        assertEquals(value.getAccountId(), USER_ID);
+        assertTrue(value.isConsentRequired());
```
DwayneJengSage: Should we assert that we added App and Study?

src/main/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentController.java
```diff
@@ -0,0 +1,72 @@
...
+        UserSession session = getAuthenticatedSession(RESEARCHER, ADMIN);
+        
+        Enrollment enrollment = parseJson(Enrollment.class);
+        enrollment.setAppId(session.getAppId());
+        enrollment.setStudyId(studyId);
+        
+        return service.enroll(enrollment);
+    }
+    
+    @PostMapping("/v5/studies/{studyId}/enrollments/{userId}")
```
DwayneJengSage: It's a little confusing to me that unenroll is a POST. Would this make more sense a DELETE? Or if not a delete, maybe a special URL with "unenroll" somewhere?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
alxdarksage: Changed this, and added a test that the value is defaulted to "now" if a timestamp is _not_ provided for withdrawnOn.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentControllerTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentController.java
alxdarksage: Our consent withdrawal is a POST but I would rather use DELETE. I've changed this.

alxdarksage: Ah, it turns out our Java client libraries don't support body content with a DELETE call. As the only value here is the short reason text, I will append it as a query parameter to the DELETE call.

DwayneJengSage (approved):

PR #271, opened 2020-09-11
Enrollments refactor

alxdarksage:
- adding studies that are available through organizational membership to the RequestContext;
- spent some time refactoring RC so it would be cleaner to use, consolidating code from a couple of places, which was pushed by the above change.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
@@ -13,7 +13,11 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 import static org.sagebionetworks.bridge.RequestContext.NULL_INSTANCE;
 import static org.sagebionetworks.bridge.Roles.DEVELOPER;
+import static org.sagebionetworks.bridge.TestConstants.LANGUAGES;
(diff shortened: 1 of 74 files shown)


src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -1284,6 +1294,12 @@ public void getSessionFromAccount() {
         assertEquals(session.getConsentStatuses(), CONSENTED_STATUS_MAP);
         
         verify(accountSecretDao).createSecret(AccountSecretType.REAUTH, USER_ID, REAUTH_TOKEN);
+        
+        RequestContext retValue = RequestContext.updateFromSession(session, null);
```
DwayneJengSage, line 1298: Would it make more sense to call RequestContext.get() here?

DwayneJengSage (approved):
Looks good. One comment about tests.

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
alxdarksage: When I do this in Eclipse, it gives me a warning: "The static method updateFromSession(UserSession, SponsorService) from the type RequestContext should be accessed in a static way." (Findbugs does not complain). So that's why I called it as a static method.

PR #274, opened 2020-09-21
Integrate enrollments into our consent system

alxdark:
Some changes that will be visible to end users:

- required subpopulations *must* now declare a study that a user will be enrolled in when the user consents;
- consent and withdrawal of consent will create enrollment records;
- permissions reworked so that in some cases, if we're limiting access to a developer or researcher, we're also checking that they are in an organization that gives them access to the relevant study.

These changes must be accompanied by integration test changes.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -4,7 +4,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.sagebionetworks.bridge.BridgeConstants.CALLER_NOT_MEMBER_ERROR;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.RESEARCHER;
-import static org.sagebionetworks.bridge.Roles.SUPERADMIN;
+
+import java.util.Set;
 
 import com.google.common.collect.ImmutableSet;
 
@@ -12,29 +13,32 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.sagebionetworks.bridge.exceptions.UnauthorizedException;
-import org.sagebionetworks.bridge.services.SponsorService;
 
 public class AuthUtils {
     private static final Logger LOG = LoggerFactory.getLogger(AuthUtils.class);
-    
-    public static boolean checkSelfOrResearcher(String targetUserId) {
-        RequestContext context = RequestContext.get();
-        String callerUserId = context.getCallerUserId();
-        
-        return context.isInRole(RESEARCHER) || targetUserId.equals(callerUserId);
+
+    public static boolean checkResearcherOrAdmin(String studyId) {
+        return isInRoles(studyId, ADMIN, RESEARCHER);
     }
     
-    public static void checkSelfOrResearcherAndThrow(String targetUserId) {
-        if (!checkSelfOrResearcher(targetUserId)) {
+    public static void checkResearcherOrAdminAndThrow(String studyId) {
+        if (!checkResearcherOrAdmin(studyId)) {
+            throw new UnauthorizedException();
+        }
+    }
+    
+    public static boolean checkSelfResearcherOrAdmin(String studyId, String targetUserId) {
+        return isSelf(targetUserId) || isInRoles(studyId, ADMIN, RESEARCHER);
+    }
+    
+    public static void checkSelfResearcherOrAdminAndThrow(String studyId, String targetUserId) {
+        if (!checkSelfResearcherOrAdmin(studyId, targetUserId)) {
             throw new UnauthorizedException();
         }
     }
     
     public static boolean checkOrgMembership(String targetOrgId) {
-        RequestContext context = RequestContext.get();
-        String callerOrgMembership = context.getCallerOrgMembership();
-        
-        return context.isInRole(ADMIN) || targetOrgId.equals(callerOrgMembership);
+        return isInOrganization(targetOrgId);
     }
     
     public static void checkOrgMembershipAndThrow(String targetOrgId) {
@@ -48,18 +52,16 @@ public class AuthUtils {
      * organization must match the caller's organization. We do not need to validate the org ID since 
      * it was validated when it was set as an organizational relationship on the account. 
      */
-    public static void checkAssessmentOwnership(String appId, String ownerId) {
-        checkNotNull(appId);
-        checkNotNull(ownerId);
-        
-        RequestContext rc = RequestContext.get();
-        String callerOrgMembership = rc.getCallerOrgMembership();
-        if (rc.isInRole(ImmutableSet.of(SUPERADMIN, ADMIN)) || ownerId.equals(callerOrgMembership)) {
-            return;
-        }
-        throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+    public static boolean checkAssessmentOwnership(String appId, String ownerId) {
+        return isInOrganization(ownerId);
     }
-    
+
+    public static void checkAssessmentOwnershipAndThrow(String appId, String ownerId) {
+        if (!checkAssessmentOwnership(appId, ownerId)) {
+            throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
+        }
+    }
+
     /**
      * The same rules apply as checkOwnership, however you are examining the caller against a compound
      * value stored in the assessment's originId. The call succeeds if you're a superadmin (there are 
@@ -71,8 +73,8 @@ public class AuthUtils {
         checkNotNull(guid);
         checkNotNull(ownerId);
 
-        RequestContext rc = RequestContext.get();
-        if (rc.isInRole(ImmutableSet.of(ADMIN))) {
+        RequestContext context = RequestContext.get();
+        if (context.isInRole(ImmutableSet.of(ADMIN))) {
             return;
         }
         String[] parts = ownerId.split(":", 2);
@@ -83,25 +85,36 @@ public class AuthUtils {
         }
         String originAppId = parts[0];
         String originOrgId = parts[1];
-        String callerOrgMembership = rc.getCallerOrgMembership();
+        String callerOrgMembership = context.getCallerOrgMembership();
         if (originAppId.equals(callerAppId) && originOrgId.equals(callerOrgMembership)) {
             return;
         }
         throw new UnauthorizedException(CALLER_NOT_MEMBER_ERROR);
     }
     
-    public static boolean checkSelfAdminOrSponsor(SponsorService sponsorService, String studyId, String userId) {
+    /**
+     * If the user is an admin or a superadmin, this returns true. For other roles, the 
+     * caller must be in one of the roles, and they must be a member of an organization 
+     * that gives them access to the study.
+     */
+    private static final boolean isInRoles(String studyId, Roles... roles) {
+        Set<Roles> roleSet = ImmutableSet.copyOf(roles);
         RequestContext context = RequestContext.get();
-        String callerOrgMembership = context.getCallerOrgMembership();
-        String callerUserId = context.getCallerUserId();
-        return context.isInRole(ADMIN) || 
-                sponsorService.isStudySponsoredBy(studyId, callerOrgMembership) ||
-                (callerUserId != null && callerUserId.equals(userId));
+        if (roleSet.contains(ADMIN) && context.isInRole(ADMIN)) {
+            return true;
+        }
+        Set<String> sponsoredStudies = context.getOrgSponsoredStudies();
+        boolean canAccessStudy = sponsoredStudies.isEmpty() || studyId == null 
+                || sponsoredStudies.contains(studyId);
+        return (canAccessStudy && context.isInRole(roleSet));
     }
     
-    public static void checkSelfAdminOrSponsorAndThrow(SponsorService sponsorService, String studyId, String userId) {
-        if (!checkSelfAdminOrSponsor(sponsorService, studyId, userId)) {
-            throw new UnauthorizedException();
-        }
+    private static final boolean isInOrganization(String orgId) {
+        RequestContext context = RequestContext.get();
+        return context.isInRole(ADMIN) || orgId.equals(context.getCallerOrgMembership());
+    }
+    
+    private static final boolean isSelf(String userId) {
+        return userId.equals(RequestContext.get().getCallerUserId());
     }
 }
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -62,7 +62,7 @@ public class AuthUtils {
     
     /**
      * The same rules apply as checkOwnership, however you are examining the caller against a compound
-     * value stored in the assessment's originId. The call succeeds if you're a superadmin (ther are 
+     * value stored in the assessment's originId. The call succeeds if you're a superadmin (there are 
      * no shared app admins), or if the caller is in the app and organization that owns the shared 
      * assessment. 
      */
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -84,7 +84,7 @@ public class AuthUtils {
      * caller must be in one of the roles, and they must be a member of an organization 
      * that gives them access to the study.
      */
-    private static final boolean isInRoles(String studyId, Roles... roles) {
+    public static final boolean isInRoles(String studyId, Roles... roles) {
         Set<Roles> roleSet = ImmutableSet.copyOf(roles);
         RequestContext context = RequestContext.get();
         if (roleSet.contains(ADMIN) && context.isInRole(ADMIN)) {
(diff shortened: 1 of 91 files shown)


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -4,37 +4,41 @@
...
-    public static boolean checkSelfOrResearcher(String targetUserId) {
-        RequestContext context = RequestContext.get();
-        String callerUserId = context.getCallerUserId();
-        
-        return context.isInRole(RESEARCHER) || targetUserId.equals(callerUserId);
+    public static void checkResearcherOrAdminAndThrow(String studyId) {
+        if (!checkResearcherOrAdmin(studyId)) {
+            throw new UnauthorizedException();
+        }
+    }
```
DwayneJengSage: Do we need unit tests for these methods?

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
```diff
@@ -168,7 +167,7 @@ public void checkSelfOrResearcherFails() {
                 .withCallerRoles(ImmutableSet.of(DEVELOPER))
                 .withCallerUserId("notUserId").build());
         
-        assertFalse(AuthUtils.checkSelfOrResearcher(USER_ID));
+        assertFalse(AuthUtils.checkSelfResearcherOrAdmin(TEST_STUDY_ID, USER_ID));
```
DwayneJengSage: What about versions of this unit test for

* admins
* researcher is in the correct study
* researcher is in the wrong study
* called with null studyId

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -135,34 +135,48 @@ public static boolean isExternalIdAccount(StudyParticipant participant) {
...
-            Set<String> callerStudies = context.getCallerStudies();
-            if (BridgeUtils.isEmpty(callerStudies)) {
+            Set<String> callerStudies = context.getOrgSponsoredStudies();
+            String callerUserId = context.getCallerUserId();
+            
+            // If this is a call for one’s own record, or the caller is has no 
+            // study relationships, return the account. Note that we are migrating
+            // away from a system that allows an account to have no specific study
+            // associations, but it remains necessary during the transition
+            if (isEmpty(callerStudies) || account.getId().equals(callerUserId)) {
```
DwayneJengSage, line 169: Do we have a unit test case for getting your own account?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
```diff
@@ -162,9 +163,13 @@ QueryBuilder makeQuery(String prefix, String appId, AccountId accountId, Account
...
-        if (!callerStudies.isEmpty()) {
-            builder.append("AND enrollment.studyId IN (:studies)", "studies", callerStudies);
+        
+        // If the caller is a non-admin administrator (researcher, worker, other roles we dream up), then
+        // their organization dictates the accounts they see. We are still making an exception for accounts
+        // with no study associations, while we transition to the new permissions system.
+        Set<String> callerStudies = context.getOrgSponsoredStudies();
+        if (context.isAdministrator() && !context.isInRole(ADMIN) && !callerStudies.isEmpty()) {
+            builder.append("AND enrollment.studyId IN (:studies)", "studies", callerStudies);        
         }
```
DwayneJengSage: Unit tests for Admins being able to get anything?

Also, in what situations would context.isAdministrator() be false?

src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
```diff
@@ -52,7 +51,6 @@ static Enrollment create(String appId, String studyId, String accountId, String
     String getAppId();
     void setAppId(String appId);
     
-    @JsonIgnore
     String getStudyId();
     void setStudyId(String studyId);
```
DwayneJengSage: Do we have a Json serialization test for this class?

src/main/java/org/sagebionetworks/bridge/services/AppService.java
```diff
@@ -342,7 +346,20 @@ public App createApp(App app) {
             throw new EntityAlreadyExistsException(App.class, IDENTIFIER_PROPERTY, app.getIdentifier());
         }
         
-        subpopService.createDefaultSubpopulation(app);
+        Study study = Study.create();
+        study.setAppId(app.getIdentifier());
+        study.setIdentifier(app.getIdentifier() + "-study");
+        study.setName("Initial Study");
```
DwayneJengSage: Would it be more confusing or less confusing if the default study name was the same as the app name?

src/test/java/org/sagebionetworks/bridge/services/AppServiceTest.java
```diff
@@ -1589,8 +1590,10 @@ public void crudApp() {
 
         verify(mockCacheProvider).setApp(app);
         
+        ArgumentCaptor<Study> studyCaptor = ArgumentCaptor.forClass(Study.class);
+        
         // A default, active consent should be created for the app.
-        verify(mockSubpopService).createDefaultSubpopulation(app);
+        verify(mockSubpopService).createDefaultSubpopulation(eq(app), studyCaptor.capture());
```
DwayneJengSage, line 1603: Should we also verify the calls to studyService and organizationService?

src/test/java/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
```diff
@@ -358,11 +358,19 @@ public void withdrawConsentWithParticipant() throws Exception {
...
         assertEquals(email.getMessageParts().get(0).getContent(), "<p>User   &lt;" + EMAIL
                 + "&gt; withdrew from the study on October 28, 2015. </p><p>Reason:</p><p>For reasons.</p>");
+        
+        verify(mockEnrollmentService, times(1)).unenroll(eq(account), enrollmentCaptor.capture());
+        Enrollment withdrawalEnrollment = enrollmentCaptor.getValue();
+        assertEquals(withdrawalEnrollment.getWithdrawalNote(), WITHDRAWAL.getReason());
+        assertEquals(withdrawalEnrollment.getWithdrawnOn().getMillis(), SIGNED_ON + 10000);
+        assertEquals(withdrawalEnrollment.getAppId(), app.getIdentifier());
+        assertEquals(withdrawalEnrollment.getStudyId(), TEST_STUDY_ID);
+        assertEquals(withdrawalEnrollment.getAccountId(), ID);
```
DwayneJengSage, line 371: What about the same test for withdrawConsentForApp?

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -181,6 +181,7 @@ public void deleteOrganization(String appId, String identifier) {
         
         AccountSummarySearch scopedSearch = new AccountSummarySearch.Builder()
                 .copyOf(search)
+                .withAdminOnly(true)
```
DwayneJengSage, line 184: Add something to the unit test to check that this value is true?

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
```diff
@@ -24,11 +24,10 @@
     
     @BeforeMethod
     public void before() {
-        validator = new SubpopulationValidator(TestConstants.USER_DATA_GROUPS, TestConstants.USER_STUDY_IDS);
+        validator = new SubpopulationValidator(USER_DATA_GROUPS, USER_STUDY_IDS);
     }
     
-    @Test
-    public void testEntirelyValid() {
+    Subpopulation getSubpopulation() {
```
DwayneJengSage: This doesn't appear to be used anywhere.

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
```diff
@@ -24,11 +24,10 @@
     
     @BeforeMethod
     public void before() {
-        validator = new SubpopulationValidator(TestConstants.USER_DATA_GROUPS, TestConstants.USER_STUDY_IDS);
+        validator = new SubpopulationValidator(USER_DATA_GROUPS, USER_STUDY_IDS);
     }
     
-    @Test
-    public void testEntirelyValid() {
```
DwayneJengSage: Is there a replacement for this deleted test?

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
```diff
@@ -97,16 +94,17 @@ public void nullListsOK() {
...
-        Map<String,List<String>> errors = e.getErrors();
-        List<String> messages = errors.get(propName);
-        assertTrue(messages.get(0).contains(propName + error));
+    @Test
+    public void test() {
+        Subpopulation subpop = Subpopulation.create();
+        subpop.setAppId(TEST_APP_ID);
+        subpop.setGuidString("AAA");
+        subpop.setName("Name");
     }
```
DwayneJengSage, line 122: This test doesn't appear to test anything.

src/main/java/org/sagebionetworks/bridge/validators/SubpopulationValidator.java
```diff
@@ -46,13 +48,19 @@ public void validate(Object object, Errors errors) {
...
             }
         }
-        for (String studyId : subpop.getStudyIdsAssignedOnConsent()) {
-            if (!studyIds.contains(studyId)) {
-                String listStr = (studyIds.isEmpty()) ? "<empty>" : COMMA_SPACE_JOINER.join(studyIds);
-                String message = String.format("'%s' is not in enumeration: %s", studyId, listStr);
-                errors.rejectValue("studyIdsAssignedOnConsent", message);
+        if (subpop.getStudyIdsAssignedOnConsent().isEmpty()) {
+            errors.rejectValue("studyIdsAssignedOnConsent", CANNOT_BE_EMPTY);
+        } else {
```
DwayneJengSage, line 53: Unit test case for empty studyIdsAssignedOnConsent?

src/test/java/org/sagebionetworks/bridge/services/ExternalIdServiceTest.java
```diff
@@ -141,24 +141,6 @@ public void createExternalIdEnforcesAppId() {
...
-                .withCallerAppId(TEST_APP_ID)
-                .withCallerStudies(STUDIES).build());
-        
-        ExternalIdentifier newExtId = ExternalIdentifier.create(TEST_APP_ID,
-                extId.getIdentifier());
-        externalIdService.createExternalId(newExtId, false);
-        
-        // still matches and verifies
-        verify(externalIdDao).createExternalId(extId);
-    }
```
DwayneJengSage: Was this test accidentally deleted?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -637,20 +637,15 @@ private void updateAccountAndRoles(App app, Account account, StudyParticipant pa
...
-        Set<String> callerStudies = RequestContext.get().getCallerStudies();
+        // If the caller is not in a study, any study tags are allowed. If the caller is
+        // assigned to any studies, and also adding studies to the account, then the studies
+        // must all be associated to the caller or it is a security violation.
+        Set<String> callerStudies = RequestContext.get().getOrgSponsoredStudies();
         if (!callerStudies.isEmpty()) {
             Set<String> studyIds = BridgeUtils.collectStudyIds(account);
-            if (studyIds.isEmpty()) {
-                throw new BadRequestException("Participant must be assigned to one or more of these studies: "
-                        + BridgeUtils.COMMA_JOINER.join(callerStudies));
```
DwayneJengSage, line 648: If we delete this block, what prevents callers from creating participants not associated to any study?

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -671,15 +672,6 @@ public void createParticipantSynapseUserIdWithDeviationIsDisabled() {
...
     }
     
-    @Test(expectedExceptions = BadRequestException.class,
-            expectedExceptionsMessageRegExp=".*must be assigned to one or more of these studies: studyId.*")
-    public void createParticipantMustIncludeCallerStudy() {
-        RequestContext.set(new RequestContext.Builder()
-                .withCallerStudies(ImmutableSet.of(STUDY_ID)).build());
-        
-        participantService.createParticipant(APP, PARTICIPANT, false);
-    }
```
DwayneJengSage: What's the expected outcome for this scenario? Should we still have a test to test that scenario?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: I ran code coverage and added tests for checkSelfStudyResearcherOrAdminSucceedsForSelf and isInRole which were still not fully covered (this was updated by the separate AuthUtils PR to simplify the number and names of the methods in AuthUtils).

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
alxdarksage: The complexity comes from the isInRole method which now tests your study access, if it has been declared (after migration it will get more stringent). Made the method public and added tests for these different scenarios.

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDao.java
alxdarksage: context.isAdministrator() doesn't seem to be needed, I don't see any tests or code path where this is executed for a participant account. Unit and integration tests pass without it. Removed. Added a test in HibernateAccountDao to verify that this part of query is skipped for admins who otherwise are a member of an organization.

src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
alxdarksage: No, because it's no longer returned from the DAO, all calls wrap it in a details object (that is tested for its serialization in the API). I allowed the studyId to be visible because when a user asks for their own set of enrollment records, they need to see which record goes to which study.

src/main/java/org/sagebionetworks/bridge/services/AppService.java
alxdarksage: I don't know. I'm okay with that, because it's easily changed or deleted. In many cases one app may still have one study and it would make sense. Updated.

src/test/java/org/sagebionetworks/bridge/services/AppServiceTest.java
alxdarksage: Yes. Added.

src/test/java/org/sagebionetworks/bridge/services/ConsentServiceMockTest.java
alxdarksage: Added.

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
alxdarksage: Added.

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
alxdarksage: Removed. I had split this test up into testing each field separately, which would be better, but then decided against it because it just added more changes to review.

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
alxdarksage: Added back

src/test/java/org/sagebionetworks/bridge/validators/SubpopulationValidatorTest.java
alxdarksage: It verifies that the null collection is converted to an empty list, and is treated as valid.

src/main/java/org/sagebionetworks/bridge/validators/SubpopulationValidator.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: This is now allowed, since this association is taken from enrollments. When an admin updates another admin, they might have rights to see participants in specific organizations, but that doesn't mean they have to enroll that admin into one of those studies. Similarly, updating an account record doesn't imply that the user consented to enroll in a researcher study. Study IDs in the participant object is going to become read-only as part of cleaning up the current external ID code in favor of the enrollment API, then enrollments won't be changed by account create/update at all.

nategbrown9 (approved):
Looks good pending review of comments.

PR #276, opened 2020-09-25
Refactor AuthUtils to consolidate and clarify the checks.

alxdarksage:
Removing some methods that were too specifically named for what they were checking; consolidate some methods that were actually doing the same thing; removed public methods that return a boolean (not really being used); and shortened method names by removing "AndThrow" (all checkFoo methods throw).


nategbrown9 (approved):
Looks good

PR #278, opened 2020-09-25
Enrollment Detail object

alxdarksage:
Augment the paged enrollment API so that the actors referenced are given a more complete reference object that can be used to show the enrollment in APIs. Otherwise we need to retrieve these through the participants API, in a page of 25 records that could be 25-75 additional requests. Uses the constructor feature of Hibernate to avoid loading all the ancillary tables of an account.


nategbrown9 (approved):
Looks good.

PR #279, opened 2020-09-28
A couple of APIs for migrating enrollments.

alxdarksage:


src/main/java/org/sagebionetworks/bridge/models/studies/EnrollmentMigration.java
```diff
@@ -0,0 +1,107 @@
+package org.sagebionetworks.bridge.models.studies;
+
+import org.joda.time.DateTime;
+
+/**
+ * A (temporary?) model object so supeadmins can migrate user enrollments.
```
nategbrown9: superadmins

nategbrown9 (approved):
Looks good. Just a spelling fix in a comment.

PR #280, opened 2020-10-06
Add a filter to the study enrollments api to include/exclude accounts marked as test users

alxdarksage:
This has to be merged after the main enrollments PR, then this is only about 5 files.


nategbrown9 (approved):
Looks good.

PR #282, opened 2020-10-16
Consolidating study checks and adding exception for worker role callers.

alxdarksage:
Adding worker along with admin as a caller that doesn't have to pass study access checks (as will be the case with developers, researchers, and future roles like organization administrator).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/RequestContext.java
+++ b/src/main/java/org/sagebionetworks/bridge/RequestContext.java
@@ -1,6 +1,5 @@
 package org.sagebionetworks.bridge;
 
-import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.models.ClientInfo.UNKNOWN_CLIENT;
 
 import java.util.List;
@@ -8,6 +7,7 @@ import java.util.Set;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 
 import org.sagebionetworks.bridge.models.ClientInfo;
 import org.sagebionetworks.bridge.models.Metrics;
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/bridge/RequestContext.java
```diff
@@ -45,7 +45,7 @@ public static RequestContext updateFromSession(UserSession session, SponsorServi
         builder.withCallerAppId(session.getAppId());
 
         StudyParticipant participant = session.getParticipant();
-        if (participant.getOrgMembership() != null && !BridgeUtils.isInRole(participant.getRoles(), ADMIN)) {
```
alxdarksage, line 48: We don't want to engage in shenanigans here...if the role prevents the check, then check for the role, don't just make the list of studies an empty list. The empty list checks are going away.

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
```diff
@@ -244,4 +245,54 @@ public void isInRoleExcludedWrongRole() {
...
+    public void isStudyScopedToCallerSucceedsForGlobalUser() {
+        RequestContext.set(new RequestContext.Builder().build());
+        
+        assertTrue( AuthUtils.isStudyScopedToCaller(TEST_STUDY_ID) );
+    }
+
+    @Test
+    public void isStudyScopedToCallerSucceedsForOrgSponsoredStudy() {
+        RequestContext.set(new RequestContext.Builder()
+                .withCallerRoles(ImmutableSet.of(ADMIN))
```
alxdarksage: Remove role.

src/test/java/org/sagebionetworks/bridge/RequestContextTest.java
```diff
@@ -267,12 +267,10 @@ public void updateFromSessionForAdmin() {
         session.setAppId(TEST_APP_ID);
         
         RequestContext retValue = RequestContext.updateFromSession(session, mockSponsorService);
-        assertEquals(retValue.getOrgSponsoredStudies(), ImmutableSet.of());
+        assertEquals(retValue.getOrgSponsoredStudies(), USER_STUDY_IDS);
```
alxdarksage, line 270: Why did you change this?

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
alxdarksage: Removed

PR #283, opened 2020-10-18
Fixing validation of tech domain email addresses, among others.

alxdarksage:
This fixes an issue discovered by one of our external partners when they tried to add a *.tech email address. The updated validation library accounts for the expanded list of top-level domains that now exist.


nategbrown9 (approved):
Looks good.

PR #285, opened 2020-10-21
Add minutes to complete

alxdarksage:
This is information that is very common in every design of shared components/assessments. Given in minutes so times can be aggregated for constructs like sessions.


nategbrown9 (approved):
Looks good.

PR #286, opened 2020-10-21
Allow app to have other oauth clients that authenticate with Synapse.

alxdarksage:
For MTB we propose to create a separate web application that will also sign in using Synapse. Allow the OAuth providers in an App to provide information about those clients so this can be added on a per-app basis.


nategbrown9 (approved):
Looks good.

PR #288, opened 2020-10-30
Clean up of external IDs

alxdarksage:
1) remove the table of external IDs that duplicates the use of external IDs as part of enrollments (this was never really used as expected);

2) remove the ability to associate users to studies or to add enrollments through the participants API, with one exception: when creating an account with only an external ID as an identifier, it needs to be done as part of creating an account and it does enroll the account being created. Otherwise, use the enrollment or membership APIs.

3) Changed external ID APIs to operate for a specific study. APIs under the /v5/studies/{studyId} will replace most of the global APIs working with accounts.


src/main/java/org/sagebionetworks/bridge/dao/ExternalIdDao.java
```diff
@@ -13,46 +10,14 @@
  * or re-assigned to another user if assigned.
  */
 public interface ExternalIdDao {
-
-    String EXTERNAL_ID_GET_RATE = "external.id.get.rate";
```
DwayneJengSage, line 17: Minor: We should also remove this from src/main/resources/BridgeServer2.conf

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateExternalIdDao.java
```diff
@@ -0,0 +1,96 @@
...
+        AccountId accountId = AccountId.forExternalId(extId.getAppId(), extId.getIdentifier());
+        Optional<Account> opt = accountDao.getAccount(accountId);
+        
+        if (opt.isPresent()) {
+            Account account = opt.get();
+            if (BridgeUtils.filterForStudy(account) != null) {
+                Optional<Enrollment> enOpt = account.getEnrollments().stream()
+                        .filter(en -> extId.getIdentifier().equals(en.getExternalId()))
+                        .findFirst();
+                if (enOpt.isPresent()) {
```
DwayneJengSage: In what case would we not find the external ID? Should we add a test case for this?

If it's unusual to not find the external ID, should we log a warning and/or an error?

src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java
```diff
@@ -330,50 +328,32 @@ public void resetPassword(PasswordReset passwordReset) throws BridgeServiceExcep
...
-        // No account and user doesn't want to create it, treat as a 404
-        if (account == null && !createAccount) {
+        
+        Optional<Enrollment> opt = account.getEnrollments().stream()
+                .filter(enrollment -> externalId.equals(enrollment.getExternalId())).findAny();
+        if (!opt.isPresent()) {
             throw new EntityNotFoundException(Account.class);
         }
+        
+        AuthUtils.isStudyScopedToCaller(opt.get().getStudyId());
```
DwayneJengSage: This call returns a boolean, it doesn't throw.

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -161,14 +160,14 @@
...
-    private ExternalIdService externalIdService;
-    @Mock
     private IntentService intentService;
     @Mock
     private OAuthProviderService oauthProviderService;
     @Mock
     private SponsorService sponsorService;
     @Mock
+    private ExternalIdService externalIdService;
+    @Mock
```
DwayneJengSage: This just got removed. Did we accidentally re-add it?

src/test/java/org/sagebionetworks/bridge/services/AuthenticationServiceMockTest.java
```diff
@@ -1008,95 +1007,38 @@ public void resendPhoneVerificationInvalid() throws Exception {
...
-            service.generatePassword(app, EXTERNAL_ID, true);
+            service.generatePassword(app, EXTERNAL_ID);
             fail("Should have thrown an exception");
         } catch(EntityNotFoundException e) {
             // expected exception
         }
-        verify(accountService, never()).getAccount(any());
         verify(participantService, never()).createParticipant(any(), any(), anyBoolean());
         verify(accountService, never()).changePassword(any(), any(), any());
     }
```
DwayneJengSage: I don't think this test case sufficiently distinguishes between "no account" and "account does not have external ID".

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1204,36 +1137,14 @@ public void updateParticipantWithExternalIdValidationAddingId() {
...
-        assertEquals(enrollments.size(), 2);
+        participantService.updateParticipant(APP, PARTICIPANT);
         
-        // get() throws exception if Enrollment not found
-        enrollments.stream().filter((as) -> as.getStudyId().equals("studyA")).findAny().get();
-        enrollments.stream().filter((as) -> as.getStudyId().equals("studyB")).findAny().get();
+        verify(enrollmentService, never()).enroll(any(), any());
     }
     
-    @Test
```
DwayneJengSage, line 1236: We accidentally removed the Test annotation from this test.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -1265,38 +1176,13 @@ public void updateParticipantMustIncludeCallerStudy() {
...
-        Enrollment enB = Enrollment.create(APP.getIdentifier(), "studyB", ID, "extB");
-        account.getEnrollments().add(enB);
-        
-        participantService.updateParticipant(APP, participant);
-        
-        // We've tested this collection more thoroughly in updateParticipantTransfersStudyIdsForAdmins()
-        assertEquals(account.getEnrollments().size(), 1);
-        verify(externalIdService).unassignExternalId(account, "extB");
-        verify(cacheProvider).removeSessionByUserId(ID);
     }
```
DwayneJengSage: These two tests appear to be empty.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ExternalIdControllerV4.java
```diff
@@ -38,34 +42,37 @@ final void setExternalIdService(ExternalIdService externalIdService) {
...
+    public PagedResourceList<ExternalIdentifierInfo> getExternalIdentifiersForStudy(@PathVariable String studyId,
+            @RequestParam(required = false) String offsetBy, @RequestParam(required = false) String pageSize,
+            @RequestParam(required = false) String idFilter) {
+        UserSession session = getAuthenticatedSession(DEVELOPER, RESEARCHER);
+
+        int offsetByInt = BridgeUtils.getIntOrDefault(offsetBy, 0);
+        int pageSizeInt = getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+
+        return externalIdService.getPagedExternalIds(session.getAppId(), studyId, idFilter, offsetByInt, pageSizeInt);
     }
```
DwayneJengSage, line 65: Add unit tests for this API?

src/main/java/org/sagebionetworks/bridge/validators/StudyParticipantValidator.java
```diff
@@ -92,34 +92,30 @@ public void validate(Object object, Errors errors) {
                     errors.rejectValue("orgMembership", "cannot be set by caller");
                 }
             }
-            
+            // External IDs can be updated during creation or on update. If it's already assigned to another user, 
+            // the database constraints will prevent this record's persistence.
+            if (isNotBlank(participant.getExternalId()) && participant.getExternalIds().isEmpty()) {
+                errors.rejectValue("externalId", "must now be supplied in the externalIds property that maps a study ID to the new external ID");
+            }
```
DwayneJengSage, line 99: Add unit test for this case?

src/test/java/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -417,19 +379,6 @@ public void subsetOfStudiesOK() {
...
-                .withCallerEnrolledStudies(ImmutableSet.of("studyA", "studyC")).build());
-        try {
-            StudyParticipant participant = withStudies("studyA");
-            
-            validator = makeValidator(true);
-            assertValidatorMessage(validator, participant, "studyIds[studyA]", "is not a study");
-        } finally {
-            RequestContext.set(RequestContext.NULL_INSTANCE);
-        }
-    }
```
DwayneJengSage: Looks like we still have this logic. Should we keep this test?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #289, opened 2020-11-02
Fix revision-related query bug

alxdarksage:
The query was selecting all assessments with an identifier, sorting by revision, then weeding out by appId, which could select and then remove an assessment outside of the target app context. So for example, if you published an assessment to shared, then bumped its version in the origin app, the shared assessment stopped being returned through the API. Fixed this so the sub select also limits itself to the target app context.


nategbrown9 (approved):
Looks good.

PR #291, opened 2020-11-04
Refactor to make it easier to add roles (less code change).

alxdarksage:
We're going to add another role (ORG_ADMIN) and there are some calls where any administrative user should succeed. Just refactoring so we don't have to examine all the code every time we add a new role that would be included in these checks.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/Roles.java
+++ b/src/main/java/org/sagebionetworks/bridge/Roles.java
@@ -14,14 +14,16 @@ public enum Roles {
     SUPERADMIN;
     
     /**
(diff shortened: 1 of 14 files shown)


nategbrown9 (approved):
Looks good. Only comment is the overloading of the term admin. The ADMINISTRATIVE_ROLES set contains a role of ADMIN, and getAdminSession applies to DEVELOPER and RESEARCHER roles as well as ADMIN. I don't however have a better term to use.

alxdarksage:
It may be worth the longer name `getAdministrativeSession` to differentiate. I'll change wording.

PR #292, opened 2020-11-06
Refactor status

alxdarksage:
Changing status to reflect the fact there are multiple pathways to authenticate. If the account has at least one valid pathway it's enabled, otherwise it is unverified. Disabled overrides this and sets the account as disabled.


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -354,12 +359,24 @@ public void setRoles(Set<Roles> roles) {
...
+        }
+        // As long as there is a pathway to authenticate, the account is considered verified.
+        // Can sign in with an external ID and password
+        boolean verExtId = (passwordHash != null) && !collectExternalIds(this).isEmpty();
+        // Can sign in via email
+        boolean verEmail = (email != null && emailVerified == TRUE);
+        // Can sign in with phone number
+        boolean verPhone = (phone != null && phoneVerified == TRUE);
+        // By setting the field, it's persisted on database updates
+        this.status = (verEmail || verPhone || synapseUserId != null || verExtId) ? ENABLED : UNVERIFIED;
```
DwayneJengSage, line 379: Do we have unit tests for this?

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -274,6 +264,17 @@ public Account getAccount(AccountId accountId) {
...
+    /**
+     * This is used when enrolling a user, since the account itself is not yet in a study
+     * that is visible to the caller. There may be similar cases where study access 
+     * permissions need to be bypassed.
+     */
+    public Optional<Account> getAccountNoPermissionCheck(AccountId accountId) {
+        checkNotNull(accountId);
+        
+        return accountDao.getAccount(accountId);
+    }
```
DwayneJengSage: Do we have unit tests for this?

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -320,7 +321,7 @@ protected Account authenticateInternal(App app, Account account, SignIn signIn)
         // Auth successful, you can now leak further information about the account through other exceptions.
         // For email/phone sign ins, the specific credential must have been verified (unless we've disabled
         // email verification for older apps that didn't have full external ID support).
-        if (account.getStatus() == UNVERIFIED) {
+        if (account.getStatus() == UNVERIFIED && app.isEmailVerificationEnabled()) {
```
DwayneJengSage, line 324: Do we have a test case for an unverified account but emailVerificationEnabled is false?

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -96,12 +97,17 @@ public Enrollment enroll(Enrollment enrollment) {
 
         // verify this has appId and accountId
         Validate.entityThrowingException(INSTANCE, enrollment);
+        
+        // Verify that the caller has access to this study
+        AuthUtils.checkStudyScopedToCaller(enrollment.getStudyId());
```
DwayneJengSage: Unit test for caller not in study?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
alxdarksage: Added tests.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: added a case for authenticate and for reauthenticate

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: This is tested by `enrollNotAuthorizedAsStudyResearcher()`

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: I also changed the name to be a little clearer that the issue is filtering...if it were just a check-and-throw we could add "self" to the list of permissions, but the filtering is not so easily defeated.

alxdarksage:
This PR has been updated and is ready for a second review.

DwayneJengSage (approved):

DwayneJengSage (approved):

PR #293, opened 2020-11-11
Remove lambda

alxdarksage:
We no longer need code for a compensatory transaction when working with accounts, so this code is no longer needed. Removed.


nategbrown9 (approved):
Looks good.

PR #294, opened 2020-11-14
Rename *MockTest* classes, and combine four test files into two (removing a few tests that were redundant between the files).

alxdarksage:
Early on we had unit and integration tests in this project, and so some tests were named "*MockTest.java" to differentiate them, but they are all unit tests now. In removing the "Mock" part of the name, identified two cases where there were two test classes testing the same thing, with a few overlapping tests. Cleaned this up too.


nategbrown9 (approved):
Looks good.

PR #295, opened 2020-11-17
Return quietly when sign up contains an external ID and the account already exists.

alxdarksage:
Some apps do the following steps:

1) external ID without an account is created via the Bridge Study Manager;
2) app calls sign up with the external ID;
3) app calls sign in.

You can no longer separate an external ID from an account, so step #1 is creating the account. This fix checks for that, and continues to return 200 for step #2 so that existing apps can proceed smoothly to step #3. In the future, apps can skip step #1 by changing the JSON they submit for sign up (`"externalId": "id" --> "externalIds": {"studyId": "id"}`).


nategbrown9 (approved):
Looks good.

PR #297, opened 2020-11-19
Refactor authorization utils for greater reuse and clarity.

alxdarksage:
- clean up AuthUtils (declarative behavior with AuthEvaluator class; updating documentation);
- rename utility methods to balance length of utility names with comprehensibility;
- eliminated cases where nulls were being passed into AuthUtils checks (these should have named evaluators so we can more easily identify exceptional security rules)
- additional checks against the request context required setting the context up further in some tests.

Note that AuthUtils checks could be moved out of ParticipantController and ParticipantReportController (after discussions with Dwayne about how we'll handle this), *but* they may need to stay there since new APIs will be introduced to work with participants in the scope of a specific study.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
@@ -1,10 +1,10 @@
 package org.sagebionetworks.bridge;
 
-import static org.sagebionetworks.bridge.BridgeConstants.CALLER_NOT_MEMBER_ERROR;
 import static org.sagebionetworks.bridge.RequestContext.NULL_INSTANCE;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.DEVELOPER;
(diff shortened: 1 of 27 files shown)


src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
```diff
@@ -74,106 +76,96 @@ public void checkSelfStudyResearcherOrAdminFailsForDev() {
...
+        AuthUtils.checkSelfOrStudyResearcher(USER_ID, TEST_STUDY_ID);
     }
 
-    @Test
-    public void checkOrgMembershipSucceedsForAdmin() {
-        RequestContext.set(new RequestContext.Builder()
-                .withCallerRoles(ImmutableSet.of(ADMIN)).build());
-        
-        AuthUtils.checkOrgMembership(TEST_ORG_ID);
-    }
```
alxdarksage, line 86: Restored. It still passes for admins.

nategbrown9 (approved):
Looks good

PR #298, opened 2020-11-20
Admins can create accounts with only external ID

DwayneJengSage:
Work in progress. Has not been fully tested.


alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -120,9 +120,10 @@ public UserSession createUser(App app, StudyParticipant participant, Subpopulati
         checkNotNull(app, "App cannot be null");
         checkNotNull(participant, "Participant cannot be null");
         
-        // Validate app + email or phone. This is the minimum we need to create a functional account.
+        // Validate app + email or phone or external ID. This is the minimum we need to create a functional account.
         SignIn signIn = new SignIn.Builder().withAppId(app.getIdentifier()).withEmail(participant.getEmail())
-                .withPhone(participant.getPhone()).withPassword(participant.getPassword()).build();
+                .withPhone(participant.getPhone()).withExternalId(participant.getExternalId())
+                .withPassword(participant.getPassword()).build();
```
alxdarksage, line 126: Don't we want to pass .withExternalIds() here and not withExternalId()?

src/main/java/org/sagebionetworks/bridge/services/UserAdminService.java
```diff
@@ -120,9 +120,10 @@ public UserSession createUser(App app, StudyParticipant participant, Subpopulati
...
         Validate.entityThrowingException(SignInValidator.MINIMAL, signIn);
         
         IdentifierHolder identifier = null;
```
alxdarksage, line 129: Should we add a test?

PR #299, opened 2020-11-20
Organization + assessment

kss39:


src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -162,6 +169,12 @@ public void deleteOrganization(String appId, String identifier) {
         
         Organization existing = orgDao.getOrganization(appId, identifier)
                 .orElseThrow(() -> new EntityNotFoundException(Organization.class));        
+        if (assessmentDao.hasAssessmentFromOrg(appId, identifier)) {
+            throw new ConstraintViolationException.Builder().withMessage(
+                    "Cannot delete organization: it is referenced by an assessment that is still accessible")
```
alxdarksage: How about: "Cannot delete organization (it still owns one ore more assessments)."

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -162,6 +169,12 @@ public void deleteOrganization(String appId, String identifier) {
...
+                    .withEntityKey("appId", appId)
+                    .withEntityKey("organization", identifier).build();
```
alxdarksage: How about "orgId"? We seem to be standardizing on that form for IDs.

alxdarksage (approved):
Some minor comments about wording and variable name, otherwise approved.

PR #300, opened 2020-11-25
Organization admin APIs

alxdarksage:
Introduces a new role to manage organizations and their administrative accounts. 

https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/886440163/Bridge+v2+Implementation+Design#Organization-Administrators

https://www.pivotaltracker.com/story/show/174481392


src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountTest.java
```diff
@@ -18,24 +22,93 @@
...
 import com.google.common.collect.Sets;
 
 public class HibernateAccountTest {
+    
+    // We're only concerned with serializing this model...the StudyParticipant can
+    // be used to deserialize an Account model sent from our SDK (it's a superset
+    // of an account record).
+    @Test
+    public void canSerialize() throws Exception {
+        Account account = new HibernateAccount();
```
DwayneJengSage, line 75: Should add JsonIgnored attributes to verify that they are being nulled out when we serialize to Json. (Including getAllConsentSignatureHistories from the interface.)

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantControllerTest.java
```diff
@@ -1287,77 +1262,6 @@ public void getActivityHistoryV3DefaultsToNulls() throws Exception {
         assertNull(endTimeCaptor.getValue());
     }
 
-    @Test
-    public void updateIdentifiersWithPhone() throws Exception {
```
DwayneJengSage: Was this API deleted from ParticipantController? Did we forget to include ParticipantController in the PR?

DwayneJengSage (changes requested):
Overall looks good. A few comments about Account serialization tests and missing files.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountTest.java
alxdarksage: Added this, removed reauthToken (I don't know why we would return it outside of a session), and added consent records to verify they are not there in the serialization as well.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantControllerTest.java
alxdarksage: I've restored all these tests. At one point I thought this could be moved entirely to the Accounts controller, but I'm not 100% sure this is never used by clients to add email/phone, or never would be, so I ended up leaving it here as well.

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -200,18 +201,31 @@ public void deleteOrganization(String appId, String identifier) {
...
+            .withAdminOnly(true)
+            .withOrgMembership("<none>").build();
+
+        return accountDao.getPagedAccountSummaries(appId, scopedSearch);
+    }
+    
+    /**
+     * Note that we currently allow organization admins to re-assign people from 
+     * other organizations, which we might want to change.
+     */
```
DwayneJengSage: This was fine previously, since it was only open to admins, and iirc admins had permissions over the entire app. Now that addMember is available to org admins, this would allow one bad acting organization to steal the personal Info for participants not associated with their organization. I strongly recommend restricting this.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountTest.java
```diff
@@ -4,38 +4,140 @@
...
+        assertNull(node.get("reauthToken"));
+        assertNull(node.get("timeZone"));
+        assertNull(node.get("sharingScope"));
+        assertNull(node.get("notifyByEmail"));
+        assertNull(node.get("migrationVersion"));
+        assertNull(node.get("enrollments"));
+        assertNull(node.get("reauthToken"));
+        assertNull(node.get("activeConsentSignature"));
+        assertNull(node.get("consentSignatureHistory"));
+        assertNull(node.get("allConsentSignatureHistories"));
```
DwayneJengSage, line 154: We should add these attributes to the Accounts object so we can distinguish between the field not being present and the field being ignored by Jackson.

Also, we're missing getActiveEnrollments.

DwayneJengSage (changes requested):
Follow-up comment on unit test, and a potential security issue on addMember() that I just noticed.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountTest.java
alxdarksage: Added these fields. Verified that removing the @JsonIgnore annotation from the interface causes the test to fail (the field is output in the JSON).

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
alxdarksage: I've added a check in the service: if the account is already assigned and you're not an admin, it'll throw a bad request exception, preventing it. (I think we need to allow admins to fix bad assignments if they ever happen.)

DwayneJengSage (approved):

PR #304, opened 2020-12-04
Adding git secrets to prevent accidental commits

kss39:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+-   repo: https://github.com/awslabs/git-secrets
+    rev: 80230afa8c8bdeac766a0fece36f95ffaa0be778
```
zaro0508: i believe you only need first 7 numbers of the SHA?  Also wondering why you don't set it to a released tag?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
...
+    hooks:
+        - id: git-secrets
```
zaro0508: hmm, does this work?  don't we need to run the git-secrets script like the one in https://github.com/awslabs/git-secrets/blob/master/.pre-commit-hooks.yaml#L4 ?

zaro0508 (changes requested):

.pre-commit-config.yaml
kss39: Hi,

Actually, the most recent released tag (1.3.0) does not support the pre_commit_hook structure. It was introduced after this release (this commit: b4a3208)

.pre-commit-config.yaml
zaro0508: just verified myself.  this, as-is, does indeed work.

zaro0508 (changes requested):
since running pre-commit locally (on your own dev machine) is voluntary i would advise you to also make this run as part of the CI process.  That means you should setup pre-commit to run on every PR and merge.  You can find examples of this setup just about all of the `Sage-Bionetworks/*-infra` repos that have a .pre-commit-config.yaml file.

cascadianblue:
Have you considered adding any other pre-commit hooks (i.e. linters, etc.)? I'm betting that a lot of committers will forget to run/install pre-commit locally (especially if developers work across multiple environments) even if you add it to your documentation. Adding more other hooks might make it more likely for committers to accidentally trip up the CI-side pre-commit check (e.g. with forgetting a newline at the end of a Python file) and force them to install pre-commit locally. Make sense?

.pre-commit-config.yaml
boyleconnor: Have you considered adding any other pre-commit hooks (i.e. linters, etc.)? I'm betting that a lot of committers will forget to run/install pre-commit locally (especially if developers work across multiple environments) even if you add it to your documentation. Adding more other hooks might make it more likely for committers to accidentally trip up the CI-side pre-commit check (e.g. with forgetting a newline at the end of a Python file) and force them to install pre-commit locally. Make sense?

boyleconnor (changes requested):

DwayneJengSage (approved):

alxdarksage:
Please add something to the README about any actions developers need to take to ensure this is doing its job. Anything we need to install or run?

.pre-commit-config.yaml
kss39: Thank you, yes, I will add more checks for it.

kss39:
> Please add something to the README about any actions developers need to take to ensure this is doing its job. Anything we need to install or run?

Yes, as far as I tested, all we need to do is `pip install pre-commit` and
`pre-commit install` in BridgeServer2/ folder.

As for where to put the README, how about we put it on the wiki for developer bootstrapping? https://github.com/Sage-Bionetworks/BridgeServer2/wiki/bootstrap

I double checked, this pre-commit checks also works with IDE GUIs, say that if I committed an AWS key using IntelliJ's GUI commit window (instead of command line), IntelliJ will fail the commit process and raise an error.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
+repos:
+-   repo: https://github.com/awslabs/git-secrets
+    rev: 80230af
+    hooks:
+        -   id: git-secrets
+-   repo: https://github.com/pre-commit/pre-commit-hooks
```
boyleconnor, line 6: Might be good to include some general-purpose Java linting hooks, e.g. one of [these](https://github.com/gherynos/pre-commit-java).

boyleconnor (approved):

zaro0508 (approved):

PR #305, opened 2020-12-08
Add new API to update mPower @home events

alxdarksage:
We want to add some custom events for some users, which seems like a reasonable API for a researcher (in fact we'll need it for MTB as well).


nategbrown9 (approved):
Looks good.

PR #308, opened 2020-12-11
Security refactor

alxdarksage:
Remove the static method wrappers around the evaluators so it's easier to track what is being enforced, and where.


nategbrown9 (approved):
Looks good. My only question is if the repeated strings such as "orgId" and "ownerId" should be pulled out into constants.

alxdarksage:
Changed these arguments to be an enum and checked they aren't null.

PR #309, opened 2020-12-18
Study coordinator

alxdarksage:
This check-in adds a new role, STUDY_COORDINATOR, which an associated set of APIs. These are comparable to the researcher role and the /v3/participants APIs, but they are scoped using the new organization associations (study coordinators can only see and edit participants who are in studies they have access to through their organization).

Some additions were made to the account code to scope account searches to enrolled members of a specific study.

Changes were made to security checks now that there is a specific role for scoped researchers, rather than checking for the researcher role + study access. Some APIs may need to be adjusted to allow study coordinators to work (e.g. until we have a new consent system, they need to be able to read subpopulations).

Enrollment migration code was removed.

JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-2899
Discussion of rationale for this approach: https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/1641250824/Bridge+Authorization+Rules

Study coordinator is mentioned in the implementation guide (briefly): https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/886440163/Bridge+v2+Implementation+Design#Study-Coordinators


DwayneJengSage:
This will open up some APIs from study researchers to all researchers in the app. Is this intended?

alxdarksage:
> This will open up some APIs from study researchers to all researchers in the app. Is this intended?

Yes, these people will now be study coordinators. Since study coordinator is now a researcher scoped to specific studies, the researcher role no longer needs to be scoped to specific studies (it's also less useful, although Sage users might use it). A lot of the checks in the AuthUtils class that are exempting callers with no associated studies (callerConsideredGlobal) are because we frequently create researcher role accounts with no study associations...these are usually Sage Bionetworks users.

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -169,11 +185,27 @@ public void createStudy() {
         assertFalse(persisted.isDeleted());
         assertNotEquals(persisted.getCreatedOn(), timestamp);
         assertNotEquals(persisted.getModifiedOn(), timestamp);
+        
+        verify(sponsorService).addStudySponsor(TEST_APP_ID, "oneId", TEST_ORG_ID);
+    }
+    
+    @Test
+    public void createStudyWithoutSponsorship() {
```
DwayneJengSage, line 193: Should we add the test case for flag set to false and orgSponsorship is specified?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
```diff
@@ -222,21 +227,29 @@ public StatusMessage signOut(@PathVariable String userId,
...
-        if (!IS_ORGADMIN.check(ORG_ID, orgId)) {
-            throw new EntityNotFoundException(Account.class);
+    /**
+     * This kind of code is driving me nuts... I'm not sure how to rationalize it further. We need
+     * to retrieve the account and test its organization membership. 
+     */
+    public Account verifyOrgAdminIsActingOnOrgMember(String appId, String callerOrgId, String userIdToken) {
+        // The caller needs to be associated to an organization
+        if (callerOrgId == null) {
+            throw new UnauthorizedException();
```
DwayneJengSage: What about admins?

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantControllerTest.java
```diff
@@ -0,0 +1,833 @@
...
+        StudyParticipant participant = new StudyParticipant.Builder()
+                .withHealthCode("healthCode").build();
+        when(mockParticipantService.getParticipant(app, "healthcode:"+TEST_USER_ID, true)).thenReturn(participant);
+        
+        Enrollment en = Enrollment.create(TEST_APP_ID, TEST_STUDY_ID, "healthcode:"+TEST_USER_ID);
+        List<EnrollmentDetail> list = ImmutableList.of(new EnrollmentDetail(en, null, null, null));
+        when(mockEnrollmentService.getEnrollmentsForUser(TEST_APP_ID, TEST_STUDY_ID, "healthcode:"+TEST_USER_ID))
+            .thenReturn(list);
+
+        controller.getParticipant(TEST_STUDY_ID, "healthcode:"+TEST_USER_ID, true);
```
DwayneJengSage, line 365: What about cases where the Admin can lookup by healthcode, or where coords can lookup by healthcode if healthcode export is enabled?

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantControllerTest.java
```diff
@@ -0,0 +1,833 @@
...
+        
+        mockAccountInStudy();
+
+        String retValue = controller.getParticipant(TEST_STUDY_ID, TEST_USER_ID, true);
+        StudyParticipant deser = BridgeObjectMapper.get().readValue(retValue, StudyParticipant.class);
+        assertEquals(deser.getHealthCode(), "healthCode");
+    }
+
+    @Test
+    public void getParticipantIncludesHealthCodeIfConfigured() throws Exception {
```
DwayneJengSage, line 304: What about the case where we filter out healthcode if not admin and not enabled?

src/main/resources/db/changelog/changelog.sql
```diff
@@ -423,3 +423,8 @@ ADD CONSTRAINT `unique_extId` UNIQUE (studyId, externalId);
 
 ALTER TABLE `AccountRoles`
 MODIFY COLUMN `role` enum('DEVELOPER','RESEARCHER','ADMIN','ORG_ADMIN','WORKER','SUPERADMIN') NOT NULL;
+
+-- changeset bridge:23
+
+ALTER TABLE `AccountRoles`
+MODIFY COLUMN `role` enum('DEVELOPER','RESEARCHER','ADMIN','ORG_ADMIN','STUDY_COORDINATOR','WORKER','SUPERADMIN') NOT NULL;
```
DwayneJengSage: Inserting a new enum value in the middle of the list can have pretty major performance implications https://stackoverflow.com/questions/9638801/changing-an-enum-type-in-mysql

It would be better to always add enum values to the end of the list.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
alxdarksage: Added (it should not add a sponsoring org; it doesn't).

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
alxdarksage: Because this is an API for admin accounts, I'm assuming admins need to create the account in an organization. The account could be picked up through the unassigned API, but it's better if the account is immediately in an organization. If we accounts in multiple organizations or teams, this might need to be rethought.

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantControllerTest.java
alxdarksage: Added test of admin can look up by health code even if disabled, and study coords can lookup if enabled.

src/test/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantControllerTest.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
DwayneJengSage: Admins and superadmins will no longer have access to these APIs. This could be problematic if admins and superadmins need access these APIs for administrative or debugging purposes.

src/main/resources/db/changelog/changelog.sql
alxdarksage: I didn't know this...I'll do this from now on (changed).

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
alxdarksage: Admins should be in an organization (despite the fact that they can access everything in an app, all administrative accounts would/should be associated to an organization), but super admins do not have to be (indeed they can't because the organizations are different between apps). Okay I will add an exception here for super admins.

alxdarksage: No never mind, admins should *also* be able to retrieve accounts outside of their organization, adding that as well.

DwayneJengSage (approved):

PR #311, opened 2021-01-18
Only show studies that the caller should see.

alxdarksage:
Studies are the first of a number of objects that study-scoped callers should have filtered to only those that are owned by their organization. In this case we get the information directly from the list of organization-sponsored studies.


nategbrown9 (approved):
Looks good.

PR #313, opened 2021-01-21
Non time series participant merge into develop

kllysng:
This PR is mainly a sanity check - it will give you the opportunity to check my progress and make sure that I'm on the right track. Thank you!


src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
```diff
@@ -50,6 +50,7 @@
 import org.hibernate.engine.spi.SessionFactoryImplementor;
 import org.hibernate.event.service.spi.EventListenerRegistry;
 import org.hibernate.service.spi.ServiceRegistryImplementor;
+import org.sagebionetworks.bridge.dynamodb.*;
```
alxdarksage: I would avoid this. Both Eclipse and IntelliJ will just expand this again, causing a lot of churn in PRs.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+     * @return
+     */
+    ForwardCursorPagedResourceList<ParticipantData> getParticipantDataV4(String userId, String offsetKey, int pageSize);
+
+    //TODO check the functions below..
+
+    /**
+     * Get participant data record for the given userId and configId
+     * @param userId
+     * @param configId
```
alxdarksage: configId for something named data? What about just identifier, or key?

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+    ForwardCursorPagedResourceList<? extends ParticipantData> getParticipantData(String userId, String offsetKey, int pageSize);
+
+    /**
+     * Get participant data for the given userId
+     * @param userId
+     * @param offsetKey
+     * @param pageSize
+     * @return
+     */
+    ForwardCursorPagedResourceList<ParticipantData> getParticipantDataV4(String userId, String offsetKey, int pageSize);
```
alxdarksage: I don't think you'll need two APIs, one with a "V4" postfix. You are creating the API so there's nothing to support or adapt.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param pageSize
+     * @return
+     */
+    ParticipantData getParticipantDataRecordV4(String userId, String configId, String offsetKey, int pageSize);
+
+    /**
+     * Writes a participant data to the backing store.
+     * @param data - String data
+     */
+    void saveParticipantData(ParticipantData data); //TODO: how do we know the userId and configId of this data?
```
alxdarksage: Usually we add appId and userId in every method call (because we need both to load the user in a way that confirms this user is in the app context that the caller is signed into).

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Delete all records
+     * @param userId
+     *          userId to delete
+     */
+    void deleteParticipantData(String userId);
```
alxdarksage: deleteAllParticipantData?

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Delete a single participant data record.
+     * @param userId
+     * @param configId
+     */
+    void deleteParticipantDataRecord(String userId, String configId);
```
alxdarksage: deleteParticipantData (we don't have, so far as I can tell, a series of data under a key...so you probably don't need the "record" concept here. Data = one database record.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.Objects;
+
+/**
+ *
+ */
+@DynamoDBTable(tableName = "ParticipantData")
+public class DynamoParticipantData implements ParticipantData {
+
+    private String userId;
```
alxdarksage, line 14: Purely for reasons of convention, I would use healthCode as the user identifier here, not userId.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,93 @@
...
+    private String configId;
+    private String data;
+    private Long version;
+
+    //TODO: figure out which methods require a @JsonIgnore annotation
+    //TODO: figure out which methods require a @DynamoDBIgnore annotation
```
alxdarksage: Only the healthCode hash key column needs to be suppressed. I don't think there's any reason to ever expose it, so @JsonIgnore would work.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+        DynamoDBQueryExpression<DynamoParticipantData> query =
+                new DynamoDBQueryExpression<DynamoParticipantData>().withHashKeyValues(hashKey);
+        List<DynamoParticipantData> results = mapper.query(DynamoParticipantData.class, query);
+
+        return new ForwardCursorPagedResourceList<DynamoParticipantData>(results, offsetKey);
+    }
+
+    @Override
+    public ForwardCursorPagedResourceList<ParticipantData> getParticipantDataV4(final String userId, final String offsetKey, final int pageSize) {
+        //TODO: why do we use final for these params but not the others?
```
alxdarksage: No reason. We don't consistently use final for this purpose; so I would lean against using it purely for convention.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+        checkNotNull(userId);
+
+        int pageSizeWithIndicatorRecord = pageSize + 1; //TODO: what is indicatorRecord?
```
alxdarksage: Let's say you retrieve 10 records. You want to return the ID of the 11th record to indicate where the next page should begin, so we get page + 1 records to get that 11th record ID.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+    final void setParticipantStateMapper(DynamoDBMapper participantStateMapper) {
+        this.mapper = participantStateMapper;
+    }
+
+    @Override
+    public ForwardCursorPagedResourceList<? extends ParticipantData> getParticipantData(String userId, String offsetKey, int pageSize) {
+        checkNotNull(userId);
+
+        DynamoParticipantData hashKey = new DynamoParticipantData();
+        hashKey.setUserId(userId);
```
alxdarksage: If the offsetKey exists, it's going to be the identifier and you'll want to set that in the key.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+        PaginatedQueryList<DynamoParticipantData> resultPage = mapper.query(DynamoParticipantData.class, query);
+
+        List<ParticipantData> list = Lists.newArrayListWithCapacity(pageSizeWithIndicatorRecord);
+        for (ParticipantData pariticipantData : resultPage) {
+            list.add(pariticipantData);
+        }
+
+        String nextPageOffsetKey = null;
+        if (list.size() == pageSizeWithIndicatorRecord) {
+            nextPageOffsetKey = Iterables.getLast(list).getConfigId(); // TODO: is this right? We want to get the range key essentially
```
alxdarksage: Yes, is it the range key that serves as a pointer to where the next page of records should begin. DDB only pages forward in this manner.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+        hashkey.setUserId(userId);
+
+        DynamoDBQueryExpression<DynamoParticipantData> query =
+                new DynamoDBQueryExpression<DynamoParticipantData>().withHashKeyValues(hashkey);
+        List<DynamoParticipantData> objectsToDelete = mapper.query(DynamoParticipantData.class, query);
+
+        if (!objectsToDelete.isEmpty()) {
+            List<DynamoDBMapper.FailedBatch> failures = mapper.batchDelete(objectsToDelete);
+            //TODO: in DynamoReportDataDao, this is FailedBatch only, not DynamoDBMapper.FailedBatch. Is that an issuE?
+            BridgeUtils.ifFailuresThrowException(failures);
```
alxdarksage, line 92: I think it's just a difference in imports...? If so it doesn't matter.

src/main/java/org/sagebionetworks/bridge/models/ParticipantData.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.sagebionetworks.bridge.dynamodb.DynamoParticipantData;
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+
+@BridgeTypeName("ParticipantData")
+@JsonDeserialize(as= DynamoParticipantData.class)
+public interface ParticipantData extends BridgeEntity{
+
+    // TODO: I'm not sure what this is for but it's in ReportData
+    static TypeReference<ForwardCursorPagedResourceList<ParticipantData>> PAGED_REPORT_DATA = new TypeReference<ForwardCursorPagedResourceList<ParticipantData>>() {
```
alxdarksage: It's a type reference to convert JSON into the ForwardCursorPagedResourceList<ParticipantData> object. I think you'd only do this for tests, it's not something we receive through the API.

src/main/java/org/sagebionetworks/bridge/models/ParticipantData.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    String getUserId();
+    void setUserId(String userId);
+
+    String getConfigId();
+    void setConfigId(String configId);
+
+    String getData();
+    void setData(String data);
```
alxdarksage: Typically if we're storing JSON, we'd store this as a JsonNode and use type conversion annotations to store it (and I would check because DDB might just handle the conversation to a string automatically for us). We definitely don't want this double escaped so it is returned to the client as a string of JSON...it should just be JSON.

src/main/java/org/sagebionetworks/bridge/services/ParticipantDataService.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
+
+@Component
+public class ParticipantDataService {
+    private ParticipantDataDao participantDataDao;
```
alxdarksage, line 21: Some things to check in this class:
1) is the user ID real and does the caller have access to it (you may do this in the controller however)
2) what if the identifier does not exist? (throw EntityNotFoundException)
3) what if pageSize is not provided?
4) what happens if someone saves data that has been saved already, leading to a version field mismatch? Do we throw a Bridge exception when this happens?

src/main/java/org/sagebionetworks/bridge/services/ParticipantDataService.java
```diff
@@ -0,0 +1,75 @@
...
+        participantDataDao.deleteParticipantData(userId);
+    }
+
+    public void deleteParticipantDataRecord(String userId, String configId) {
+        participantDataDao.deleteParticipantDataRecord(userId, configId);
+    }
+
+    //TODO: do we need a dedicated update function?
+
+    //TODO: organize imports once more finalized
```
alxdarksage: Your IDE will automatically order imports. The value of allowing your IDE to do it is that they are fairly standardized, so other developers' ordering will not conflict with yours.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,70 @@
...
+public class ParticipantDataController extends BaseController {
+
+    private ParticipantDataService participantDataService;
+
+    @Autowired
+    final void setReportService(ParticipantDataService participantDataService) {
+        this.participantDataService = participantDataService;
+    }
+
+    @GetMapping("/v4/users/self/configs")
```
alxdarksage: Appears that this can just be under /v3/users like other current endpoints?

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,70 @@
...
+        UserSession session = getAuthenticatedAndConsentedSession();
+        //TODO: using authenticated AND consented session bc we don't want the user to write any data unless they're consented right?
+
+        ParticipantData participantData = parseJson(ParticipantData.class);
+        participantData.setUserId(null); // TODO: ask about "set in service, but just so no future use depends on it", ParticipantReportController: line 97
+
+        participantDataService.saveParticipantData(session.getId(), identifier, participantData);
+
+        return new StatusMessage("Participant data saved.");
+    }
```
alxdarksage, line 159: POST to /v4/users/self/configs in order to create a new record, return 201 and it's an error if the identifier already exists
POST to "/v4/users/self/configs/{identifier} to update a record, return 200 and it's an error if the identifier does NOT already exist.

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
kllysng: Thank you! I didn't even know that I did this and will fix it now. Also thank you so much for all these comments!! I'm going through and addressing them all now.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
kllysng: Do you mean that I should change configId to identifier? This would make sense because I think this was originally ParticipantConfig, but we changed it to ParticipantData and I didn't change configId. Oops, I'll go ahead and change it to identifier.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
kllysng: I will change it! Are there any examples I should look at to help me understand the usage or naming convention of healthCode?

src/main/java/org/sagebionetworks/bridge/config/SpringConfig.java
DwayneJengSage: Additional clarification: There's a project setting in IntelliJ to automatically collapse imports with *. It defaults to something low like 3. I bumped mine up to 999, which basically means it'll never collapse them.

The import sort order we decided on is (iirc)
* import static
* (blank line)
* import java.*
* import javax.*
* (blank line)
* import everything else
* (blank line)
* import org.sagebionetworks.bridge.*

I can send you my code style config, so hopefully your formatting will be the same as ours.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
DwayneJengSage: Actually, with query, it's a bit simpler than that. We can use query.setExclusiveStartKey() and pass in the offsetKey. The offsetKey would just be the last dataId in the result set.

DwayneJengSage: EDIT: I mispoke. exclusiveStartKey only works with queryPage, and it's a complex opaque DynamoDB object that we may or may not want to expose to API callers.

It would be better to use a rangeKeyCondition. For example:
```
        Condition rangeKeyCondition = new Condition().withComparisonOperator(ComparisonOperator.GE)
                .withAttributeValueList(new AttributeValue().withS(dataId));
        query.withRangeKeyCondition("dataId", rangeKeyCondition);
```

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Get participant data record for the given userId and configId
+     * @param userId
+     * @param configId
+     * @param offsetKey
+     * @param pageSize
+     * @return
+     */
+    ParticipantData getParticipantDataRecordV4(String userId, String configId, String offsetKey, int pageSize);
```
DwayneJengSage: You don't need offsetKey or pageSize when getting a single record.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,93 @@
...
+        return this.data;
+    }
+
+    @Override
+    public void setData(String data) {
+        this.data = data;
+    }
+
+    @Override
+    public Long getVersion() {
```
DwayneJengSage: Need the DynamoDBVersionAttribute annotation.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,130 @@
...
+        DynamoDBQueryExpression<DynamoParticipantData> query = new DynamoDBQueryExpression<DynamoParticipantData>()
+                .withHashKeyValues(hashKey)
+                .withLimit(pageSizeWithIndicatorRecord);
+
+        PaginatedQueryList<DynamoParticipantData> resultPage = mapper.query(DynamoParticipantData.class, query);
+
+        List<ParticipantData> list = Lists.newArrayListWithCapacity(pageSizeWithIndicatorRecord);
+        for (ParticipantData pariticipantData : resultPage) {
+            list.add(pariticipantData);
+        }
```
DwayneJengSage: This can be replaced with

```
List<ParticipantData> list = ImmutableList.copyOf(resultPage);
```

The reason we need to do this is because of Java generics and typing reasons. (I can go into more detail offline if you're curious.)

src/main/java/org/sagebionetworks/bridge/models/ParticipantData.java
DwayneJengSage: It's not actually used in ReportData either. You probably don't need it here.

src/main/java/org/sagebionetworks/bridge/services/ParticipantDataService.java
```diff
@@ -0,0 +1,75 @@
...
+
+        participantData.setUserId(userId);
+        participantData.setData(configId);
+
+        //TODO: do I need a ParticipantDataValidator? I think so..
+
+        participantDataDao.saveParticipantData(participantData);
+    }
+
+    public void deleteParticipantData(String userId) { //TODO: why isn't there a checkNotNull() in the ReportService code?
```
DwayneJengSage: Probably just an oversight. We should be putting checkNotNull() checks for anything that's not user input. (And doing validation for anything that is.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Autowired
+    final void setReportService(ParticipantDataService participantDataService) {
+        this.participantDataService = participantDataService;
+    }
+
+    @GetMapping("/v4/users/self/configs")
+    public ForwardCursorPagedResourceList<String> getAllDataForUser(@RequestParam(required = false) String offsetKey,
+                                                                    @RequestParam(required = false) String pageSize) {
+        UserSession session = getAuthenticatedSession();
```
DwayneJengSage: For consistency, let's also use getAuthenticatedAndConsentedSession() for all the getters. They can't write participant data unless they're consented, and calling the get methods before they do will always return empty results or 404 Not Found.

DwayneJengSage (changes requested):
Structually, the code is fine. Just a lot of nitty gritty details.

The biggest missing piece is input validation in ParticipantDataService.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
DwayneJengSage: We should probably ask Shannon and/or Nathaniel for clarification, but does ParticipantData need a separate create vs update API? The apps don't really care about the distinction between create and update. From their point of view, they just want to write a particular piece of data for a particular identifier.

The existing Participant Report API also doesn't have different APIs for create or update either.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
DwayneJengSage: Clarifying question: Just for the ParticipantData object or for the entire API? The API should definitely use userId, since it's a pain in the neck for callers (such as the workers) to use healthCode.

Using the healthCode in DynamoDB means the service would need to make an extra database call to fetch the account and get the healthCode. Which might be fine, but it's an extra step that we would need to take. (Or it might not be fine. Evidence suggests that Bridge Server has been getting slower in the last 6 months or so.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,63 @@
...
+public class ParticipantDataController extends BaseController {
+
+    private ParticipantDataService participantDataService;
+
+    @Autowired
+    final void setReportService(ParticipantDataService participantDataService) {
+        this.participantDataService = participantDataService;
+    }
+
+    @GetMapping("/v3/users/self/configs")
```
DwayneJengSage: Since we're calling this ParticipantData, we sould change the url from /configs to /data (here and in the rest of the URLs).

kllysng: Oops duh! Thanks, I’ll change it.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,45 @@
...
+public interface ParticipantDataDao {
+
+    /**
+     *
+     * @param userId
+     * @param offsetKey
+     * @param pageSize
+     * @return
+     */
+    public ForwardCursorPagedResourceList<ParticipantData> getParticipantData(String userId, String offsetKey, int pageSize);
```
DwayneJengSage: Rename to getAllParticipantData.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantDataDao.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Get participant data record for the given userId and identifier
+     * @param userId
+     * @param identifier
+     * @return
+     */
+    ParticipantData getParticipantDataRecord(String userId, String identifier);
```
DwayneJengSage: Rename to getParticipantData

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,114 @@
...
+                .withRangeKeyCondition("offsetKey", rangeKeyCondition)
+                .withLimit(pageSizeWithIndicatorRecord);
+
+        PaginatedQueryList<DynamoParticipantData> resultPage = mapper.query(DynamoParticipantData.class, query);
+
+        // List<ParticipantData> list = Collections.unmodifiableList(resultPage);
+        List<ParticipantData> list = new ArrayList<ParticipantData>();
+        for (ParticipantData participantData : resultPage) {
+            list.add(participantData);
+        }
```
DwayneJengSage: This can be replaced with

```
List<ParticipantData> list = ImmutableList.copyOf(resultPage);
```

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public void setIdentifier(String identifier) {
+        this.identifier = identifier;
+    }
+
+    @Override
+    @DynamoDBTypeConverted(converter = JsonNodeMarshaller.class)
+    //TODO is this a @DynamoDBAttribute?
```
DwayneJengSage: All methods are DynamoDBAttribute by default. You only really need to specify DynamoDBAttribute if, for example, you need to rename the attribute.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
```diff
@@ -0,0 +1,114 @@
...
+    private DynamoDBMapper mapper;
+
+    @Resource(name = "participantDataMapper")
+    final void setParticipantStateMapper(DynamoDBMapper participantStateMapper) {
+        this.mapper = participantStateMapper;
+    }
+
+    public ForwardCursorPagedResourceList<ParticipantData> getParticipantData(String userId, String offsetKey, int pageSize) {
+        checkNotNull(userId);
+        checkNotNull(offsetKey);
```
DwayneJengSage: offsetKey is not required. If offsetKey isn't specified, you should skip setting the range key condition.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,154 @@
...
+import org.sagebionetworks.bridge.models.ParticipantData;
+import org.sagebionetworks.bridge.models.StatusMessage;
+import org.sagebionetworks.bridge.models.accounts.Account;
+import org.sagebionetworks.bridge.models.accounts.AccountId;
+import org.sagebionetworks.bridge.models.accounts.UserSession;
+import org.sagebionetworks.bridge.models.apps.App;
+import org.sagebionetworks.bridge.services.ParticipantDataService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.*;
```
DwayneJengSage: Don't collapse these imports.

src/test/java/org/sagebionetworks/bridge/services/ParticipantDataServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.mockito.Mock;
+import org.sagebionetworks.bridge.dao.ParticipantDataDao;
+import org.sagebionetworks.bridge.dynamodb.DynamoParticipantData;
+import org.sagebionetworks.bridge.dynamodb.DynamoParticipantDataDao;
+import org.sagebionetworks.bridge.models.ParticipantData;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+import static org.testng.Assert.*;
```
DwayneJengSage: Don't collapse these imports.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,49 @@
+package org.sagebionetworks.bridge.spring.controllers;
+
+import org.mockito.Mockito;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.*;
```
DwayneJengSage: Don't collapse these imports.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,185 @@
...
+import org.sagebionetworks.bridge.models.ResourceList;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.sagebionetworks.bridge.TestConstants.TEST_APP_ID;
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertSame;
```
DwayneJengSage: Don't use the JUnit classes. JUnit is a different testing framework. The asserts themselves are harmless enough, but for consistency, we should avoid using them.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,154 @@
...
+
+        int pageSizeInt = getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+        ForwardCursorPagedResourceList<ParticipantData> participantData = participantDataService
+                .getParticipantData(session.getId(), offsetKey, pageSizeInt);
+        List<String> identifiers = participantData.getItems().stream().map(ParticipantData::getIdentifier).collect(Collectors.toList());
+
+        return new ForwardCursorPagedResourceList<String>(identifiers, participantData.getNextPageOffsetKey());
+    }
+
+    @GetMapping("/v4/users/self/data/{identifier}")
```
DwayneJengSage: We should change all the v4's to v3's.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -30,41 +34,121 @@ final void setReportService(ParticipantDataService participantDataService) {
...
+    }
+
+    @DeleteMapping("/v4/users/self/data/{identifier}")
+    public StatusMessage deleteDataByIdentifier(@PathVariable String identifier) {
+        UserSession session = getAuthenticatedAndConsentedSession();
+
+        ParticipantData participantData = participantDataService.getParticipantDataRecord(session.getId(), identifier);
+        if (participantData == null) {
+            throw new EntityNotFoundException(ParticipantData.class);
+        }
```
DwayneJengSage: Move this check to ParticipantDataService.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @GetMapping("/v1/apps/{appId}/participants/{userId}/data")
+    public ForwardCursorPagedResourceList<String> getAllDataForAdminWorker(@PathVariable String appId, @PathVariable String userId,
+                                                                           String offsetKey, String pageSize) {
+        getAuthenticatedSession(ADMIN, WORKER);
+
+        Account account = accountService.getAccount(AccountId.forId(appId, userId));
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
```
DwayneJengSage: We also need to check that, if the caller is an Admin, that session.getAppId() is the same as the passed in appId. (Maybe in a helper function so we don't have to copy-paste it to all of the methods.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,154 @@
...
+        return participantDataService.getParticipantDataRecord(userId, identifier);
+    }
+
+    @PostMapping("/v1/apps/{appId}/participants/{userId}/data/{identifier}")
+    @ResponseStatus(HttpStatus.CREATED)
+    public StatusMessage saveDataRecordForAdminWorker(@PathVariable String appId, @PathVariable String userId, @PathVariable String identifier) {
+        UserSession session = getAuthenticatedSession(ADMIN, WORKER);
+        App app = appService.getApp(appId); // TODO since I already have appId, do I still need to store the session var?
+        // TODO same with storing app var, since I already have appId?
```
DwayneJengSage: No, you don't actually need the app, just the appId.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,154 @@
...
+
+        return new StatusMessage("Participant data saved.");
+    }
+
+    @DeleteMapping("/v1/apps/{appId}/participants/{userId}/data/{identifier}")
+    public StatusMessage deleteDataRecordForAdmin(@PathVariable String appId, @PathVariable String userId,
+                                                  @PathVariable String identifier) {
+        UserSession session = getAuthenticatedSession(ADMIN);
+
+        participantDataService.deleteParticipantDataRecord(userId, identifier); //TODO I also am not using appID here
```
DwayneJengSage: Need to check of the account exists.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,185 @@
...
+            createParticipantData("r", "s"),
+            createParticipantData("t", "u"),
+            createParticipantData("v", "w")
+        );
+    }
+
+    @Test
+    public void testGetParticipantData() {
+        when(mockMapper.query(eq(DynamoParticipantData.class), any())).thenReturn(mockQueryList);
+        when(mockQueryList.iterator()).thenReturn(participantDataList.iterator());
```
DwayneJengSage: There's some weird stuff that goes on with PaginatedQueryList. It might be easier to use a queryHelper and spy it. See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3Dao.java#L153 and https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataEx3DaoTest.java

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Test
+    public void testGetParticipantDataRecord() {
+    }
+
+    @Test
+    public void testSaveParticipantData() {
+        dao.saveParticipantData(participantData0);
+
+        verify(mockMapper).save(participantDataCaptor.capture());
```
DwayneJengSage: You don't need to use a captor. You can just say

```
         verify(mockMapper).save(participantData0);
```

and Mockito will make sure that participantData0 was passed into mockMapper.save().

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,94 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        DynamoParticipantData that = (DynamoParticipantData) o;
+        return Objects.equals(userId, that.userId) && Objects.equals(identifier, that.identifier) &&
+                Objects.equals(data, that.data) && Objects.equals(version, that.version);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(userId, identifier, data, version);
+    }
```
DwayneJengSage: If we don't actually need equals() and hashCode(), let's not include them.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataTest.java
```diff
@@ -0,0 +1,49 @@
...
+        objectNode.put("c", 10);
+
+        participantData.setUserId(userId);
+        participantData.setIdentifier(identifier);
+        participantData.setData(objectNode);
+
+        String json = MAPPER.writeValueAsString(participantData);
+
+        JsonNode node = MAPPER.readTree(json);
+        assertNull(node.get("userId")); // TODO a little confused here
```
DwayneJengSage: There's a JsonIgnore annotation on userId, so when we serialize it to JSON, it gets omitted. We do this because there's not really any reason for our API to show the userId.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertEquals(node.get("identifier").asText(), identifier);
```
DwayneJengSage: We should use textValue() not asText(). asText() will convert non-strings into strings, which could semantically change some of our test results.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataTest.java
```diff
@@ -0,0 +1,49 @@
...
+        String identifier = "baz";
+
+        ObjectNode objectNode = JsonNodeFactory.instance.objectNode();
+        objectNode.put("a", true);
+        objectNode.put("b", "string");
+        objectNode.put("c", 10);
+
+        participantData.setUserId(userId);
+        participantData.setIdentifier(identifier);
+        participantData.setData(objectNode);
```
DwayneJengSage: Let's add the Version attribute so we can verify that's being serialized also.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertTrue(node.get("data").get("a").booleanValue());
+        assertEquals(node.get("data").get("b").textValue(), "string");
+        assertEquals(node.get("data").get("c").intValue(), 10);
+        assertEquals(node.get("type").textValue(), "ParticipantData");
+        assertEquals(node.size(), 3);
+
+        ParticipantData deser = MAPPER.readValue(json, ParticipantData.class);
+        assertTrue(deser.getData().get("a").asBoolean());
+        assertEquals(deser.getData().get("b").asText(), "string");
+        assertEquals(deser.getData().get("c").asInt(), 10);
```
DwayneJengSage: We should also verify deser.getIdentifier() and deser.getVersion().

DwayneJengSage (changes requested):
The structure of the code and the existing tests looks good. Mostly low-level nitty gritty issues.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDao.java
kllysng: I was hoping to talk about this with you today - I tried to this but for some reason I'm not able to mock this line in the Unit test. Maybe you have a way that I'm not aware of!

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
kllysng: Oh, I see this as potentially the resolution to my question in your comment about using ImmutableList.copyOf(). Thank you, I'll look into this!!

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
kllysng: Looking through the custom exceptions I see UnauthorizedException - if the appId of the Admin and the passed in appId don't match, is that the correct exception to throw?

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantData.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    @DynamoDBVersionAttribute
+    public Long getVersion() {
+        return this.version;
+    }
+
+    @Override
+    public void setVersion(Long version) {
+        this.version = version;
+    }
```
DwayneJengSage: After chatting with Nathaniel, we don't actually want the Version attribute here. The apps don't need version conflict detection, and reading before writing will make the API harder to use for them.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
DwayneJengSage: The convention is that we throw 404 not found when someone tries to access something outside of their app or study. This prevents people from fishing for a user or identifier in a study outside of their own.

403 Unauthorized is generally used of the caller doesn't have access to an API, for example if developer tries to call a researcher API.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,167 @@
...
+        return new StatusMessage("Participant data deleted.");
+    }
+
+    @GetMapping("/v1/apps/{appId}/participants/{userId}/data/{identifier}")
+    public ParticipantData getDataByIdentifierForAdminWorker(@PathVariable String appId, @PathVariable String userId,
+                                                             @PathVariable String identifier) {
+        getAuthenticatedSession(ADMIN, WORKER);
+
+        checkAccountExists(appId, userId);
+        checkAccountExists(appId, userId);
```
DwayneJengSage: I think you meant checkAdminSessionAppId().

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,167 @@
...
+
+        participantDataService.deleteParticipantData(userId, identifier);
+
+        return new StatusMessage("Participant data deleted.");
+    }
+
+    private void checkAdminSessionAppId(UserSession session, String appId) {
+        if (session.isInRole(ADMIN) && (!appId.equals(session.getAppId()))) { //TODO: waiting to hear from dwayne if this is the correct exception
+            throw new UnauthorizedException("Caller does not have permission to access participant data.");
+        }
```
DwayneJengSage: Based on other uses in our code base, if the account doesn't have access to the app, we throw an EntityNotFoundException.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataController.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @GetMapping("/v1/apps/{appId}/participants/{userId}/data")
+    public ForwardCursorPagedResourceList<String> getAllDataForAdminWorker(@PathVariable String appId, @PathVariable String userId,
+                                                                           String offsetKey, String pageSize) {
+        UserSession session = getAuthenticatedSession(ADMIN, WORKER);
+
+        Account account = accountService.getAccount(AccountId.forId(appId, userId));
+        if (account == null) {
+            throw new EntityNotFoundException(Account.class);
+        }
```
DwayneJengSage: Replace this with checkAccountExists()?

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+        assertEquals(participantData.getUserId(), TEST_USER_ID);
+        assertEquals(participantData.getIdentifier(), IDENTIFIER);
+    }
+
+    @Test
+    public void testSaveParticipantData() {
+        dao.saveParticipantData(participantData0);
+
+        verify(mockMapper).save(participantDataCaptor.capture());
+        // TODO if I change the above lines to verify(mockMapper).save(participantData0), then how will I get the below participantData?
```
DwayneJengSage: If you call verify(mockMapper).save(participantData0), it will verify that the passed in participantData is the same as participantData0.

src/main/java/org/sagebionetworks/bridge/services/ParticipantDataService.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public ForwardCursorPagedResourceList<ParticipantData> getAllParticipantData(String userId, String offsetKey, int pageSize) {
+        if (pageSize < API_MINIMUM_PAGE_SIZE || pageSize > API_MAXIMUM_PAGE_SIZE) {
+            throw new BadRequestException(BridgeConstants.PAGE_SIZE_ERROR);
+        }
+
+        ForwardCursorPagedResourceList<ParticipantData> allParticipantData = participantDataDao.getAllParticipantData(userId, offsetKey, pageSize);
+        if (allParticipantData == null) {
+            throw new EntityNotFoundException(ParticipantData.class);
+        }
```
DwayneJengSage: Not necessary to have a null check here. If there are no results, DynamoDB will return an empty list, and we can return an empty list to the caller just fine.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+    public void testDeleteAllParticipantData() {
+        when(mockMapper.query(eq(DynamoParticipantData.class), any())).thenReturn(mockQueryList);
+
+        dao.deleteAllParticipantData(participantData0.getUserId());
+
+        verify(mockMapper).query(eq(DynamoParticipantData.class), queryCaptor.capture());
+        DynamoDBQueryExpression<DynamoParticipantData> query = queryCaptor.getValue();
+        assertEquals(query.getHashKeyValues().getUserId(), participantData0.getUserId());
+
+        verify(mockMapper).batchDelete(dataListCaptor.capture());
```
DwayneJengSage: Similarly, we can just do verify(mockMapper).batchDelete(mockQueryList)

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+    @Test
+    public void testDeleteParticipantDataRecord() {
+        when(mockMapper.load(any())).thenReturn(participantData0);
+
+        String identifier = participantData0.getIdentifier();
+        dao.deleteParticipantData(participantData0.getUserId(), identifier);
+
+        verify(mockMapper).load(participantDataCaptor.capture());
+        assertEquals(participantDataCaptor.getValue().getUserId(), participantData0.getUserId());
+        assertEquals(participantDataCaptor.getValue().getIdentifier(), identifier);
```
DwayneJengSage: For completeness, let's also verify(mapper).delete()

src/test/java/org/sagebionetworks/bridge/services/ParticipantDataServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Mock
+    ParticipantDataDao mockDao;
+
+    @Captor
+    ArgumentCaptor<ParticipantData> participantDataCaptor;
+
+    @Spy
+    ParticipantDataService service;
+
+    ForwardCursorPagedResourceList<DynamoParticipantData> results;
```
DwayneJengSage: We should change the type to ParticipantData, so that the Dynamo implementation doesn't get leaked outside the Dynamo classes.

src/test/java/org/sagebionetworks/bridge/services/ParticipantDataServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+        ParticipantData retrieved = service.getParticipantData(TEST_USER_ID, IDENTIFIER);
+
+        verify(mockDao).getParticipantData(TEST_USER_ID, IDENTIFIER);
+        assertEquals(retrieved, result);
+    }
+
+    @Test
+    public void testSaveParticipantData() {
+        ParticipantData participantData = createParticipantData("c", "d");
+        service.saveParticipantData(TEST_USER_ID, IDENTIFIER, participantData);
```
DwayneJengSage: We should call our service with a different userId and identifier just to make sure that our code properly sets the userId and identifier on the saved ParticipantData.

src/test/java/org/sagebionetworks/bridge/services/ParticipantDataServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Test(expectedExceptions = EntityNotFoundException.class)
+    public void testGetAllDataNoData() {
+        service.getAllParticipantData("", "", PAGE_SIZE);
+    }
+
+    private static DynamoParticipantData createParticipantData(String fieldValue1, String fieldValue2) {
+        ObjectNode node = JsonNodeFactory.instance.objectNode();
+        node.put("field1", fieldValue1);
+        node.put("field2", fieldValue2);
+        DynamoParticipantData participantData = new DynamoParticipantData();
```
DwayneJengSage: You can avoid mentioning Dynamo by using ParticipantData.create()

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+    ParticipantDataController controller;
+
+    UserSession session;
+    ParticipantData participantData;
+
+    @BeforeMethod
+    public void beforeMethod() {
+        MockitoAnnotations.initMocks(this);
+
+        DynamoApp app = new DynamoApp();
```
DwayneJengSage: Similarly, you can avoid mentioning DynamoApp by using App.create().

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+    public void testGetAllDataForSelf() {
+        doReturn(makeResults(OFFSET_KEY, PAGE_SIZE_INT)).when(mockParticipantDataService)
+                .getAllParticipantData(TEST_USER_ID, OFFSET_KEY, PAGE_SIZE_INT);
+
+        ForwardCursorPagedResourceList<String> result = controller.getAllDataForSelf(OFFSET_KEY, PAGE_SIZE_STRING);
+
+        assertEquals(NEXT_PAGE_OFFSET_KEY, result.getNextPageOffsetKey());
+        assertEquals(result.getItems().get(0), IDENTIFIER);
+        assertEquals(result.getItems().get(1), IDENTIFIER);
+    }
```
DwayneJengSage, line 154: Add a test case without offsetKey and pageSize and verify that the default pageSize is passed into the Service.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+    public void testGetDataByIdentifierForSelf() {
+        doReturn(participantData).when(mockParticipantDataService).getParticipantData(session.getId(), IDENTIFIER);
+
+        ParticipantData result = controller.getDataByIdentifierForSelf(IDENTIFIER);
+
+        assertEquals(result.getUserId(), participantData.getUserId());
+        assertEquals(result.getIdentifier(), participantData.getIdentifier());
+        assertEquals(result.getData(), participantData.getData());
+        assertEquals(result.getVersion(), participantData.getVersion());
+        assertSame(result, participantData);
```
DwayneJengSage: If you verify that result and participantData are the same, you don't need to verify that all the parameters are equal, because they are guaranteed to be equal.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+        StatusMessage result = controller.deleteAllParticipantDataForAdmin(session.getAppId(), session.getId());
+        assertEquals(result.getMessage(), "Participant data deleted.");
+
+        verify(mockParticipantDataService).deleteAllParticipantData(TEST_USER_ID);
+    }
+
+    @Test (expectedExceptions = UnauthorizedException.class)
+    public void testDeleteAllParticipantDataForAdminUnAuthorizedException() {
+        controller.deleteAllParticipantDataForAdmin(TEST_APP_ID, TEST_USER_ID);
+    }
```
DwayneJengSage: I don't think this test case actually tests our checkAdminSessionAppId() logic. Also, we need test cases for all the other admin APIs as well.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+        doReturn(expected).when(mockParticipantDataService).getAllParticipantData(session.getId(), OFFSET_KEY, PAGE_SIZE_INT);
+
+        ForwardCursorPagedResourceList<String> result = controller.getAllDataForAdminWorker(session.getAppId(),
+                session.getId(), OFFSET_KEY, PAGE_SIZE_STRING);
+
+        assertEquals(expected.getItems().get(0).getIdentifier(), result.getItems().get(0));
+        assertEquals(expected.getItems().size(), result.getItems().size());
+
+        verify(mockParticipantDataService).getAllParticipantData(TEST_USER_ID, OFFSET_KEY, PAGE_SIZE_INT);
+    }
```
DwayneJengSage, line 215: Similarly, we need a test case for null offsetKey and default pageSize.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+    public void testGetDataByIdentifierForAdminWorker() {
+        doReturn(participantData).when(mockParticipantDataService).getParticipantData(session.getId(), IDENTIFIER);
+
+        ParticipantData result = controller.getDataByIdentifierForAdminWorker(TEST_APP_ID, TEST_USER_ID, IDENTIFIER);
+
+        assertEquals(result.getUserId(), participantData.getUserId());
+        assertEquals(result.getIdentifier(), participantData.getIdentifier());
+        assertEquals(result.getData(), participantData.getData());
+        assertEquals(result.getVersion(), participantData.getVersion());
+        assertSame(result, participantData);
```
DwayneJengSage: Similarly, we don't need all the assertEquals() if we use assertSame()

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantDataDaoTest.java
```diff
@@ -0,0 +1,211 @@
...
```
DwayneJengSage: For completeness, let's also verify(mapper).delete()

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,323 @@
...
+    public void testGetAllDataForSelf() {
+        doReturn(makeResults(OFFSET_KEY, PAGE_SIZE_INT)).when(mockParticipantDataService)
+                .getAllParticipantData(TEST_USER_ID, OFFSET_KEY, PAGE_SIZE_INT);
+
+        ForwardCursorPagedResourceList<String> result = controller.getAllDataForSelf(OFFSET_KEY, PAGE_SIZE_STRING);
+
+        assertEquals(NEXT_PAGE_OFFSET_KEY, result.getNextPageOffsetKey());
+        assertEquals(result.getItems().get(0), IDENTIFIER);
+        assertEquals(result.getItems().get(1), IDENTIFIER);
+    }
```
DwayneJengSage, line 154: Add a test case without offsetKey and pageSize and verify that the default pageSize is passed into the Service.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+        verify(mockParticipantDataService).deleteAllParticipantData(TEST_USER_ID);
+    }
+
+    @Test (expectedExceptions = UnauthorizedException.class)
+    public void testDeleteAllParticipantDataForAdminUnAuthorizedException() {
+        controller.deleteAllParticipantDataForAdmin(TEST_APP_ID, TEST_USER_ID);
+    }
+
+    // TODO: test checkAdminSessionAppId() logic
```
DwayneJengSage: TODO

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,323 @@
...
+        doReturn(expected).when(mockParticipantDataService).getAllParticipantData(session.getId(), OFFSET_KEY, PAGE_SIZE_INT);
+
+        ForwardCursorPagedResourceList<String> result = controller.getAllDataForAdminWorker(session.getAppId(),
+                session.getId(), OFFSET_KEY, PAGE_SIZE_STRING);
+
+        assertEquals(expected.getItems().get(0).getIdentifier(), result.getItems().get(0));
+        assertEquals(expected.getItems().size(), result.getItems().size());
+
+        verify(mockParticipantDataService).getAllParticipantData(TEST_USER_ID, OFFSET_KEY, PAGE_SIZE_INT);
+    }
```
DwayneJengSage, line 215: Similarly, we need a test case for null offsetKey and default pageSize.

DwayneJengSage (changes requested):
Looks good so far. Just adding in these comments to make sure we don't forget them.

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,386 @@
...
+    @Test
+    public void testGetAllDataForSelfPageSizeInService() {
+        ArgumentCaptor<Integer> integerCaptor = ArgumentCaptor.forClass(Integer.class);
+        doReturn(makeResults(null, null)).when(mockParticipantDataService)
+                .getAllParticipantData(eq(TEST_USER_ID), eq(null), integerCaptor.capture());
+
+        controller.getAllDataForSelf(null, null);
+
+        assertEquals(API_DEFAULT_PAGE_SIZE, integerCaptor.getValue().intValue());
+    }
```
DwayneJengSage: It would be better to use verify. Also, you don't need a captor for the integer. You can just do

```
verify(mockParticipantDataService).getAllParticipantData(TEST_USER_ID, null, API_DEFAULT_PAGE_SIZE)
```

src/test/java/org/sagebionetworks/bridge/spring/controllers/ParticipantDataControllerTest.java
```diff
@@ -0,0 +1,386 @@
...
+    @Test
+    public void testGetAllDataForAdminWorkerDefaultPageSizeInService() {
+        ArgumentCaptor<Integer> integerCaptor = ArgumentCaptor.forClass(Integer.class);
+        doReturn(makeResults(null, null)).when(mockParticipantDataService)
+                .getAllParticipantData(eq(TEST_USER_ID), eq(null), integerCaptor.capture());
+
+        controller.getAllDataForAdminWorker(TEST_APP_ID, TEST_USER_ID, null, null);
+
+        assertEquals(API_DEFAULT_PAGE_SIZE, integerCaptor.getValue().intValue());
+    }
```
DwayneJengSage: Similarly here.

DwayneJengSage (changes requested):
Overall looks good. Just two minor comments about tests.

DwayneJengSage (approved):

PR #314, opened 2021-01-25
Security rules cleanup after study coordinator merge.

alxdarksage:
There are still issues, documented here: https://sagebionetworks.jira.com/browse/BRIDGE-2897

However this is a renaming step that touches a lot of files, so I wanted to do it first. There is one test change in the integration tests that accompanies this refactor.


nategbrown9 (commented):
Looks good.

nategbrown9 (approved):

PR #316, opened 2021-01-26
CRC API to request lab shipped to home

liujoshua:


src/main/java/org/sagebionetworks/bridge/models/crc/gbf/external/CheckOrderStatusResponse.java
```diff
@@ -0,0 +1,32 @@
...
+        public final String orderNumber;
+        public final String orderStatus;
+
+        @JsonCreator
+        public OrderStatus(@JsonProperty("orderNumber") String orderNumber, @JsonProperty("orderStatus") String orderStatus) {
+            this.orderNumber = orderNumber;
+            this.orderStatus = orderStatus;
+        }
+    }
+}
```
alxdarksage, line 32: I think we'd like to nounify payloads, so this would be an OrderStatus.

src/main/java/org/sagebionetworks/bridge/services/GBFOrderService.java
```diff
@@ -0,0 +1,188 @@
...
+        } catch (IOException e) {
+            LOG.error("placeOrder failed to read contents of HttpResponse", e);
+            throw new BridgeServiceException(GBF_SERVICE_ERROR_MESSAGE);
+        }
+
+        if (!response.success) {
+            LOG.warn("placeOrder received error from remote service: {}", response.message);
+            throw new BridgeServiceException(GBF_SERVICE_ERROR_MESSAGE);
+        }
+    }
```
alxdarksage, line 91: What do we do with response?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -181,6 +203,129 @@ String getUserAgent() {
...
+    public ResponseEntity<StatusMessage> postUserLabShipmentRequest() {
+        // caller enrolled studies
+        UserSession session = getAuthenticatedSession();
+        App app = appService.getApp(session.getAppId());
+        RequestContext requestContext = RequestContext.get();
+        Set<String> callerEnrolledStudies = requestContext.getCallerEnrolledStudies();
+        Set<String> orgSponsoredStudies = requestContext.getOrgSponsoredStudies();
+        RequestContext.set(requestContext.toBuilder().withOrgSponsoredStudies(callerEnrolledStudies)
+                .withCallerEnrolledStudies(
+                        orgSponsoredStudies).build());
```
alxdarksage: ?! I would rather not swap these as that seems like it will be *really* confusing...how about setting them to nulls or empty sets?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -181,6 +203,129 @@ String getUserAgent() {
...
+        );
+    }
+
+    // Waiting for integration workflow to be finalized
+    //@GetMapping(path = "v1/cuicm/labshipments/{orderId}/status", produces = {APPLICATION_JSON_UTF8_VALUE})
+    public String getLabShipmentStatus(@PathVariable String orderId) throws JsonProcessingException {
+        App app = httpBasicAuthentication();
+        CheckOrderStatusResponse response = gbfOrderService.checkOrderStatus(orderId);
+        return MAPPER.writeValueAsString(response);
+    }
```
alxdarksage, line 319: Can't return CheckOrderStatusResponse and let Spring convert it?

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -181,6 +203,129 @@ String getUserAgent() {
...
+    public String getLabShipmentConfirmations(@RequestParam String startDate,
+                                              @RequestParam String endDate) throws JsonProcessingException {
+        App app = httpBasicAuthentication();
+
+        LocalDate startDateObj = getLocalDateOrDefault(startDate, null);
+        LocalDate endDateObj = getLocalDateOrDefault(endDate, null);
+
+        ShippingConfirmations shippingConfirmations = gbfOrderService.requestShippingConfirmations(startDateObj, endDateObj);
+
+        return MAPPER.writeValueAsString(shippingConfirmations);
```
alxdarksage: I would let Spring convert it...

src/test/java/org/sagebionetworks/bridge/spring/controllers/CRCControllerTest.java
```diff
@@ -1439,7 +1571,7 @@ private String makeAppointment(String identifier, AppointmentStatus status) {
         return FHIR_CONTEXT.newJsonParser().encodeResourceToString(appt);
     }
     
-    private String makeProcedureRequest() { 
+    private String makeProcedureRequest() {
         ProcedureRequest procedure = new ProcedureRequest();
         
         Identifier id = new Identifier();
```
alxdarksage, line 1667: Some tests that appear to be missing:

internalLabShipmentRequest()
    - account not found test
    - ship requests limited to one
    
validateAndGetAddress()
    - missing first name test
    - missing last name test
    - missing address1 test
    - missing state test
    - missing zip code test
    - test if phone is invalid
    - no test of getLabShipmentStatus

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
liujoshua: Participant self call does not have OrgSponsoredStudies

Admin call for a participant does have OrgSponsoredStudies. I remove this so the report's studies match when called from either endpoint

alxdarksage (approved):

PR #317, opened 2021-01-28
Enrollment events

alxdarksage:
EDIT: I am going to change the deletion behavior so you call a DELETE endpoint like every other API on Bridge. This will certainly change where the timestamp validation occurs. Hold off on reviewing until this is updated.

An enrollment event is now created no matter how someone is enrolled in a study (and it's study-specific where appropriate).

It's now possible to delete custom events (this will be very helpful for testing schedules).

Fixed a bug: if ActivityEvent was validated and threw an InvalidEntityException, it revealed the healthCode field. Changed how this is serialized so this won't happen and adjusted tests appropriately.

Changed the StudyParticipantController to allow for all forms of ID token. In working with the API I found it very helpful to be able to use externalId:* rather than the ID to retrieve external ID-based accounts (otherwise the client needs to do an ID lookup first).


nategbrown9 (approved):
Looks good.

PR #322, opened 2021-02-05
S3 bootstrapper

alxdarksage:
- added a "noinit" Spring profile. When you start BridgeServer2 with this profile, it takes off ~30 seconds from your start time by skipping db migrations and our boostrapping code for DynamoDB and S3;
- removed redundant S3 clients and S3 helpers;
- created an S3 initializer during bootstrapping to create missing S3 buckets.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/S3Initializer.java
+++ b/src/main/java/org/sagebionetworks/bridge/S3Initializer.java
@@ -5,8 +5,12 @@ import static org.sagebionetworks.bridge.BridgeUtils.resolveTemplate;
 import java.util.Map;
 
 import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.BucketCrossOriginConfiguration;
+import com.amazonaws.services.s3.model.CORSRule;
 import com.amazonaws.services.s3.model.CreateBucketRequest;
 import com.amazonaws.services.s3.model.Region;
+import com.amazonaws.services.s3.model.CORSRule.AllowedMethods;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
 import org.apache.commons.lang3.StringUtils;
@@ -63,15 +67,24 @@ public class S3Initializer {
             + "        }"
             + "    ]"
             + "}";
-
+    
+    private static final CORSRule ALLOW_PUT_RULE = new CORSRule()
+            .withAllowedHeaders(ImmutableList.of("*"))
+            .withAllowedMethods(ImmutableList.of(AllowedMethods.PUT))
+            .withAllowedOrigins(ImmutableList.of("*"))
+            .withMaxAgeSeconds(30000);
+    
     public static enum BucketType {
-        INTERNAL(null),
-        SYNAPSE_ACCESSIBLE(SYNAPSE_ACCESS_POLICY),
-        PUBLIC_ACCESSIBLE(PUBLIC_ACCESS_POLICY);
+        INTERNAL(null, null),
+        INTERNAL_UPLOAD_ACCESSIBLE(null, ALLOW_PUT_RULE),
+        SYNAPSE_ACCESSIBLE(SYNAPSE_ACCESS_POLICY, null),
+        PUBLIC_ACCESSIBLE(PUBLIC_ACCESS_POLICY, null);
         
         String policy;
-        private BucketType(String policy) {
+        CORSRule corsSupport;
+        private BucketType(String policy, CORSRule corsSupport) {
             this.policy = policy;
+            this.corsSupport = corsSupport;
         }
     }
     
@@ -93,7 +106,7 @@ public class S3Initializer {
             .put("attachment.bucket", BucketType.SYNAPSE_ACCESSIBLE)
             .put("health.data.bucket.raw", BucketType.SYNAPSE_ACCESSIBLE)
             .put("health.data.bucket.processed", BucketType.SYNAPSE_ACCESSIBLE)
-            .put("upload.bucket", BucketType.INTERNAL)
+            .put("upload.bucket", BucketType.INTERNAL_UPLOAD_ACCESSIBLE)
             .put("upload.cms.cert.bucket", BucketType.INTERNAL)
             .put("upload.cms.priv.bucket", BucketType.INTERNAL)
             .put("consents.bucket", BucketType.INTERNAL)
@@ -124,6 +137,11 @@ public class S3Initializer {
                 if (type.policy != null) {
                     String policy = resolveTemplate(type.policy, ImmutableMap.of("bucketName", bucketName));
                     s3Client.setBucketPolicy(bucketName, policy);
+                }
+                if (type.corsSupport != null) {
+                    BucketCrossOriginConfiguration config = new BucketCrossOriginConfiguration()
+                            .withRules(type.corsSupport);
+                    s3Client.setBucketCrossOriginConfiguration(bucketName, config);
                 }
             }
         }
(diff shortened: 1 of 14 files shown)


src/main/java/org/sagebionetworks/bridge/services/AppService.java
```diff
@@ -339,7 +340,12 @@ public App createApp(App app) {
         
         Study study = Study.create();
         study.setAppId(app.getIdentifier());
-        study.setIdentifier(app.getIdentifier() + "-study");
+        // For legacy reasons, these two default apps use non-standard initial study IDs. 
+        if (BOOTSTRAP_STUDY_IDS.contains(app.getIdentifier())) {
+            study.setIdentifier(app.getIdentifier());
+        } else {
+            study.setIdentifier(app.getIdentifier() + "-study");    
+        }
```
DwayneJengSage: I don't think this is actually necessary. For api, the Integ Tests will immediately replace the default study with with study1, and the shared study doesn't actually have any participants, so study doesn't matter.

alxdarksage: Hmm okay, my integration tests seemed to fail but I will test this out again.

DwayneJengSage:
Adding link to https://sagebionetworks.jira.com/browse/BRIDGE-2909 for reference.

src/main/resources/BridgeServer2.conf
```diff
@@ -1,6 +1,20 @@
...
+uat.bucket.suffix = uat
+prod.bucket.suffix = prod
+
+# AWS server account
+aws.key = dummy-value
+aws.secret.key = dummy-value
+
+# AWS credentials for sending push notifications
+sns.key = dummy-value
+sns.secret.key = dummy-value
```
DwayneJengSage: As long as we're deleting the duplicate AWS keys, we might as well delete sns.key as well. In practice, this is already the same as aws.key.

src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        when(mockS3Client.doesBucketExistV2(BUCKET_NAME)).thenReturn(false);
+        
+        initializer.initBuckets();
+        
+        String resolvedPolicy = BridgeUtils.resolveTemplate(
+                S3Initializer.BucketType.SYNAPSE_ACCESSIBLE.policy, 
+                ImmutableMap.of("bucketName", BUCKET_NAME));
+        
+        verify(mockS3Client).createBucket(requestCaptor.capture());
+        verify(mockS3Client).setBucketPolicy(eq(BUCKET_NAME), stringCaptor.capture());
```
DwayneJengSage: Minor: You don't actually need stringCaptor. You can just do eq(resolvedPolicy).

src/main/resources/BridgeServer2.conf
```diff
@@ -77,21 +85,11 @@ dev.synapse.tracking.view = syn20683692
...
-# AWS credentials for doing pre-signed upload
-aws.key.upload = dummy-value
-aws.secret.key.upload = dummy-value
-
 # Upload buckets
-local.upload.bucket = org-sagebridge-upload-local
-dev.upload.bucket = org-sagebridge-upload-develop
-uat.upload.bucket = org-sagebridge-upload-uat
-prod.upload.bucket = org-sagebridge-upload-prod
+local.upload.bucket = org-sagebridge-upload-${bucket.suffix}
```
DwayneJengSage: Should this be upload.bucket instead of local.upload.bucket? (Here and everywhere else.)

src/main/resources/BridgeServer2.conf
```diff
@@ -101,58 +99,38 @@ upload.cms.certificate.organization = Sage Bionetworks
...
-prod.app.whitelist = api,asthma,breastcancer,cardiovascular,diabetes,fphs,fphs-lab,ohsu-molemapper,parkinson,parkinson-lux,lilly
+local.app.whitelist = api,shared
+dev.app.whitelist = api,shared
+uat.app.whitelist = api,shared,ios-sdk-int-tests,asthma,cardiovascular
+prod.app.whitelist = api,shared,asthma,breastcancer,cardiovascular,diabetes,fphs,fphs-lab,ohsu-molemapper,parkinson,parkinson-lux,lilly
 
-local.usersigned.consents.bucket = local.usersigned.consents.bucket
+local.usersigned.consents.bucket = org-sagebridge-usersigned-consents-${bucket.suffix}
 dev.usersigned.consents.bucket = bridgepf-develop-awss3usersignedconsentsdownloadb-apwbxc8ldmj2
 uat.usersigned.consents.bucket = bridgepf-uat-awss3usersignedconsentsdownloadbucke-hcuoz4eztd8g
```
DwayneJengSage: We can change dev and staging to follow the pattern. These aren't actually created by Infra. We'd have to create new buckets for dev and staging, but with the AWS migration, that's going to happen anyway.

src/main/resources/BridgeServer2.conf
```diff
@@ -199,8 +177,13 @@ gbf.api.key = dummy-value
...
-# local.participant-file.bucket = org-sagebridge-participantfile-prod-local-[bridge-user]
+participant-file.bucket = org-sagebridge-participantfile-${bucket.suffix}
+
+# Public documents folder
+docs.bucket = docs${host.postfix}
+
+# Note however that the CloudFormation-created buckets are not this straight-forward, and
+# will have to be special-cased when added. 
+health.data.bucket.raw = bridgeserver2-${bucket.suffix}-rawhealthdatabucket
+health.data.bucket.processed = bridgeserver2-${bucket.suffix}-processedhealthdatabucket
```
DwayneJengSage: Don't worry about these for now. These *are* created by Infra and are being wired up automatically. (We can migrate these off Infra and onto the S3Initializer later.)

src/main/java/org/sagebionetworks/bridge/S3Initializer.java
```diff
@@ -0,0 +1,148 @@
...
+    
+    @Autowired
+    public final void setS3Client(AmazonS3Client s3Client) {
+        this.s3Client = s3Client;
+    }
+    
+    private static Map<String, BucketType> S3_BUCKET_PROP_NAMES = new ImmutableMap.Builder<String, BucketType>()
+            .put("attachment.bucket", BucketType.SYNAPSE_ACCESSIBLE)
+            .put("health.data.bucket.raw", BucketType.SYNAPSE_ACCESSIBLE)
+            .put("health.data.bucket.processed", BucketType.SYNAPSE_ACCESSIBLE)
```
DwayneJengSage: Similarly, we can not worry about these for now and migrate them at a later date.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/S3Initializer.java
alxdarksage: Removed the health.data.* keys.

src/main/resources/BridgeServer2.conf
alxdarksage: Removed

src/main/resources/BridgeServer2.conf
alxdarksage: Yes, fixed this, attachment.bucket, and consents.bucket

src/main/resources/BridgeServer2.conf
alxdarksage: Removed dev and staging exceptions, local.usersigned.* renamed to usersigned.*

src/main/resources/BridgeServer2.conf
alxdarksage: Removed

src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
alxdarksage: Removed stringCaptor the two places I was using it

alxdarksage (commented):
Updates based on CR.

DwayneJengSage (approved):

PR #330, opened 2021-02-19
Security refactor 1

alxdarksage:
- researchers no longer constrained to view specific studies (this behavior is implemented through study coordinator and a separate set of APIs);
-  removed the callerConsideredGlobal() rule from the AuthEvaluator, because this was really a loophole. Tracked down the situations where the loophole was being used and it comes down to these use cases:
    - unauthenticated calls to operate on accounts, like `resendVerification` calls (now acquires the identity of the account before checks would cause this to fail);
    - was checking ownership of an account (`isSelf()`) after authentication, when the caller was not yet identified with that account. Either the check is bypassed, the already checked account is carried over to the new call, or the call picks up the identity of the account;
    - call starts a new thread and RequestContext is not carried over (we reacquire identity with the `RequestContext.acquireAccountIdentity()` call)
    
 JIRAs this addresses:
 - https://sagebionetworks.jira.com/browse/BRIDGE-2897
 - https://sagebionetworks.jira.com/browse/BRIDGE-2899
 
 I think this is it for security changes for now, unless we add a study developer role.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/RequestContext.java
+++ b/src/main/java/org/sagebionetworks/bridge/RequestContext.java
@@ -12,7 +12,6 @@ import com.google.common.collect.Sets;
 import org.sagebionetworks.bridge.models.ClientInfo;
 import org.sagebionetworks.bridge.models.Metrics;
 import org.sagebionetworks.bridge.models.accounts.Account;
-import org.sagebionetworks.bridge.models.accounts.ExternalIdentifier;
 import org.sagebionetworks.bridge.models.accounts.StudyParticipant;
 import org.sagebionetworks.bridge.models.accounts.UserSession;
 import org.sagebionetworks.bridge.services.SponsorService;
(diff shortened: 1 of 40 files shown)


src/main/java/org/sagebionetworks/bridge/RequestContext.java
```diff
@@ -62,22 +62,17 @@ public static RequestContext updateFromSession(UserSession session, SponsorServi
...
-        RequestContext context = get();
-        RequestContext.Builder builder = context.toBuilder();
-        if (externalId.getStudyId() != null) {
-            builder.withCallerEnrolledStudies(new ImmutableSet.Builder<String>()
-                .addAll(context.getCallerEnrolledStudies())
-                .add(externalId.getStudyId()).build());
-        }
+    public static RequestContext acquireAccountIdentity(Account account) {
+        RequestContext.Builder builder = get().toBuilder();
+        builder.withCallerUserId(account.getId());
```
DwayneJengSage, line 71: Unit tests?

src/test/java/org/sagebionetworks/bridge/validators/ActivityEventValidatorTest.java
```diff
@@ -47,7 +47,6 @@ public void eventIdNull() {
         assertValidatorMessage(INSTANCE, event, "eventId", EVENT_ID_ERROR);
     }
 
-    @Test
```
DwayneJengSage: Was this test intentionally removed?

DwayneJengSage (changes requested):
Overall looks good. A few comments about tests.

src/test/java/org/sagebionetworks/bridge/validators/ActivityEventValidatorTest.java
alxdarksage: No. Added it back. :-( I may have been contemplating removing this event type since it's never been used, but it's still part of our API.

src/main/java/org/sagebionetworks/bridge/RequestContext.java
alxdarksage: Added.

DwayneJengSage (approved):

PR #331, opened 2021-02-20
Add mimeType validation

alxdarksage:
Mime type is not required when you create a file, but then later your presigned URL will not be correct and you'll never be able to download the file again. So, require the mime type when creating a file handle for upload.


nategbrown9 (approved):
Looks good.

PR #332, opened 2021-02-22
Add clientData to study

alxdarksage:
This will allow MTB administrative app to store some state information about how users should be added to the study.


nategbrown9 (approved):
Looks good.

PR #333, opened 2021-02-24
Study additions

alxdarksage:
Some security-related tweaks based on manual testing in the Bridge Study Manager and our evolving discussions about security roles on the Bridge Server as we implement MTB.

- developers can edit studies, so they should be able to read studies (added this permission);
- if you don't include the version attribute in the summary study objects, you can't submit them to logically delete from BSM's summary/list view (so I've added that field);
- if you are a user who unfortunately is not associated to any studies, then you should see _no_ studies, not _all_ studies, so there is a difference in the check between a null set of studies and an empty set of studies (fixed this);
- developers/study developers can create new studies, at least for now. They therefore need these studies to be sponsored by their organizations, but a security check prevented devs from changing sponsorship relationships (now fixed);
- opened a couple of more endpoints for participants to be accessible to developers as long as they are operating on their own account. This is currently our support for testing...we create an account for a developer and then they use it to enroll, and they can see what is going on with the account as a participant. We need more to get people off of the researcher role, but this is at least working for now.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -41,7 +41,8 @@ public class AuthUtils {
             .hasAnyRole(ADMIN);
     
     /**
-     * Can the caller and/remove organization members? Must be the organizations's admin.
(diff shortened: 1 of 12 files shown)


nategbrown9 (approved):
Looks good.

PR #342, opened 2021-03-18
Support the 'externalId' field on sign up where an account doesn't already exist

alxdarksage:
After pushing changes to production, we learned that PKU study was receiving 400s on sign up because it was not passing validation with a sign up payload. The payload only included an external ID, and did not already have an account in the system (which is a rare configuration...I can't confirm there are other apps that do this). So the existing trap for this was insufficient. I've added more aggressive catches to maintain this app and a couple of others where it could conceivably be an issue, and moved this code to AuthenticationService where it makes more sense and is easier to test.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
@@ -412,17 +412,6 @@ public class ParticipantService {
             throwExceptionIfLimitMetOrExceeded(app);
         }
         
-        // Fix for callers that include only externalId and not the new externalId map: in these cases, tools are 
-        // creating the participant before the sign up call (through the Bridge Study Manager). Check and if 
-        // the account with this external ID already exists, return quietly. Otherwise proceed as before.
-        if (participant.getExternalId() != null && participant.getExternalIds().isEmpty()) {
-            AccountId accountId = AccountId.forExternalId(app.getIdentifier(), participant.getExternalId());
-            Account account = accountService.getAccount(accountId); 
-            if (account != null) {
-                return new IdentifierHolder(account.getId());
-            }
-        }
-        
         StudyParticipantValidator validator = new StudyParticipantValidator(studyService, organizationService, app,
                 true);
         Validate.entityThrowingException(validator, participant);
@@ -498,7 +487,7 @@ public class ParticipantService {
         }
         return new IdentifierHolder(account.getId());
     }
-    
+
     // Provided to override in tests
     protected Account getAccount() {
         return Account.create();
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -412,17 +412,6 @@ public IdentifierHolder createParticipant(App app, StudyParticipant participant,
...
-        // Fix for callers that include only externalId and not the new externalId map: in these cases, tools are 
-        // creating the participant before the sign up call (through the Bridge Study Manager). Check and if 
-        // the account with this external ID already exists, return quietly. Otherwise proceed as before.
-        if (participant.getExternalId() != null && participant.getExternalIds().isEmpty()) {
-            AccountId accountId = AccountId.forExternalId(app.getIdentifier(), participant.getExternalId());
-            Account account = accountService.getAccount(accountId); 
-            if (account != null) {
-                return new IdentifierHolder(account.getId());
-            }
-        }
```
DwayneJengSage, line 424: Sometimes, integration tests need to create participants using External ID, to test the External ID only test cases. Since that bypasses SignUp, should this code remain in ParticipantService?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -559,18 +548,16 @@ private void updateAccountAndRoles(App app, Account account, StudyParticipant pa
...
+        // Legacy apps do this so we must continue to support it.
+        if (isNew) {
             for (Map.Entry<String, String> entry : participant.getExternalIds().entrySet()) {
                 String studyId = entry.getKey();
                 String externalId = entry.getValue();
                 
                 Enrollment enrollment = Enrollment.create(account.getAppId(), studyId, account.getId(), externalId);
-                enrollmentService.addEnrollment(account, enrollment);
+                enrollmentService.updateEnrollment(account, enrollment, enrollment);
+                account.getEnrollments().add(enrollment);
```
DwayneJengSage: EnrollmentService.addEnrollment() handles things like permissions checking and enrollment validation. Should we be bypassing those checks just to gain a bit of performance? Bypassing this layer of abstraction risks introducing bugs in the future by calling our internal API inconsistently.

Relatedly, should EnrollmentService.updateEnrollment actually be private?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: I believe we should switch to using the externalIds map rather than using this externalId field for the integration tests, and it appears we have since no tests failed after this change.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: A better solution is to acquire the account ID in the request context, and then call the original method which will then succeed, so I switched to doing that, and made the updateEnrollment method private. For the most part that was fine, I did have to remove a couple of assertions in the ParticipantService test, but that test shouldn't be testing the behavior of the EnrollmentService anyway.

DwayneJengSage (approved):

PR #343, opened 2021-03-18
Schedules v2

alxdarksage:
See: https://sagebionetworks.jira.com/wiki/spaces/BRIDGE/pages/914883048/Scheduling+API (probably not 100% accurate at this point)

And https://developer.sagebridge.org/articles/v2/scheduling.html which needs to be accurate by the time this is released to production, and is probably the closest at this point.

Scheduling epic: https://sagebionetworks.jira.com/browse/BRIDGE-2682


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSchedule2Dao.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public Schedule2 updateSchedule(Schedule2 schedule) {
+        checkNotNull(schedule);
+
+        Set<String> sessionGuids = schedule.getSessions().stream()
+                .map(Session::getGuid).collect(toSet());
+
+        hibernateHelper.executeWithExceptionHandling(schedule, (session) -> {
```
alxdarksage, line 116: Add a comment about why this code was necessary.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSchedule2Dao.java
```diff
@@ -0,0 +1,152 @@
...
+        
+        schedule.setDeleted(true);
+        hibernateHelper.update(schedule);
+    }
+
+    @Override
+    public void deleteSchedulePermanently(Schedule2 schedule) {
+        checkNotNull(schedule);
+        
+        hibernateHelper.executeWithExceptionHandling(schedule, (session) -> {
```
alxdarksage, line 149: Add comment here too about this as a workaround

src/main/java/org/sagebionetworks/bridge/hibernate/MySQLHibernatePersistenceExceptionConverter.java
```diff
@@ -0,0 +1,109 @@
...
+            if (rawMessage.contains("Duplicate entry")) {
+                displayMessage = selectMsg(
+                        rawMessage, UNIQUE_KEY_CONSTRAINTS, UNIQUE_CONSTRAINT_MSG, name, displayMessage);
+            } else if (rawMessage.contains("a foreign key constraint fails")) {
+                displayMessage = selectMsg(
+                        rawMessage, FOREIGN_KEY_CONSTRAINTS, FK_CONSTRAINT_MSG, name, displayMessage);
+            }
+            return new ConstraintViolationException.Builder().withMessage(displayMessage).build();
+        }
+        return exception;
```
alxdarksage, line 94: HibernateAssessment needs a @BridgeTypeName annotation, and other objects might as well.

PR #346, opened 2021-03-24
Fix the issue I found with resetting passwords sent via a phone number.

alxdarksage:
See: https://sagebionetworks.jira.com/browse/BRIDGE-2952


nategbrown9 (approved):
Looks good.

PR #347, opened 2021-03-25
Event customization

alxdarksage:
See:
https://sagebionetworks.jira.com/browse/BRIDGE-2271
https://sagebionetworks.jira.com/browse/BRIDGE-2935

This will be needed functionality for MTB as released in June.


src/main/java/org/sagebionetworks/bridge/dynamodb/MapMarshaller.java
```diff
@@ -0,0 +1,44 @@
...
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMappingException;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTypeConverter;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+
+import org.apache.commons.lang3.StringUtils;
+
+import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+
+public abstract class MapMarshaller<S,T> implements DynamoDBTypeConverter<String, Map<S, T>> {
```
DwayneJengSage: This class is a duplicate of https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/dynamodb/StringKeyMapMarshaller.java

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoApp.java
```diff
@@ -51,6 +54,15 @@
...
     }
     
+    public static class CustomEventsMarshaller extends MapMarshaller<String, ActivityEventUpdateType> {
+        private static final TypeReference<Map<String,ActivityEventUpdateType>> EVENT_MAP_TYPE = 
+                new TypeReference<Map<String,ActivityEventUpdateType>>() {};
+        @Override
+        public TypeReference<Map<String,ActivityEventUpdateType>> getTypeReference() {
+            return EVENT_MAP_TYPE;
+        }
+    }
```
DwayneJengSage: It's probably overkill to marshall this as JSON. DDB supports maps as a top level field, and Enums can be marshalled as strings.

We might need to write a custom Map-String-to-Enum Marshaller. See for example https://stackoverflow.com/questions/36495898/dynamodb-mapping-list-of-enum See also https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/dynamodb/EnumMarshaller.java

If for whatever reason we want to keep this as JSON, we can use StringKeyMarshaller instead. See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/dynamodb/OAuthProviderMapMarshaller.java

src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventUpdateType.java
```diff
@@ -0,0 +1,31 @@
...
+    FUTURE_ONLY;
+    
+    public boolean canDelete(ActivityEvent persistedEvent, ActivityEvent newEvent) {
+        return (persistedEvent != null && this == MUTABLE);
+    }
+    
+    public boolean canUpdate(ActivityEvent persistedEvent, ActivityEvent newEvent) {
+        return persistedEvent == null || this == MUTABLE ||
+            (this == FUTURE_ONLY && newEvent.getTimestamp() > persistedEvent.getTimestamp());
+    }
```
DwayneJengSage, line 30: Unit tests?

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -75,16 +79,17 @@ final void setAppService(AppService appService) {
...
-            throw new BadRequestException("App's ActivityEventKeys does not contain eventKey: " + eventKey);
+        if (!app.getCustomEvents().containsKey(eventKey) && !app.getAutomaticCustomEvents().containsKey(eventKey)) {
+            throw new BadRequestException("App's customEvents do not contain event ID: " + eventKey);
         }
         
         ActivityEvent event = new DynamoActivityEvent.Builder()
                 .withHealthCode(healthCode)
-                .withObjectType(ActivityEventObjectType.CUSTOM)
+                .withObjectType(CUSTOM)
+                .withUpdateType(app.getCustomEvents().get(eventKey))
```
DwayneJengSage, line 89: Should unit tests validate the correct update type?

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -98,20 +103,21 @@ public void publishCustomEvent(App app, String studyId, String healthCode, Strin
...
+        if (!app.getCustomEvents().containsKey(eventKey)
                 && !app.getAutomaticCustomEvents().containsKey(eventKey)) {
             throw new BadRequestException("App's ActivityEventKeys does not contain eventKey: " + eventKey);
         }
 
         ActivityEvent event = new DynamoActivityEvent.Builder()
                 .withHealthCode(healthCode)
-                .withObjectType(ActivityEventObjectType.CUSTOM)
+                .withObjectType(CUSTOM)
+                .withUpdateType(app.getCustomEvents().get(eventKey))
```
DwayneJengSage, line 114: Should unit tests validate the correct update type?

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -327,7 +338,8 @@ private void createAutomaticCustomEvents(App app, String studyId, String healthC
                 DateTime automaticEventTime = new DateTime(event.getTimestamp()).plus(automaticEventDelay);
                 ActivityEvent automaticEvent = new DynamoActivityEvent.Builder()
                         .withHealthCode(healthCode)
-                        .withObjectType(ActivityEventObjectType.CUSTOM)
+                        .withObjectType(CUSTOM)
+                        .withUpdateType(MUTABLE)
```
DwayneJengSage, line 342: Should unit tests validate the correct update type?

src/test/java/org/sagebionetworks/bridge/validators/ActivityEventValidatorTest.java
```diff
@@ -36,12 +38,14 @@ public void validates() {
...
+            expectedExceptionsMessageRegExp = ".*No update type configured.*")
     public void eventIdNull() {
         ActivityEvent event = getEvent()
                 .withObjectType(null)
                 .withObjectId(null)
                 .withEventType(null)
+                .withUpdateType(null)
                 .withAnswerValue(null)
                 .build();
         assertValidatorMessage(INSTANCE, event, "eventId", EVENT_ID_ERROR);
```
DwayneJengSage: This test originally tested if the eventId is null. Is this still possible? If so, the test should test that scenario.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/dynamodb/MapMarshaller.java
alxdarksage: Deleted, switched to use of StringKeyMapMarshaller

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoApp.java
alxdarksage: We could write a Map<String,String> <-> Map<String, Enum> but this seems like the same amount of work as storing JSON. The JSON is pretty similar to what Dynamo uses to serialize.

src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventUpdateType.java
alxdarksage: Added test of these methods directly.

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
alxdarksage: Added assertions for all the tests to verify the update types

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
alxdarksage: Added

src/test/java/org/sagebionetworks/bridge/validators/ActivityEventValidatorTest.java
alxdarksage: Yes you're right, this is a separate test. Added a test for a null ID back to the tests.

DwayneJengSage (approved):

PR #350, opened 2021-03-31
Timelines

alxdarksage:


src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
```diff
@@ -63,6 +63,17 @@ public AuthEvaluator canAccessStudy() {
...
+    /**
+     * The caller is enrolled in the target study.
+     */
+    public AuthEvaluator isEnrolledInStudy() {
+        predicates.add((factMap) -> {
+            String studyId = factMap.get(STUDY_ID);
+            return RequestContext.get().getCallerEnrolledStudies().contains(studyId);
+        });
+        return this;
+    }
```
alxdarksage: Can this be removed for now?

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -734,4 +735,28 @@ public static String formatActivityEventId(Set<String> activityEventIds, String
...
+        }
+        T englishItem = null;
+        for (T item : items) {
+            for (String lang : languages) {
+                if (lang.equalsIgnoreCase(item.getLang())) {
+                    return item;
+                }
+            }
+            if ("en".equalsIgnoreCase(item.getLang())) {
+                englishItem = item;
```
alxdarksage: So if my Langs were 'es' and 'de' and the items in order were 'de' and 'es', would this select es?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSchedule2Dao.java
```diff
@@ -98,61 +118,125 @@ final void setHibernateHelper(HibernateHelper hibernateHelper) {
...
+            session.setJdbcBatchSize(batchSize);
+
+            session.save(schedule);
+
+            if (!metadata.isEmpty()) {
+                Stopwatch stopwatch = Stopwatch.createStarted();
+                // Batch deleting/recreating rather than updating is 2-3 orders of magnitude faster.
+                NativeQuery<?> query = session.createNativeQuery(DELETE_TIMELINE_RECORDS);
+                query.setParameter(SCHEDULE_GUID, schedule.getGuid());
+                query.executeUpdate();
```
alxdarksage: You need to delete these records every time!

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSchedule2Dao.java
```diff
@@ -98,61 +118,125 @@ final void setHibernateHelper(HibernateHelper hibernateHelper) {
...
+
             session.update(schedule);
+
+            if (!metadata.isEmpty()) {
+                Stopwatch stopwatch = Stopwatch.createStarted();
+                
+                // Batch deleting/recreating rather than updating is 2-3 orders of magnitude faster.
+                query = session.createNativeQuery(DELETE_TIMELINE_RECORDS);
+                query.setParameter(SCHEDULE_GUID, schedule.getGuid());
+                query.executeUpdate();
```
alxdarksage: Delete outside the block, this is cleanup

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/AssessmentInfo.java
```diff
@@ -0,0 +1,95 @@
...
+        this.identifier = identifier;
+        this.revision = revision;
+        this.label = label;
+        this.colorScheme = colorScheme;
+        this.minutesToComplete = minutesToComplete;
+    }
+    
+    public String getKey() {
+        return String.valueOf(this.hashCode());
+    }
```
alxdarksage, line 83: Is this safe from conflicts? hashCode is not guaranteed to be unique.

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/Scheduler.java
```diff
@@ -0,0 +1,159 @@
...
+            scheduledSession.withEndDay(endDay);
+            scheduledSession.withStartTime(window.getStartTime());
+            scheduledSession.withExpiration(window.getExpiration());
+            scheduledSession.withPersistent(window.isPersistent());
+            // If we have a delay under one day, and it's the first day, then skip the startTime 
+            // and delay for the indicated time before showing.
+            if (startDay == 0 && delay != null && delay.toStandardDays().getDays() == 0) {
+                scheduledSession.withDelayTime(delay);
+                scheduledSession.withStartTime(null);
+            }
```
alxdarksage, line 116: This is ridiculously complex. I can't even describe it so what should we do here? Let's delete it.

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/Scheduler.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Create a unique GUID for a session instance that is the same each time you calculate a Timeline
+     * from a Schedule. It should look like a GUID and not a compound identifier, as client developers 
+     * have (in the past) parsed compound identifiers, and we want to discourage this.
+     */
+    String generateSessionInstanceGuid(String scheduleGuid, String sessionGuid, String windowGuid,
+            int windowOccurrence) {
+        return windowGuid.substring(0, 8) + Integer.toHexString(windowOccurrence) 
+            + sessionGuid.substring(0, 8) + scheduleGuid.substring(0, 6);
+    }
```
alxdarksage, line 159: Maybe make these a little longer to ensure you have the same amount of randomness.

src/main/java/org/sagebionetworks/bridge/services/Schedule2Service.java
```diff
@@ -49,7 +51,7 @@
     
     private OrganizationService organizationService;
     
-    private Schedule2Dao dao;
+    private Schedule2Dao scheduleDao;
```
alxdarksage: rename this back to DAO

src/main/resources/application.properties
```diff
@@ -7,3 +7,6 @@ spring.liquibase.change-log=classpath:db/changelog/changelog.sql
 
 # We don't ever use these, and these tend to leak email addresses and sign-in tokens.
 server.tomcat.accesslog.enabled=false
+
+hibernate.jdbc.batch_size = 100
+
```
alxdarksage: Remove this, it's in BridgeConfig

src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
alxdarksage: Added this as a test and adjusted code so it works.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateSchedule2Dao.java
alxdarksage: Removed the test for an empty collection, verified with manual testing this is not an issue, updated unit tests.

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/AssessmentInfo.java
alxdarksage: Replace it with a more robust hashing function.

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/Scheduler.java
alxdarksage: I left in the delayTime for this particular day, but also left in startTime, so it's now just an additional piece of information that the client can ignore or interpret as it will. In the simplest terms, if delayTime is there, the client should probably try and delay that amount of time from the event timestamp, unless the local time is such that this seems impractical. It could then use the startTime of the next day, or show the scheduled session immediately. I'll write this suggestion in the docs.

src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/Scheduler.java
alxdarksage: Converted to using 128 byte murmur3 algorithm, with Base 64 encoding to reduce the length of the result, across the full value of the GUIDs.

nategbrown9 (approved):
Looks good.

PR #351, opened 2021-04-02
Download participant worker changes

kllysng:


src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -121,6 +138,10 @@
     
     private EnrollmentService enrollmentService;
 
+    private BridgeConfig bridgeConfig;
+
+    private AmazonSQSClient sqsClient;
```
DwayneJengSage, line 143: Do we have setters for these?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -90,6 +99,14 @@
 @Component
 public class ParticipantService {
     private static final Logger LOG = LoggerFactory.getLogger(ParticipantService.class);
+    static final String CONFIG_KEY_DOWNLOAD_ROSTER_SQS_URL = "udd.sqs.queue.url"; //TODO this isn't right but need placeholder
```
DwayneJengSage: Should be workerPlatform.request.sqs.queue.url, and also needs to be added to the BridgeServer2.conf.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -824,6 +845,44 @@ public StudyParticipant updateIdentifiers(App app, CriteriaContext context, Iden
...
+        String regex = "^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)[a-zA-Z\\d]{8,}$";
+
+        Pattern pattern = Pattern.compile(regex);
+        Matcher matcher = pattern.matcher(password);
+
+        if (!matcher.matches()) {
+            throw new BadRequestException("Password must contain at least 8 characters: " +
+                    "1 upper case letter, 1 lower case letter, and 1 number.");
+        }
+    }
```
DwayneJengSage: Consider using ValidateUtils.validatePassword(). See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/validators/PasswordResetValidator.java#L45

You might want to pass in the ParticipantRosterRequest into the service and make a ParticipantRosterRequestValidator. See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/services/AuthenticationService.java#L395

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -602,6 +604,26 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
         participantService.sendSmsMessage(app, userId, template);
         return new StatusMessage("Message sent.");
     }
+
+    @PostMapping("/v3/participants/emailRoster")
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public StatusMessage getParticipantRosterForWorker() throws JsonProcessingException {
+        UserSession session = getAuthenticatedSession(WORKER);
```
DwayneJengSage: This wouldn't be called by a worker. This would be called by a researcher or study coordinator.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -602,6 +604,26 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
...
+        String appId = session.getAppId();
+
+        StudyParticipant participant = session.getParticipant();
+        if (!participant.getEmailVerified() && !participant.getPhoneVerified()) {
+            throw new BadRequestException("Cannot request user data");
+        }
```
DwayneJengSage: We don't need to check phoneVerified. We should check that participant has an email address and that email verified is true.

Additional checks
* If the caller is a Study Coordinator but not a Researcher (use RequestContext.get().isInRole()), the caller must have an orgMembership (RequestContext.get().getCallerOrgMembership()).
* If the caller specifies a study and has org membership, we need to verify that RequestContext.get().getOrgSponsoredStudies() contains that study.

src/main/java/org/sagebionetworks/bridge/spring/controllers/SponsorController.java
```diff
@@ -53,6 +54,18 @@
         
         return sponsorService.getSponsoredStudies(session.getAppId(), orgId, offsetByInt, pageSizeInt);
     }
+
+    @GetMapping("/v1/organizations/{orgId}/apps/{appId}/studies")
```
DwayneJengSage: appId should be before orgId

src/main/java/org/sagebionetworks/bridge/spring/controllers/SponsorController.java
```diff
@@ -53,6 +54,18 @@
...
+    public PagedResourceList<Study> getSponsoredStudiesForWorker(@PathVariable String appId, @PathVariable String orgId,
```
DwayneJengSage: Consider renaming this to getSponsoredStudiesForApp(), which better describes what it is used for.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
kellyyangsong: Do you mean something like so: config.key.download.roster.sqs.url = workerPlatform.request.sqs.queue.url?

src/main/resources/BridgeServer2.conf
```diff
@@ -146,6 +146,9 @@ dev.udd.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-
 uat.udd.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-WorkerPlatform-Request-uat
 prod.udd.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-WorkerPlatform-Request-prod
 
+#Bridge Participant Roster Download Worker SQS queues
+workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-local
```
DwayneJengSage: Should also specify for dev, uat, and prod. Check the AWS console to get the right URL.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -846,9 +862,10 @@ public StudyParticipant updateIdentifiers(App app, CriteriaContext context, Iden
         return getParticipant(app, account.getId(), false);
     }
 
-    public void downloadParticipantRoster(String appId, String userId, String password, String studyId) throws JsonProcessingException {
-        // password must contain at least 8 characters, 1 upper case, 1 lower case, and 1 number
-        checkPassword(password);
+    public void downloadParticipantRoster(String appId, String userId, String password, String studyId,
+                                          ParticipantRosterRequest request) throws JsonProcessingException {
+        validatePassword(password);
+        Validate.entityThrowingException(new ParticipantRosterRequestValidator(), request);
```
DwayneJengSage: We can create a static instance ParticipantRosterRequestValidator.INSTANCE so we don't have to create a new one with every request.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -866,22 +883,16 @@ public void downloadParticipantRoster(String appId, String userId, String passwo
...
-
-        if (!matcher.matches()) {
-            throw new BadRequestException("Password must contain at least 8 characters: " +
-                    "1 upper case letter, 1 lower case letter, and 1 number.");
-        }
+    private void validatePassword(String password) {
+        Errors error = Validate.getErrorsFor(password);
+        PasswordPolicy passwordPolicy = new PasswordPolicy(8, true, false, true, true);
+        ValidatorUtils.validatePassword(error, passwordPolicy, password);
     }
```
DwayneJengSage: You should move this to ParticipantRosterRequestValidator. ValidatorUtils.validatePassword() won't throw if there's an error.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -602,6 +605,34 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
         participantService.sendSmsMessage(app, userId, template);
         return new StatusMessage("Message sent.");
     }
+
+    @PostMapping("/v3/participants/emailRoster")
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public StatusMessage getParticipantRosterForWorker() throws JsonProcessingException {
```
DwayneJengSage: Minor: Rename this to getParticipantRoster.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -602,6 +605,34 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
...
+        UserSession session = getAuthenticatedSession(RESEARCHER, STUDY_COORDINATOR);
+        String appId = session.getAppId();
+
+        StudyParticipant participant = session.getParticipant();
+        if (participant.getEmail() != null && !participant.getEmailVerified()) {
```
DwayneJengSage: Should be participant.getEmail() == null || !participant.getEmailVerified()

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -608,19 +609,27 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
...
         String appId = session.getAppId();
 
         StudyParticipant participant = session.getParticipant();
-        if (!participant.getEmailVerified() && !participant.getPhoneVerified()) {
+        if (participant.getEmail() != null && !participant.getEmailVerified()) {
             throw new BadRequestException("Cannot request user data");
         }
+        if (RequestContext.get().isInRole(STUDY_COORDINATOR) && !RequestContext.get().isInRole(RESEARCHER)
+            && RequestContext.get().getCallerOrgMembership() == null) {
+            throw new BadRequestException("Caller is a study coordinator without an org membership.");
```
DwayneJengSage: This should probably be an UnauthorizedException.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -608,19 +609,27 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
...
+        }
 
         ParticipantRosterRequest request = parseJson(ParticipantRosterRequest.class);
         String password = request.getPassword();
         String studyId = request.getStudyId();
 
-        participantService.downloadParticipantRoster(appId, session.getId(), password, studyId);
+        if (studyId != null && !RequestContext.get().getOrgSponsoredStudies().contains(studyId)) {
+            throw new BadRequestException("Requested studyId is not sponsored by the caller's org.");
```
DwayneJengSage: This should also be an UnauthorizedException.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -846,9 +862,10 @@ public StudyParticipant updateIdentifiers(App app, CriteriaContext context, Iden
         return getParticipant(app, account.getId(), false);
     }
 
-    public void downloadParticipantRoster(String appId, String userId, String password, String studyId) throws JsonProcessingException {
-        // password must contain at least 8 characters, 1 upper case, 1 lower case, and 1 number
-        checkPassword(password);
+    public void downloadParticipantRoster(String appId, String userId, String password, String studyId,
+                                          ParticipantRosterRequest request) throws JsonProcessingException {
```
DwayneJengSage: We should probably take in just ParticipantRosterRequest, not password or studyId. This way, we can't accidentally have studyId or password be different from request.

src/main/java/org/sagebionetworks/bridge/validators/ParticipantRosterRequestValidator.java
```diff
@@ -0,0 +1,37 @@
...
+            errors.rejectValue("ParticipantRosterRequest", Validate.WRONG_TYPE);
+        } else {
+            ParticipantRosterRequest request = (ParticipantRosterRequest) target;
+
+            // password
+            if (request.getPassword() == null) {
+                errors.rejectValue("password", Validate.CANNOT_BE_NULL);
+            } else if (request.getPassword().isEmpty()) {
+                errors.rejectValue("password", Validate.CANNOT_BE_BLANK);
+            }
```
DwayneJengSage: You can use org.apache.commons.lang3.StringUtils.isBlank().

DwayneJengSage (changes requested):

src/main/resources/BridgeServer2.conf
kllysng: Maybe I don't have access to the uat and prod queues - the only WorkerPlatform queues I see end in alxdark, develop, dwanejeng, local, and staging. There are also some dead letter queues.

kllysng: However it seems that the urls differ by just the "-develop" or "-local" at the end. I'll put in some place holders and could you verify if they match up with the queues you see on your end?

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
kllysng: wow interesting I learned something new!

src/main/resources/BridgeServer2.conf
```diff
@@ -147,7 +147,10 @@ uat.udd.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-
 prod.udd.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-WorkerPlatform-Request-prod
 
 #Bridge Participant Roster Download Worker SQS queues
-workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-local
+local.workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-local
+dev.workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-develop
+uat.workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-uat
+prod.workerPlatform.request.sqs.queue.url = https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-prod
```
DwayneJengSage: uat should be https://sqs.us-east-1.amazonaws.com/420786776710/Bridge-WorkerPlatform-Request-staging

prod should be https://sqs.us-east-1.amazonaws.com/649232250620/Bridge-WorkerPlatform-Request-prod

src/main/java/org/sagebionetworks/bridge/validators/ParticipantRosterRequestValidator.java
```diff
@@ -34,4 +41,10 @@ public void validate(Object target, Errors errors) {
             }
         }
     }
+
+    private void validatePassword(String password) {
+        Errors error = Validate.getErrorsFor(password);
+        PasswordPolicy passwordPolicy = new PasswordPolicy(8, true, false, true, true);
+        ValidatorUtils.validatePassword(error, passwordPolicy, password);
+    }
```
DwayneJengSage: This code doesn't actually do anything if the password fails to validate. You should instead pass in the errors from validate().

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantController.java
```diff
@@ -608,28 +608,27 @@ public StatusMessage sendSmsMessageForWorker(@PathVariable String appId, @PathVa
...
     @PostMapping("/v3/participants/emailRoster")
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public StatusMessage getParticipantRosterForWorker() throws JsonProcessingException {
+    public StatusMessage getParticipantRoster() throws JsonProcessingException {
         UserSession session = getAuthenticatedSession(RESEARCHER, STUDY_COORDINATOR);
         String appId = session.getAppId();
 
         StudyParticipant participant = session.getParticipant();
-        if (participant.getEmail() != null && !participant.getEmailVerified()) {
+        if (participant.getEmail() != null || !participant.getEmailVerified()) {
```
DwayneJengSage: Should be participant.getEmail() == null

DwayneJengSage (changes requested):
Also, we need unit tests.

src/main/java/org/sagebionetworks/bridge/models/ParticipantRosterRequest.java
```diff
@@ -0,0 +1,54 @@
...
+        private String studyId;
+
+        /** @see ParticipantRosterRequest#getPassword() */
+        public ParticipantRosterRequest.Builder withPassword(String password) {
+            this.password = password;
+            return this;
+        }
+
+        /** @see ParticipantRosterRequest#getStudyId() */
+        @JsonDeserialize(using = DateTimeDeserializer.class)
```
DwayneJengSage: Looks like a copy-paste error here. We don't want to deserialize this as a DateTime. (Does Jackson not throw in this case?)

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -824,6 +854,30 @@ public StudyParticipant updateIdentifiers(App app, CriteriaContext context, Iden
         // return updated StudyParticipant to update and return session
         return getParticipant(app, account.getId(), false);
     }
+
+    public void getParticipantRoster(String appId, String userId, ParticipantRosterRequest request) throws JsonProcessingException {
+        Validate.entityThrowingException(ParticipantRosterRequestValidator.INSTANCE, request);
+
+        ObjectMapper jsonObjectMapper = new ObjectMapper();
```
DwayneJengSage: Should use BridgeObjectMapper.get() for consistency.

src/main/java/org/sagebionetworks/bridge/spring/controllers/SponsorController.java
```diff
@@ -53,6 +54,18 @@
...
+    public PagedResourceList<Study> getSponsoredStudiesForApp(@PathVariable String appId, @PathVariable String orgId,
+                                                                 @RequestParam(required = false) String offsetBy,
+                                                                 @RequestParam(required = false) String pageSize) {
+        getAuthenticatedSession(WORKER);
+
+        int offsetByInt = BridgeUtils.getIntOrDefault(offsetBy, 0);
+        int pageSizeInt = BridgeUtils.getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+
+        return sponsorService.getSponsoredStudies(appId, orgId, offsetByInt, pageSizeInt);
+    }
```
DwayneJengSage, line 68: Unit tests?

DwayneJengSage (changes requested):
Mostly good. A few more comments.

src/main/java/org/sagebionetworks/bridge/models/ParticipantRosterRequest.java
kllysng: Yes, this was a copy pasta! Thank you for seeing this. No, there was no error thrown.

DwayneJengSage (approved):

PR #352, opened 2021-04-06
Move audit-like code somewhere more accessible to other classes

alxdarksage:
which will do something similar under the current design plan. Arguably this simple code could be useful even if we create an audit table, to look up account information for a user ID (which would be what is in the audit table).


alxdarksage:
I explored a few ways of cleaning this up that are better than this, relying more on Hibernate to do the work. Closing for now.

PR #353, opened 2021-04-08
Study updates

alxdarksage:
- added a number of fields to Study object, including a collection of Contact objects
- very minimal enforcement of phase: existing studies have "legacy," new studies are in "design," and you can't currently change this through an update;
- added some fields to the summary version of a study (createdOn, modifiedOn, phase);
- remove the study persistence exception handler and use the more robust MySQL exception handler, moving over the foreign key constraint and the tests for it.


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
```diff
@@ -40,11 +40,11 @@ public HibernateHelper(SessionFactory hibernateSessionFactory, PersistenceExcept
...
     public <T> void deleteById(Class<T> clazz, Serializable id) {
+        T obj = getById(clazz, id);
         // Hibernate optimistic versioning also applies to deletes. However, unlike updates, when we delete something,
         // we want it gone, so we generally don't care about optimistic versioning. In order to handle this in
         // Hibernate, we need to load the whole object before deleting it.
-        executeWithExceptionHandling(null, session -> {
-            T obj = session.get(clazz, id);
+        executeWithExceptionHandling(obj, session -> {
             session.delete(obj);
             return null;
```
alxdarksage, line 49: Did you write a test for this?

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -21,15 +27,37 @@ public void validate(Object object, Errors errors) {
         Study study = (Study)object;
         
         if (isBlank(study.getIdentifier())) {
-            errors.rejectValue("id", "is required");
-        } else if (!study.getIdentifier().matches(BridgeConstants.BRIDGE_EVENT_ID_PATTERN)) {
-            errors.rejectValue("id", BridgeConstants.BRIDGE_EVENT_ID_ERROR);
+            errors.rejectValue("identifier", "is required");
+        } else if (!study.getIdentifier().matches(BRIDGE_EVENT_ID_PATTERN)) {
+            errors.rejectValue("identifier", BRIDGE_EVENT_ID_ERROR);
```
alxdarksage, line 32: Make sure BSM is showing error now that identifier has changed name!

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -320,13 +326,16 @@ public void updateStudy() {
...
         assertNotNull(returnedValue.getModifiedOn());
     }
     
-    @Test(expectedExceptions = InvalidEntityException.class)
+    @Test(expectedExceptions = EntityNotFoundException.class)
     public void updateStudyInvalidStudy() {
-        service.updateStudy(TEST_APP_ID, Study.create());
+        Study study = Study.create();
+        study.setIdentifier("oneId");
+        service.updateStudy(TEST_APP_ID, study);
```
alxdarksage, line 338: So, why did this change? Is this correct?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateHelper.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
alxdarksage: Yes, it does.

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
alxdarksage: Phase is required, so to ensure this doesn't trigger an NPE, we set it from the existing object before validation. But this requires having an identifier, which we virtually always will now that the controller sets it from the path (which is our default). So the ID is required and the exception thrown has changed.

nategbrown9 (approved):
Looks good.

PR #354, opened 2021-04-09
Adding validation to prevent no duration periods (e.g. “PT0S”)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/validators/ValidatorUtils.java
```diff
@@ -206,13 +225,18 @@ private static void validateDuration(Set<DurationFieldType> durations, Errors er
...
         int[] values = period.getValues();
         for (int i=0; i < values.length; i++) {
             if (values[i] < 0) {
                 errors.rejectValue(fieldName, CANNOT_BE_NEGATIVE);
                 break;
             }
         }
+        if (periodInMilliseconds(period) == 0L) {
+            errors.rejectValue(fieldName, "cannot be of no duration");
+        }
```
alxdarksage, line 239: We don't want them to do things like "PT180S" (3 minutes), so *any* value outside of the allowable fields should be prohibited, which would catch 0S (but also check for zero values).

alxdarksage: Joda serializes a zero-duration duration (no matter how it was initially serialized) as "PT0S." And, the Joda APIs don't differentiate between 0 as in "this field wasn't specified in the string" and 0 as in "this field was specified to be zero." So the current approach is necessary to catch everything.

nategbrown9 (approved):
Looks good.

PR #355, opened 2021-04-13
Adding startEventId and time window GUID to the metadata record

alxdarksage:
In addition, fixed a bug where a couple of session fields were not being copied over to the timeline metadata record, and added a method to return the record as a map. Added a test for these fields and the map. 

We're going to take the map, add some user-specific metadata, and attach it to the health data record in the v3 version of the exporter. Along with metadata submitted by the client, that's the tagging of the exported file that researchers will have available to search for/filter/analyze their data.


nategbrown9 (approved):
Looks good.

PR #356, opened 2021-04-20
Connect schedule to study

alxdarksage:
A couple of endpoints to retrieve a timeline for a user (either the user retrieving their own timeline, or retrieving a user's timeline if you're an administrative user). Currently these are the same because all users in a study are given the same schedule...but this will change without changing these endpoints.


nategbrown9 (approved):
Looks good.

PR #357, opened 2021-04-23
Fix case where a supplemental consent does not enroll a participant in a study.

alxdarksage:
JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-2974


nategbrown9 (approved):
Looks good.

PR #358, opened 2021-04-24
Upload API to include metadata as a top-level field

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2607


alxdarksage:
At some point in the process do we want to type the metadata as a JsonObject rather than JsonNode? We're going to need to hook in a call to add scheduling metadata, if it exists for the upload. The client will need to include some known properties in their JSON (instanceGuid and eventTimestamp) if the scheduling metadata is to be added to the JSON metadata, so I think we'll want to assume at the top level that the metadata consists of some key/value pairs. Is this not currently true for some apps? The API also won't provide any clue to API consumers that these properties exist and are special. Should they be pulled out alongside metadata? Would scheduling metadata be added to the JSON blob...that makes sense to me but again, we'd have to assume it's an object to add further information safely.

src/main/java/org/sagebionetworks/bridge/models/upload/Upload.java
```diff
@@ -44,6 +45,13 @@ static Upload create() {
     /** @see #getHealthCode */
     void setHealthCode(String healthCode);
 
+    /**
+     * Metadata fields for this upload, as submitted by the app. This corresponds with the
+     * uploadMetadataFieldDefinitions configured in the app.
+     */
+    JsonNode getMetadata();
+    void setMetadata(JsonNode metadata);
+
```
alxdarksage, line 54: However, we're not validating these align at the time an upload request is received...?

DwayneJengSage:
Changed JsonNode to ObjectNode in both UploadRequest and Upload. (Additional commits coming shortly.)

For validation, I don't think it makes sense to add validation. In the general case (for both Exporter 2.0 and Exporter 3.0), Bridge doesn't have any semantic knowledge of what the metadata "means", so having Bridge validate usually causes a lot of trouble. Additionally, in the past, when Bridge did a lot more validation, this caused a lot of frustration for the researchers because Bridge would do the validation "wrong", and we'd have to go back and fix it. Going forward, I think it's better for Bridge to treat metadata s a black box and just export it straight across to Synapse. (Incidentally, Bridge Exporter currently *doesn't* do this. I'm making a change to Bridge Exporter to fix this.)

On top of that, we don't want to validate the metadata fields against the pre-defined metadata fields, because these are going away in Exporter 3.0. (Also, in Exporter 2.0, we didn't validate them anyway, and for the reasons I mentioned above, I don't think we should start now.)

As for making fields like instanceGuid and timestamp top-level fields, I have no strong opinions about that, but until the scheduling stuff is more fleshed out, I think it's premature to add that in.

alxdarksage:
I think we will just check this metadata object for particular fields, and if they are present, add some further metadata from the scheduling system. So I wouldn't necessarily call this validation...if the fields are not there we'll just silently skip this step, so you can use the upload system without knowing about the scheduling system at all.

alxdarksage (approved):

alxdarksage (approved):

PR #366, opened 2021-05-11
Activity event history

alxdarksage:
For compliance, we must maintain a history of timestamps for activity events that are not immutable. This cannot be retrofitted into the existing model without scalability issues, so I am reimplementing activity events for studies in a manner that will allow for full history of timestamps, as well as some other information that the simpler first implementation could not store (such as the time zone of the event for display back to participants at a later time).

This PR is also adding appId to AdherenceRecords...I became concerned that the foreign key constraint on studies did not include the study's entire primary key. As a result, it worried me that deleting a study in one app, could delete the adherence records for similarly-ID'ed studies in other apps. App is not part of the foreign key constraint.

Finally, to unify processing of ActivityEvent and StudyActivityEvent, I made ActivityEvent's timestamp field a DateTime rather than a Long. It should have always been a DateTime record...this has no impact on the API or Dynamo DB.


src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
```diff
@@ -0,0 +1,213 @@
...
+        checkNotNull(studyId);
+
+        Account account = accountService.getAccountNoFilter(AccountId.forId(appId, userId))
+                .orElseThrow(() -> new EntityNotFoundException(Account.class));
+
+        List<StudyActivityEvent> events = dao.getRecentStudyActivityEvents(userId, studyId);
+        events.add(new StudyActivityEvent(CREATED_ON_FIELD, account.getCreatedOn()));
+        events.add(makeStudyStartDate(events, account.getCreatedOn()));
+        
+        return new ResourceList<>(events);
```
alxdarksage, line 133: Do we need the created_on timestamp to be written back to the database?
Also talk to Dwayne about study_start_date.

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
```diff
@@ -0,0 +1,213 @@
...
+        if (pageSize != null && (pageSize < API_MINIMUM_PAGE_SIZE || pageSize > API_MAXIMUM_PAGE_SIZE)) {
+            throw new BadRequestException(PAGE_SIZE_ERROR);
+        }
+        String appId = request.getAppId();
+        request.customEvents(appService.getApp(appId).getCustomEvents());
+        request.autoCustomEvents(appService.getApp(appId).getAutomaticCustomEvents());
+        
+        String userId = request.getUserId();
+        String studyId = request.getStudyId();
+        String eventId = request.getObjectId();
```
alxdarksage, line 152: So, if this is invalid and therefore null, wouldn't you get NPEs below?

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantController.java
```diff
@@ -142,11 +149,15 @@ final void setSchedule2Service(Schedule2Service scheduleService) {
...
         Schedule2 schedule = scheduleService.getScheduleForStudy(session.getAppId(), study);
         cacheProvider.setObject(scheduleModificationTimestamp(studyId), schedule.getModifiedOn().toString());
         
-        activityEventService.publishActivitiesRetrieved(app, studyId, session.getHealthCode(), DateTime.now());
+        studyActivityEventService.publishEvent(new StudyActivityEventRequest()
+                .appId(session.getAppId())
+                .studyId(studyId)
+                .userId(session.getId())
+                .objectType(TIMELINE_RETRIEVED)
+                .timestamp(getDateTime()));
```
alxdarksage, line 160: What about first sign in as an event?

alxdarksage: Consider removing study_start_date since it is in fact mutable.

This scenario is problematic: download events, then download timeline, and the study_start_date will be different because timeline was retrieved (all of a sudden, order of operations matter).

alxdarksage: Another option: publish this if you retrieve activity events (and include it in that map) and/or when you retrieve the timeline, so they are not out-of-sync.

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
alxdarksage: Answer: no, the APIs point at the global system and I haven't removed anything there about persisting a created_on timestamp from the Account.

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
alxdarksage: Changed this so it throws BadRequestException if the eventID is nulled out.

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyParticipantController.java
alxdarksage: study_start_date removed. timeline_retrieved added to both the study activities and timeline endpoints. first sign in I am not adding because there's nothing about it that's specific to any study. We'd have to record it as a global event, then retrieve it and include it in the study-specific events...enrollment or timeline_retrieved would be better.

nategbrown9 (approved):

PR #372, opened 2021-05-20
Small changes to scheduling

alxdarksage:
1) Time window GUIDs are added to the timeline. Knowing these is very helpful when visualizing the timeline in certain ways, since it allows you to regroup timeline elements into separate streams;

2) Added configURL to assessment info: instead of documenting how this should be calculated, it's now just available and adjusts for whether or not the assessment is a shared assessment, or study specific;

3) Sessions in the timeline now appear in the same order they appear in the schedule. Again for certain UI visualizations, this is helpful.


nategbrown9 (approved):
Looks good.

PR #374, opened 2021-05-25
Reimplement notifications

alxdarksage:


src/main/java/org/sagebionetworks/bridge/validators/SessionValidator.java
```diff
@@ -32,72 +88,169 @@ public void validate(Object obj, Errors errors) {
...
+            errors.pushNestedPath(NOTIFICATIONS_FIELD+"["+i+"]");
+            
+            if (notification.getNotifyAt() == null) {
+                errors.rejectValue(NOTIFY_AT_FIELD, CANNOT_BE_NULL);
+            }
+            validateFixedLengthPeriod(errors, notification.getOffset(), OFFSET_FIELD, true);
+            validateFixedLengthPeriod(errors, notification.getInterval(), INTERVAL_FIELD, false);
+            if (!session.getTimeWindows().isEmpty()) {
+                TimeWindow shortestWindow = shortestTimeWindow(session);
+                if (shortestWindow.getExpiration() != null) {
```
alxdarksage, line 176: We might have to check all of these, since it skips this validation if the window does not expire, which means the next longest window now becomes an issue and should be compared against the notification time.

nategbrown9 (approved):
Looks good.

PR #375, opened 2021-05-25
Health Data Documentation

kllysng:
I still have to implement unit tests but here is what I have on Health Data Documentation. Thank you!


src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentation.java
```diff
@@ -0,0 +1,115 @@
...
+public class DynamoHealthDataDocumentation implements HealthDataDocumentation {
+    String title;
+    String parentId;
+    String identifier;
+    Long version;
+    String documentation;
+    String createdBy;
+    DateTime createdOn;
+    String modifiedBy;
+    DateTime modifiedOn;
```
DwayneJengSage: Minor: These should all be marked as private.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentation.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public DateTime getCreatedOn() {
+        return this.createdOn;
+    }
+
+    @Override
+    public void setCreatedOn(DateTime createdOn) {
+        this.createdOn = createdOn;
+    }
```
DwayneJengSage: For createdOn and modifiedOn, these are generated by the server, so timezone information doesn't matter. With that in mind, I think it would be better to use a Long (epoch milliseconds) instead of a DateTime. (Capital L Long, so that it's nullable.)

To parse this from an ISO timestamp, you'll need the following annotations on getCreatedOn() and getModifiedOn()

```
@JsonInclude(JsonInclude.Include.NON_DEFAULT)
@JsonSerialize(using = DateTimeToLongSerializer.class)
```

JsonInclude.NON_DEFAULT ignores the default values (namely 0 and null), which avoids very strange bugs. DateTimeToLongSerializer converts the epoch milliseconds to an ISO DateTime.

Similarly, you'll need to add the following annotation to setCreatedOn() and setModifiedOn()

```
@JsonDeserialize(using = DateTimeToLongDeserializer.class)
```

This does the inverse of the serializer.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+        HealthDataDocumentation documentation = parseJson(HealthDataDocumentation.class);
+
+        if (documentation.getDocumentation() == null) {
+            throw new BadRequestException("Documentation is required to store health data documentation.");
+        }
+
+        // check documentation size (max allowed is 100 KB)
+        if (documentation.getDocumentation().getBytes(Charset.defaultCharset()).length > MAX_DOCUMENTATION_BYTES) {
+            throw new BadRequestException("Documentation must be less than 100 KB");
+        }
```
DwayneJengSage: We move this logic into a HealthDataDocumentationValidator and have the HealthDataDocumentationService call the validator.

Also, since DynamoDB uses UTF-8 for calculating string sizes (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-items), we should use UTF-8, not defaultCharset().

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+
+        // update health data documentation attributes
+        if (documentation.getCreatedOn() == null) {
+            documentation.setCreatedOn(DateTime.now());
+            documentation.setCreatedBy(session.getId());
+        } else {
+            documentation.setModifiedOn(DateTime.now());
+            documentation.setModifiedBy(session.getId());
+        }
```
DwayneJengSage: Similarly, this should also be moved to the HealthDataDocumentationService.

In the Service, the userId can be gotten from RequestContext.getCallerUserId().

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentation.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    @DynamoDBHashKey
+    @Override
+    public String getParentId() {
+        return this.parentId;
+    }
```
DwayneJengSage: It might be worth putting `@JsonIgnore` on getParentId(), since this won't be useful in our external API. Also, callers won't be able to set it; it will be enforced through our API (eg, we'll likely have separate APIs for study documentation and assessment documentation, and we'll be enforcing the parent ID at that level).

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+
+        return healthDataDocumentationService.createOrUpdateHealthDataDocumentation(documentation);
```
DwayneJengSage, line 40: It might be worth setting the parentId to the appId here, so that callers can't use this API to create documentation for parents they don't have access to.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+        // and for Assessments it would be “{appId}:assessment:{assessmentId}”.
+        String parentId = session.getAppId();
+        return healthDataDocumentationService.getHealthDataDocumentationForId(identifier, parentId);
+    }
+
+    /** Get all health data documentation with the given parentId. */
+    @GetMapping(path="/v3/healthdataDocumentation")
+    public ForwardCursorPagedResourceList<HealthDataDocumentation> getAllHealthDataDocumentationForParentId(@RequestParam String parentId,
+        @RequestParam(required = false) String pageSize, @RequestParam(required = false) String offsetKey) {
+        getAuthenticatedSession(Roles.RESEARCHER, Roles.DEVELOPER);
```
DwayneJengSage: I think instead of taking in the parentId as an input, we should just use the appId from the session.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+        String parentId = session.getAppId(); // placeholder
+        healthDataDocumentationService.deleteHealthDataDocumentation(identifier, parentId);
+
+        return new StatusMessage("Health data documentation has been deleted for the given identifier.");
+    }
+
+    /** Delete all health data documentation with the given parentId. */
+    @DeleteMapping(path="/v3/healthdataDocumentation")
+    public StatusMessage deleteAllHealthDataDocumentationForParentId(@RequestParam String parentId) {
+        getAuthenticatedSession(Roles.RESEARCHER, Roles.DEVELOPER);
```
DwayneJengSage: Similarly, we should use the appId here.

src/test/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationControllerTest.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Test
+    public void getAllHealthDocumentationForParentId() {
+        ForwardCursorPagedResourceList<HealthDataDocumentation> docList = new ForwardCursorPagedResourceList<>(
+                ImmutableList.of(HealthDataDocumentation.create()), null);
+        when(mockService.getAllHealthDataDocumentation(TEST_APP_ID, API_DEFAULT_PAGE_SIZE, OFFSET_KEY)).thenReturn(docList);
+
+        ForwardCursorPagedResourceList<HealthDataDocumentation> resultList =
+                controller.getAllHealthDataDocumentationForParentId(TEST_APP_ID, "50", OFFSET_KEY);
```
DwayneJengSage: This looks a bit confusing. We should use a different page size than the default so we can verify that the page size parsing is working and not just always using the default.

src/main/java/org/sagebionetworks/bridge/services/HealthDataDocumentationService.java
```diff
@@ -0,0 +1,80 @@
...
+
+        healthDataDocumentationDao.deleteDocumentationForIdentifier(identifier, parentId);
+    }
+
+    /** Delete all health data documentation for the given parentId */
+    public void deleteAllHealthDataDocumentation(String parentId) {
+        checkNotNull(parentId);
+
+        healthDataDocumentationDao.deleteDocumentationForParentId(parentId);
+    }
```
DwayneJengSage: Missing unit tests for deleteAllHealthDataDocumentation()

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDao.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.annotation.Resource;
+import java.util.List;
+
+@Component
+public class DynamoHealthDataDocumentationDao implements HealthDataDocumentationDao {
+    private DynamoDBMapper mapper;
+
+    /** DynamoDB mapper for the HealthDataDocumentation table, configured by Spring. */
+    @Resource(name = "healthDataDocumentationDbMapper")
+    public void setMapper(DynamoDBMapper mapper) {
```
DwayneJengSage: Minor: Set this to final.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
DwayneJengSage: Thinking about this some more, I think checking if createdOn is null is probably not a good way to check if this is a create or an update, since caller can specify any arbitrary value for createdOn.

I'd suggest either merging the createdOn/modifiedOn and createdBy/modifiedBy to a single set of attributes (maybe just modifiedOn and modifiedBy). Either that or splitting the APIs to a distinct create() and update() APIs.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDao.java
```diff
@@ -0,0 +1,132 @@
...
+        if (dynamoDocumentation.getIdentifier() == null) {
+            // Documentation doesn't have ID assigned yet (new documentation). Create ID and assign it.
+            dynamoDocumentation.setIdentifier(BridgeUtils.generateGuid());
+
+            // Update created on attributes.
+            dynamoDocumentation.setCreatedOn(DateTime.now());
+        } else {
+            // Update modified on attributes.
+            dynamoDocumentation.setModifiedOn(DateTime.now());
+        }
```
DwayneJengSage: I think it would be better to allow the caller to choose the identifier instead of randomly generating one.

Also, the logic for setting createdOn/modifiedOn is duplicated with the same logic in the Controller (which should be moved to the Service).

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDaoTest.java
```diff
@@ -0,0 +1,182 @@
...
+        List<DynamoHealthDataDocumentation> docList = ImmutableList.of(new DynamoHealthDataDocumentation());
+        doReturn(docList).when(dao).queryHelper(any());
+
+        // execute
+        dao.deleteDocumentationForParentId(TEST_APP_ID);
+
+        // validate
+        ArgumentCaptor<DynamoDBQueryExpression<DynamoHealthDataDocumentation>> queryCaptor = ArgumentCaptor.forClass(
+                DynamoDBQueryExpression.class);
+        verify(dao).queryHelper(queryCaptor.capture());
```
DwayneJengSage, line 85: We should also verify(mapper).batchDelete()

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDao.java
```diff
@@ -0,0 +1,132 @@
...
+    public void deleteDocumentationForIdentifier(@Nonnull String identifier, @Nonnull String parentId) {
+        DynamoHealthDataDocumentation hashKey = new DynamoHealthDataDocumentation();
+        hashKey.setParentId(parentId);
+        hashKey.setIdentifier(identifier);
+
+        DynamoHealthDataDocumentation documentationToDelete = mapper.load(hashKey);
+        if (documentationToDelete != null) {
+            mapper.delete(documentationToDelete);
+        }
+    }
```
DwayneJengSage: Missing unit tests.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDaoTest.java
```diff
@@ -0,0 +1,182 @@
...
+        dao.deleteDocumentationForParentId(TEST_APP_ID);
+
+        // query is validated, just validate that we don't call batchDelete()
+        verify(mockMapper, never()).batchDelete(anyList());
+    }
+
+    @Test
+    public void getDocumentationForIdentifier() {
+        DynamoHealthDataDocumentation doc = new DynamoHealthDataDocumentation();
+        when(mockMapper.load(any(DynamoHealthDataDocumentation.class))).thenReturn(doc);
```
DwayneJengSage, line 136: We should use an ArgumentCaptor and verify the doc ID and parent ID.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDao.java
```diff
@@ -0,0 +1,132 @@
...
+        // Batch delete.
+        if (!docsToDelete.isEmpty()) {
+            List<DynamoDBMapper.FailedBatch> failures = mapper.batchDelete(docsToDelete);
+            BridgeUtils.ifFailuresThrowException(failures);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public void deleteDocumentationForIdentifier(@Nonnull String identifier, @Nonnull String parentId) {
```
DwayneJengSage: Minor: The order of arguments should be parentId, identifier. The general convention is we use the more general category before the more specific category. (At least when there is a clear hierarchy.)

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationController.java
```diff
@@ -0,0 +1,106 @@
...
+
+        int pageSizeInt = BridgeUtils.getIntOrDefault(pageSize, API_DEFAULT_PAGE_SIZE);
+
+        return healthDataDocumentationService.getAllHealthDataDocumentation(parentId, pageSizeInt, offsetKey);
+    }
+
+    /** Delete a health data documentation with the given identifier. */
+    @DeleteMapping(path="v3/healthdataDocumentation/{identifier}")
+    public StatusMessage deleteHealthDataDocumentationForIdentifier(@PathVariable String identifier) {
+        UserSession session = getAuthenticatedSession(Roles.RESEARCHER, Roles.DEVELOPER);
```
DwayneJengSage: Thinking through this some more, I think the delete APIs should be admin only. This way, you can't accidentally delete someone else's documentation, or worse, all the documentation in an app.

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDaoTest.java
kllysng: I'm not sure what you mean here. Do we need an ArgumentCaptor<String> to verify that hashkey.setIdentifier() and hashkey.setParentId() are called? I'm not sure how to verify what hashkey calls inside the dao function.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentation.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public void setModifiedBy(String modifiedBy) {
+        this.modifiedBy = modifiedBy;
+    }
+
+    @Override
+    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+    public Long getModifiedOn() {
+        return this.modifiedOn;
+    }
```
DwayneJengSage: This is stored internally as a Long, but we want the JSON to be serialized as an ISO8601 string (YYYY-MM-DDThh:mm:ss.sssZ). This can be accomplished by annotating getModifiedOn() with 

```
@JsonSerialize(using = DateTimeToLongSerializer.class)
```

and setModifiedOn() with

```
@JsonDeserialize(using = DateTimeToLongDeserializer.class)
```

src/test/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationTest.java
```diff
@@ -0,0 +1,41 @@
...
+        documentation.setParentId(TEST_APP_ID);
+        documentation.setIdentifier(IDENTIFIER);
+        documentation.setModifiedOn(TIMESTAMP);
+        documentation.setVersion(VERSION);
+
+        String json = MAPPER.writeValueAsString(documentation);
+
+        JsonNode node = MAPPER.readTree(json);
+        assertEquals(node.get("identifier").textValue(), IDENTIFIER);
+        assertEquals(Long.valueOf(node.get("modifiedOn").longValue()), TIMESTAMP);
```
DwayneJengSage: For modifiedOn, we want this to serialize to an ISO8601 string (YYYY-MM-DDThh:mm:ss.sssZ). If you have trouble verifying this string, you could parse it back to a DateTime and use .getMillis() to check that the timestamp is the same.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoHealthDataDocumentationDao.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public HealthDataDocumentation createOrUpdateDocumentation(@Nonnull HealthDataDocumentation documentation) {
+        HealthDataDocumentation dynamoDocumentation = (DynamoHealthDataDocumentation) documentation;
+
+        if (dynamoDocumentation.getIdentifier() == null) {
+            // Documentation doesn't have ID assigned yet (new documentation). Create ID and assign it.
+            dynamoDocumentation.setIdentifier(BridgeUtils.generateGuid());
+        }
```
DwayneJengSage: If we're requiring identifier in the HealthDataDocumentationValidator, we don't need to check for null identifier in the DAO.

src/test/java/org/sagebionetworks/bridge/services/HealthDataDocumentationServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @Test
+    public void createOrUpdateHealthDataDocumentation() {
+        HealthDataDocumentation doc = makeValidDoc();
+        when(mockDao.createOrUpdateDocumentation(doc)).thenReturn(doc);
+
+        HealthDataDocumentation result = service.createOrUpdateHealthDataDocumentation(doc);
+        assertSame(result, doc);
+
+        verify(mockDao).createOrUpdateDocumentation(doc);
+    }
```
DwayneJengSage: We should also verify that modifiedOn and modifiedBy to the result.

You can mock DateTime.now() by calling DateTimeUtils.setCurrentMillisFixed(). After the test, you should call DateTimeUtils.setCurrentMillisSystem() to reset the clock. See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/test/java/org/sagebionetworks/bridge/upload/InitRecordHandlerTest.java#L43

For RequestContext, you call RequestContext.set(). After the test, you can RequestContext.set(null) to reset the RequestContext. See for example https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/test/java/org/sagebionetworks/bridge/validators/ExternalIdValidatorTest.java#L36

src/test/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationControllerTest.java
```diff
@@ -0,0 +1,130 @@
...
+        assertGet(HealthDataDocumentationController.class, "getHealthDataDocumentationForId");
+        assertGet(HealthDataDocumentationController.class, "getAllHealthDataDocumentationForParentId");
+        assertDelete(HealthDataDocumentationController.class, "deleteHealthDataDocumentationForIdentifier");
+        assertDelete(HealthDataDocumentationController.class, "deleteAllHealthDataDocumentationForParentId");
+    }
+
+    @Test
+    public void createOrUpdateHealthDataDocumentation() throws Exception {
+        HealthDataDocumentation doc = HealthDataDocumentation.create();
+        doc.setParentId(TEST_APP_ID);
```
DwayneJengSage: Instead of setting the parentId as part of the input, we should leave this blank and verify that we get the correct parentId from the session.

src/main/java/org/sagebionetworks/bridge/validators/HealthDataDocumentationValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.sagebionetworks.bridge.models.HealthDataDocumentation;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+import java.nio.charset.StandardCharsets;
+
+/** Validator for {@link org.sagebionetworks.bridge.models.HealthDataDocumentation}. */
+public class HealthDataDocumentationValidator implements Validator {
```
DwayneJengSage, line 11: Unit tests for the validator?

DwayneJengSage (changes requested):
Also, we should make sure your JavaSDK changes are merged and bump the version numbers in the pom.xml files https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/665

Also, we should add some simple integration tests.

src/test/java/org/sagebionetworks/bridge/spring/controllers/HealthDataDocumentationControllerTest.java
kllysng: @DwayneJengSage I have been stuck on this one - I can't seem to get an argument captor working to see that the healthDataDocumentation was set properly. It seems simple but I am still struggling with Mockito. Could you please help point me in a better direction?

kllysng: Is this one of those situations where I should create a helper function in the controller in order to simplify verifying the arguments, similar to queryHelper in various DAO's?

kllysng: I figured it out with the above method!

src/test/java/org/sagebionetworks/bridge/services/HealthDataDocumentationServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.sagebionetworks.bridge.models.HealthDataDocumentation;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.sagebionetworks.bridge.BridgeConstants.API_DEFAULT_PAGE_SIZE;
+import static org.sagebionetworks.bridge.TestConstants.*;
```
DwayneJengSage: Minor: We don't want to collapse our imports into .* (Same with HealthDataDocumentationControllerTest.)

I think there's an IntelliJ setting that controls this. I can send you my IntelliJ code-style settings if it helps.

DwayneJengSage (approved):
Looks good. One minor note.

Also, do we have integration tests for this?

PR #376, opened 2021-06-02
Fix bug: Can bypass phone region validation if you use international format

DwayneJengSage:
I was scrubbing the logs and found something interesting. Apparently, we were getting SQL exceptions because a couple requests were trying to write phone region codes into the database that were longer than the allowed width of 2 characters.

Digging a little deeper, I discovered that while Google's PhoneNumberUtil only allows 2-character region codes, if you supply an international format phone number to PhoneNumberUtil.parse(), it will ignore the specified region code altogether:
https://www.javadoc.io/doc/com.googlecode.libphonenumber/libphonenumber/8.4.1/com/google/i18n/phonenumbers/PhoneNumberUtil.html#parse-java.lang.String-java.lang.String-

I wrote a few unit tests to verify. PhoneTest.phoneInvalidRegionCodeInternationalFormat() fails without changes to Phone.isValid(), but succeeds after the changes are made.


alxdarksage (approved):
Looks good!

PR #377, opened 2021-06-02
Many changes based on conversations with client developers.

alxdarksage:
BRIDGE-2989, BRIDGE-3001, BRIDGE-3002, BRIDGE-3004

- More robust formatActivityEventId event which will accept session/assessment finished events;
- change PK for adherence records to use event timestamp as part of PK (not createdOn), so that:
- you can change createdOn without creating a new record, and you can delete adherence records, which seems useful for testing;
- updates to assessment adherence records will not start and finish session adherence records, since the former are what the clients are working with and the latter are more useful for adherence & compliance reporting;
- added API so a study coordinator/researcher can edit a user's adherence records (again, useful for testing);
- changes to IRB-related fields in study;
- you can mark an AR as “declined”;
- totalMinutes and totalNotifications added to the Timeline;
- test cleanup, such as moving a test method out of the main utilities class;
- if enrollment is missing in the "recent study events" or "study event history" APIs, add it.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
@@ -136,7 +136,7 @@ public class AdherenceService {
                 .withEventTimestamps(ImmutableMap.of(asmtMeta.getSessionStartEventId(), asmt.getEventTimestamp()))
                 .withInstanceGuids(instanceGuids).build());
         
-        SessionState state = new SessionState();
+        SessionState state = new SessionState(asmtMetas.size());
         
         // The session record may have been submitted, it may be persisted, or
         // it may not yet exist, and we take the records in that order.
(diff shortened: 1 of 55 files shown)


src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -94,35 +98,95 @@ public void updateAdherenceRecords(String appId, AdherenceRecordList recordList)
...
-
         for (AdherenceRecord record : recordList.getRecords()) {
-            if (record.getFinishedOn() == null) {
-                continue;
+            dao.updateAdherenceRecord(record);
+        }
+        for (AdherenceRecord record : recordList.getRecords()) {
+            updateSessionState(appId, record);
+        }
+    }
```
alxdarksage, line 120: Prevent multiple updates of the same session event, which could create N publication updates in the history. [IMPLEMENTED]

alxdarksage: Keep track of and do not duplicate session publication events. [IMPLEMENTED]

alxdarksage: If every assessment has been declined, decline the session. [IMPLEMENTED]

PR #384, opened 2021-06-09
Accept user ID 'token' in many calls and parse it.

alxdarksage:
This lead to a clean up of the getHealthCodeForAccount() method while adding a similar method to get the account ID. No other changes to functionality.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
@@ -25,6 +25,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 
+import org.sagebionetworks.bridge.BridgeUtils;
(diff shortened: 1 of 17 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
```diff
@@ -124,96 +124,113 @@ public ReportIndex getParticipantReportIndex(@PathVariable String identifier) {
...
+    private DateRangeResourceList<? extends ReportData> getParticipantReportInternal(String appId, Account account, String reportId,
             String startDateString, String endDateString) {
         LocalDate startDate = getLocalDateOrDefault(startDateString, null);
         LocalDate endDate = getLocalDateOrDefault(endDateString, null);
 
-        Account account = accountService.getAccount(AccountId.forId(appId, userId));
-        if (account == null) {
-            throw new EntityNotFoundException(Account.class);
-        }
         return reportService.getParticipantReport(appId, reportId, account.getHealthCode(), startDate, endDate);
```
alxdarksage: So, in the end you only need healthCode? Why not pass healthCode?

PR #385, opened 2021-06-16
Permissions

alxdarksage:
About 1,000 changes to the permissions for many calls, to make some of the newer roles more useful in MTB and the BSM.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -36,11 +36,12 @@ public class AuthUtils {
             .hasAnyRole(ADMIN);
     
     /**
-     * Can the caller edit assessments? Must be a member of the organization. Probably should be
-     * some kind of developer role as well!
+     * Can the caller edit assessments? Must be a study designer in the organization that 
+     * owns the assessment, or a developer.
      */
-    public static final AuthEvaluator CAN_EDIT_ASSESSMENTS = new AuthEvaluator().isInOrg().or()
-            .hasAnyRole(ADMIN);
+    public static final AuthEvaluator CAN_EDIT_ASSESSMENTS = new AuthEvaluator()
+            .isInOrg().hasAnyRole(STUDY_DESIGNER).or()
+            .hasAnyRole(DEVELOPER, ADMIN);
     
     /**
      * Can the caller and/remove organization members? Must be the organizations's admin. Note 
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -42,7 +42,14 @@ public class AuthUtils {
     public static final AuthEvaluator CAN_EDIT_ASSESSMENTS = new AuthEvaluator()
             .isInOrg().hasAnyRole(STUDY_DESIGNER).or()
             .hasAnyRole(DEVELOPER, ADMIN);
-    
+
+    /**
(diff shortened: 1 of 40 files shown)


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -85,6 +93,20 @@
...
+    public static final AuthEvaluator CAN_READ_PARTICIPANT_REPORTS = new AuthEvaluator().isSelf().or()
+            .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
+            .hasAnyRole(RESEARCHER, WORKER, ADMIN);
+    
+    /**
+     * Can the caller read study reports?
+     */
+    public static final AuthEvaluator CAN_READ_STUDY_REPORTS = new AuthEvaluator()
+            .canAccessStudy().or()
+            .hasAnyRole();
```
alxdarksage: make this DEVELOPER, RESEARCHER, ADMIN

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
```diff
@@ -697,8 +697,8 @@ private int deleteReportAndUpdateState(App app, String userId) {
             throw new EntityNotFoundException(Account.class);
         }
 
-        reportService.deleteParticipantReportRecord(app.getIdentifier(), APPOINTMENT_REPORT,
-                JAN1.toString(), account.getHealthCode());
+        reportService.deleteParticipantReportRecord(app.getIdentifier(), userId, APPOINTMENT_REPORT, JAN1.toString(),
+                account.getHealthCode());
 
         updateState(account, SELECTED);
         accountService.updateAccount(account);
```
alxdarksage, line 704: Josh should review this to make sure it will work.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
```diff
@@ -244,12 +242,16 @@ public StatusMessage saveParticipantReportForWorker(@PathVariable String identif
             throw new BadRequestException("A health code is required to save report data.");
         }
         String healthCode = node.get("healthCode").asText();
-        
+
+        // NOTE: Use of session.getAppId() here and below doesn't allow the worker to switch accounts.
+        // This is an error in the API.
+        String userId = accountService.getAccountId(session.getAppId(), "healthCode:"+healthCode)
+                .orElseThrow(() -> new EntityNotFoundException(Account.class));
```
alxdarksage, line 247: File a JIRA on this.

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: updated

src/main/java/org/sagebionetworks/bridge/spring/controllers/CRCController.java
alxdarksage: These calls should succeed because we do a study check and Columbia's machine account is in an organization that sponsors the study, BUT it now also needs to be a study coordinator. I made it a researcher which I'm sure will succeed.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantReportController.java
alxdarksage: No, it has to work this way because all the worker had for this call was a health code.

nategbrown9 (approved):
Looks good.

PR #386, opened 2021-06-17
Participant file uploads

jnelsonjava:
- Adding expiration date field for ParticipantFile upload URLs
- Removing protection against overwriting existing files

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
+++ b/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
@@ -22,6 +22,7 @@ public class DynamoParticipantFile implements ParticipantFile {
     private String appId;
     private String uploadUrl;
     private String downloadUrl;
+    private DateTime expires;
 
     public DynamoParticipantFile() {
     }
@@ -120,5 +121,18 @@ public class DynamoParticipantFile implements ParticipantFile {
     @Override
     public void setAppId(String appId) {
         this.appId = appId;
+    }
+
+    @Override
+    @DynamoDBIgnore
+    @JsonSerialize(using = DateTimeSerializer.class)
+    public DateTime getExpires() {
+        return this.expires;
+    }
+
+    @Override
+    @JsonDeserialize(using = DateTimeDeserializer.class)
+    public void setExpires(DateTime expires) {
+        this.expires = expires;
     }
 }
--- a/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
+++ b/src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
@@ -22,7 +22,7 @@ public class DynamoParticipantFile implements ParticipantFile {
     private String appId;
     private String uploadUrl;
     private String downloadUrl;
-    private DateTime expires;
+    private DateTime expiresOn;
 
     public DynamoParticipantFile() {
     }
@@ -126,13 +126,13 @@ public class DynamoParticipantFile implements ParticipantFile {
     @Override
     @DynamoDBIgnore
     @JsonSerialize(using = DateTimeSerializer.class)
-    public DateTime getExpires() {
-        return this.expires;
+    public DateTime getExpiresOn() {
+        return this.expiresOn;
     }
 
     @Override
     @JsonDeserialize(using = DateTimeDeserializer.class)
-    public void setExpires(DateTime expires) {
-        this.expires = expires;
+    public void setExpiresOn(DateTime expiresOn) {
+        this.expiresOn = expiresOn;
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -163,12 +150,14 @@ private String getFilePath(ParticipantFile file) {
     }
 
     private URL generatePresignedRequest(ParticipantFile file, HttpMethod method) {
-        Date expiration = new DateTime().plusMinutes(EXPIRATION_IN_MINUTES).toDate();
+        DateTime expiration = file.getCreatedOn().plusMinutes(EXPIRATION_IN_MINUTES);
```
DwayneJengSage: This will break getParticipantFile().

This method is used for both get and create. The intent is that the pre-signed URL should expire 24 hours after the request is generated. However, since get requests can come long after the file is first created, this will generate expiration times in the past.

You want to keep the `new DateTime()`, or better yet, use `DateTime.now()`.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -163,12 +150,14 @@ private String getFilePath(ParticipantFile file) {
...
+
         GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(bucketName, getFilePath(file), method);
-        request.setExpiration(expiration);
+        request.setExpiration(expiration.toDate());
         if (PUT.equals(method)) {
             request.setContentType(file.getMimeType());
             request.addRequestParameter(Headers.SERVER_SIDE_ENCRYPTION, ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
+            file.setExpires(expiration);
```
DwayneJengSage: Pre-signed URLs expire for both get and put. We should put in the expiration date regardless.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -153,6 +152,8 @@ public void createParticipantFile() {
         assertEquals(result.getUploadUrl(), upload);
         assertEquals(result.getAppId(), "api");
         assertEquals(TestConstants.TIMESTAMP.compareTo(result.getCreatedOn()), 0);
+        assertEquals(TestConstants.TIMESTAMP.plusDays(1).toDate(), result.getExpires().toDate());
```
DwayneJengSage: I'm not sure what the old code was doing, but you can just say

```
assertEquals(TestConstants.TIMESTAMP, result.getCreatedOn())
```

And similarly for expires.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantFile.java
```diff
@@ -22,6 +22,7 @@
     private String appId;
     private String uploadUrl;
     private String downloadUrl;
+    private DateTime expires;
```
DwayneJengSage: To be consistent with our other timestamps, let's rename this to `expiresOn`.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -175,7 +174,7 @@ public void createParticipantFileInvalidFile() {
         service.createParticipantFile("api", "test_user", file);
     }
 
-    @Test(expectedExceptions = EntityAlreadyExistsException.class)
+    @Test
     public void createParticipantFileAlreadyExists() {
```
DwayneJengSage: I think this test can be just deleted. Since mocks have no state, the mockDao won't "remember" that a previous file of the same ID has already been created. Since we no longer query the mockDao for the existing file, this test functionally does the same thing as createParticipantFile().

DwayneJengSage (changes requested):
Also, should we assign https://sagebionetworks.jira.com/browse/BRIDGE-2926 to Jack?

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
```diff
@@ -152,7 +152,9 @@ public void createParticipantFile() {
         assertNotNull(result.getCreatedOn());
         assertEquals(result.getUploadUrl(), upload);
         assertEquals(result.getAppId(), "api");
-        assertEquals(TestConstants.TIMESTAMP.compareTo(result.getCreatedOn()), 0);
+        assertEquals(result.getCreatedOn().toDateTime(DateTimeZone.UTC), TestConstants.TIMESTAMP);
+        assertEquals(result.getExpiresOn().toDateTime(DateTimeZone.UTC), TestConstants.TIMESTAMP.plusDays(1));
```
DwayneJengSage: Optional: Use .getMillis() instead of .toDateTime(timezone).

DwayneJengSage (approved):
Looks good to me, other than the one optional change I commented on.

src/test/java/org/sagebionetworks/bridge/services/ParticipantFileServiceTest.java
jnelsonjava: I appreciate the reminder, just made the switch.

PR #387, opened 2021-06-17
Proposal to calculate daysInStudy for each study based on event defined for that study.

alxdarksage:
DON'T MERGE. I'm looking at other options that won't hit the table size limits

Before working further on this I wanted to review the design. I believe we need to retrofit this into the v2 exporter because scientists are already looking for it on studies like MindKind, but there may be additional work to include it in v3 as well (if we're including this value in v3). 

JIRA: https://sagebionetworks.jira.com/browse/BRIDGE-2996

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
+++ b/src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
@@ -3,6 +3,8 @@ package org.sagebionetworks.bridge.upload;
 import static org.sagebionetworks.bridge.BridgeUtils.collectExternalIds;
 import static org.sagebionetworks.bridge.BridgeUtils.mapStudyMemberships;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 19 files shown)


src/test/java/org/sagebionetworks/bridge/services/StudyActivityEventServiceTest.java
```diff
@@ -476,4 +479,39 @@ public void doNotAddEnrollmentToHistoryIfNotMissing() {
         assertEquals(retValue.getTotal(), Integer.valueOf(1));
         assertEquals(retValue.getItems().get(0).getTimestamp(), CREATED_ON); // not modifiedOn
     }
+
+    @Test
+    @Ignore
```
DwayneJengSage, line 484: This test is ignored? Is this test currently disabled because it is not yet ready for review? Or is it disabled because we forget to un-ignore it?

src/test/java/org/sagebionetworks/bridge/services/StudyActivityEventServiceTest.java
```diff
@@ -54,6 +55,8 @@
 import org.sagebionetworks.bridge.models.apps.App;
 import org.sagebionetworks.bridge.models.studies.Enrollment;
 
+import jdk.nashorn.internal.ir.annotations.Ignore;
```
DwayneJengSage, line 58: This appears to be the wrong Ignore annotation.

src/test/java/org/sagebionetworks/bridge/services/StudyServiceTest.java
```diff
@@ -228,6 +235,7 @@ public void createStudyWithSponsorship() {
         DateTime timestamp = DateTime.now().minusHours(2);
         study.setCreatedOn(timestamp);
         study.setModifiedOn(timestamp);
+        study.setStudyStartEventId("enrollment");
```
DwayneJengSage, line 238: Should we also verify that the studyStartEventId was parsed and set properly? (Here and in the update() test as well.)

src/main/java/org/sagebionetworks/bridge/upload/TranscribeConsentHandler.java
```diff
@@ -66,15 +68,28 @@ public void handle(@Nonnull UploadValidationContext context) {
...
+
         } else {
             // default sharing to NO_SHARING
             record.setUserSharingScope(SharingScope.NO_SHARING);
         }
     }
+
+    protected int calculateDaysSince(DateTime studyStartTime) {
+        LocalDate studyStartDate = studyStartTime.withZone(BridgeConstants.LOCAL_TIME_ZONE).toLocalDate();
+        LocalDate todayLocalDate = DateUtils.getCurrentCalendarDateInLocalTime();
```
DwayneJengSage, line 90: So originally, we made the design decision to use the current time to calculate dayInStudy. However, this has the side effect that if we redrive the upload on a later date, the dayInStudy may change. This is probably not desirable behavior. Consider changing this to use record.getCreatedOn() instead. (NOTE: This is guaranteed to exist by UploadFormatHandler, and can be overwritten by IosSchemaValidationHandler2 and GenericUploadFormatHandler as they parse the upload.)

Also, while we're at it, instead of using BridgeConstants.LOCAL_TIME_ZONE, we can use record.getCreatedOnTimeZone() to get the participant's local time zone. (And fall back to BridgeConstants.LOCAL_TIME_ZONE if it doesn't exist.)

DwayneJengSage (changes requested):
A few comments about the implementation. Comments about the design are left in Jira and will be discussed separately.

PR #388, opened 2021-06-21
Filter assessments by owner

alxdarksage:
In addition to applying ownership rules to the editing of assessments, these changes limit visibility by ownership. Ownership cannot be checked from the request context, because shared assessments do not have this limitation, and both kinds of assessments converge to the same code path. So organization ID is passed through many methods as an argument to check when querying the database.

JIRA: BRIDGE-2948

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -151,7 +151,7 @@ public class AuthUtils {
      * Can the caller import a shared assessment under the supplied organization? Yes if
      * the caller is a developer or admin, or if the caller is a study designer assigning
      * the assessment to their own organization. Note that this expects an orgId to be 
-     * supplied, now a compound ownerId (appId + orgId), because it is verifying the final
+     * supplied, not a compound ownerId (appId + orgId), because it is verifying the final
      * assignment of the supplied orgId.
      */
     public static final AuthEvaluator CAN_IMPORT_SHARED_ASSESSMENTS = new AuthEvaluator()
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -36,12 +36,20 @@ public class AuthUtils {
             .hasAnyRole(ADMIN);
     
     /**
-     * Can the caller edit assessments? Must be a member of the organization. Probably should be
-     * some kind of developer role as well!
+     * Can the caller edit assessments? Must be a study designer in the organization that 
+     * owns the assessment, or a developer.
      */
-    public static final AuthEvaluator CAN_EDIT_ASSESSMENTS = new AuthEvaluator().isInOrg().or()
+    public static final AuthEvaluator CAN_EDIT_ASSESSMENTS = new AuthEvaluator()
+            .isInOrg().hasAnyRole(STUDY_DESIGNER).or()
             .hasAnyRole(DEVELOPER, ADMIN);
-    
+
+    /**
+     * Can the caller and/remove organization members? Must be the organizations's admin. Note 
+     * that this check is currently also used for sponsors...which are not members.
+     */
+    public static final AuthEvaluator CAN_READ_MEMBERS = new AuthEvaluator()
+            .isInOrg().or().hasAnyRole(ADMIN);
+
     /**
      * Can the caller and/remove organization members? Must be the organizations's admin. Note 
      * that this check is currently also used for sponsors...which are not members.
@@ -86,6 +94,20 @@ public class AuthUtils {
             .hasAnyRole(RESEARCHER, WORKER, ADMIN);
     
     /**
+     * Can the caller read participant reports? 
+     */
+    public static final AuthEvaluator CAN_READ_PARTICIPANT_REPORTS = new AuthEvaluator().isSelf().or()
+            .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
+            .hasAnyRole(RESEARCHER, WORKER, ADMIN);
+    
+    /**
+     * Can the caller read study reports?
+     */
+    public static final AuthEvaluator CAN_READ_STUDY_REPORTS = new AuthEvaluator()
+            .canAccessStudy().or()
+            .hasAnyRole(DEVELOPER, RESEARCHER, WORKER, ADMIN);
+    
+    /**
      * Can the caller enroll or withdraw participants from a study? Must be enrolling self, or 
      * be a study coordinator with access to the study involved, or be a researcher. 
      */
@@ -113,7 +135,7 @@ public class AuthUtils {
      * Can the caller edit studies? Caller must be a study coordinator, or a developer.
      */
     public static final AuthEvaluator CAN_UPDATE_STUDIES = new AuthEvaluator()
-            .canAccessStudy().hasAnyRole(STUDY_COORDINATOR, STUDY_DESIGNER).or()
+            .canAccessStudy().hasAnyRole(STUDY_DESIGNER).or()
             .hasAnyRole(DEVELOPER, ADMIN);
     
     /**
@@ -126,13 +148,24 @@ public class AuthUtils {
             .hasAnyRole(RESEARCHER, ADMIN);
     
     /**
+     * Can the caller import a shared assessment under the supplied organization? Yes if
+     * the caller is a developer or admin, or if the caller is a study designer assigning
+     * the assessment to their own organization. Note that this expects an orgId to be 
+     * supplied, now a compound ownerId (appId + orgId), because it is verifying the final
+     * assignment of the supplied orgId.
+     */
+    public static final AuthEvaluator CAN_IMPORT_SHARED_ASSESSMENTS = new AuthEvaluator()
+            .isInOrg().hasAnyRole(STUDY_DESIGNER).or()
+            .hasAnyRole(DEVELOPER, ADMIN);
+
+    /**
      * Can the caller edit shared assessments? The caller must be a member of an organization 
      * expressed in the shared organization ID format, or "appId:orgId" (which is used in 
      * shared assessments so that organization IDs do not collide between applications). 
      */
     public static final AuthEvaluator CAN_EDIT_SHARED_ASSESSMENTS = new AuthEvaluator()
-            .isSharedOwner().or()
-            .hasAnyRole(ADMIN);
+            .isSharedOwner().hasAnyRole(STUDY_DESIGNER).or()
+            .hasAnyRole(DEVELOPER, ADMIN);
 
     /**
      * Can the caller read the schedules? They must be enrolled in the study, a study-scoped
@@ -141,7 +174,7 @@ public class AuthUtils {
      * yet. 
      */
     public static final AuthEvaluator CAN_READ_SCHEDULES = new AuthEvaluator()
-            .isInOrg().hasAnyRole(STUDY_DESIGNER).or()
+            .isInOrg().or()
             .isEnrolledInStudy().or()
             .hasAnyRole(DEVELOPER, ADMIN);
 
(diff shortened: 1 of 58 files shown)


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
```diff
@@ -109,20 +121,30 @@ final void setHibernateHelper(HibernateHelper hibernateHelper) {
     }
     
     @Override
-    public Optional<Assessment> getAssessment(String appId, String guid) {
+    public Optional<Assessment> getAssessment(String appId, String ownerId, String guid) {
+        QueryBuilder builder = new QueryBuilder();
+        builder.append(GET_BY_GUID, APP_ID, appId, GUID, guid);
+        if (ownerId != null) {
+            builder.append(LIMIT_TO_OWNER, OWNER_ID, ownerId); 
+        }
```
alxdarksage, line 130: But don't we check this after we load it using AuthUtils?

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -147,6 +147,17 @@
             .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
             .hasAnyRole(RESEARCHER, ADMIN);
     
+    /**
+     * Can the caller import a shared assessment under the supplied organization? Yes if
+     * the caller is a developer or admin, or if the caller is a study designer assigning
+     * the assessment to their own organization. Note that this expects an orgId to be 
+     * supplied, now a compound ownerId (appId + orgId), because it is verifying the final
```
alxdarksage: now -> "not"

src/main/java/org/sagebionetworks/bridge/services/AssessmentConfigService.java
```diff
@@ -53,13 +53,10 @@ AssessmentConfigValidator getValidator() {
         return AssessmentConfigValidator.INSTANCE;
     }
     
-    public AssessmentConfig getAssessmentConfig(String appId, String guid) {
+    public AssessmentConfig getAssessmentConfig(String appId, String ownerId, String guid) {
```
alxdarksage: Is this being called by the client for a participant, and if so, would this block it?

src/main/java/org/sagebionetworks/bridge/services/AssessmentResourceService.java
```diff
@@ -145,7 +145,7 @@ public AssessmentResource updateSharedResource(String callerAppId, String assess
         checkArgument(isNotBlank(assessmentId));
         checkNotNull(resource);
         
-        Assessment assessment = assessmentService.getLatestAssessment(SHARED_APP_ID, assessmentId);
+        Assessment assessment = assessmentService.getLatestAssessment(SHARED_APP_ID, null, assessmentId);
```
alxdarksage, line 148: So why is ownerId not supplied here? Isn't it required for an update?

src/main/java/org/sagebionetworks/bridge/services/AssessmentResourceService.java
```diff
@@ -218,7 +218,7 @@ public void deleteResourcePermanently(String appId, String assessmentId, String
         checkArgument(isNotBlank(assessmentId));
         
         // Must have published the assessment already before you move resources
-        Assessment assessment = assessmentService.getLatestAssessment(SHARED_APP_ID, assessmentId);
+        Assessment assessment = assessmentService.getLatestAssessment(SHARED_APP_ID, null, assessmentId);
```
alxdarksage, line 221: So there are two ways here to verify the caller owns the assessment, one using the query, one checking the object, do we need both?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentConfigController.java
```diff
@@ -37,33 +37,36 @@ final void setAssessmentConfigService(AssessmentConfigService service) {
     public AssessmentConfig getAssessmentConfig(@PathVariable String guid) {
         UserSession session = getAuthenticatedSession();
         String appId = session.getAppId();
+        String ownerId = session.getParticipant().getOrgMembership();
         
-        return service.getAssessmentConfig(appId, guid);
+        return service.getAssessmentConfig(appId, ownerId, guid);
```
alxdarksage: This is wrong. Anyone can read the config, most particularly, participants!

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentConfigController.java
```diff
@@ -37,33 +37,36 @@ final void setAssessmentConfigService(AssessmentConfigService service) {
...
     }
     
     @PostMapping("/v1/assessments/{guid}/config")
     public AssessmentConfig updateAssessmentConfig(@PathVariable String guid) {
         UserSession session = getAuthenticatedSession(DEVELOPER, STUDY_DESIGNER);
         String appId = session.getAppId();
+        String ownerId = session.getParticipant().getOrgMembership();
```
alxdarksage: For assessment config, also allow developers and admins to manipulate any config (as assessment controller).

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
```diff
@@ -145,26 +159,30 @@ public Assessment publishAssessment(@PathVariable String guid,
...
     }
         
     @DeleteMapping("/v1/assessments/{guid}")
     public StatusMessage deleteAssessment(@PathVariable String guid, @RequestParam(required = false) String physical) {
         UserSession session = getAuthenticatedSession(DEVELOPER, STUDY_DESIGNER, ADMIN);
 
         String appId = session.getAppId();
+        // An admin can logically delete and in that case, we do want to pass the correct
+        // ownerId, so we don't nullify it here.
+        String ownerId = session.getParticipant().getOrgMembership();
```
alxdarksage: Really needed? Who cares if null is passed in below?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentResourceController.java
```diff
@@ -65,77 +73,81 @@ final void setAssessmentResourceService(AssessmentResourceService service) {
...
+
+        return service.updateResource(appId, ownerId, assessmentId, resource);
     }
 
     @DeleteMapping("/v1/assessments/identifier:{assessmentId}/resources/{guid}")
     public StatusMessage deleteAssessmentResource(@PathVariable String assessmentId, @PathVariable String guid,
             @RequestParam(required = false) String physical) {
         UserSession session = getAuthenticatedSession(DEVELOPER, STUDY_DESIGNER, ADMIN);
         String appId = session.getAppId();
+        String ownerId = session.getParticipant().getOrgMembership();
```
alxdarksage, line 135: This can use the getOwnerId() method, and then you don't have to do the check below

src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
```diff
@@ -47,6 +47,7 @@ void setLogger(Logger log) {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity<String> handleException(HttpServletRequest request, Exception ex) throws JsonProcessingException {
+        ex.printStackTrace();
```
alxdarksage: REMOVEME

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDaoTest.java
```diff
@@ -135,13 +150,31 @@ public void getAssessmentsExcludeDeleted() {
...
+        when(mockHelper.nativeQueryCount(queryCaptor.capture(), any())).thenReturn(5);
+        when(mockHelper.nativeQueryGet(queryCaptor.capture(), paramsCaptor.capture(), eq(0), eq(20), eq(HibernateAssessment.class)))
+                .thenReturn(list);
+        
+        PagedResourceList<Assessment> page = dao.getAssessments(APP_ID_VALUE, TEST_ORG_ID, 0, 20, null, false);
+        assertEquals(queryCaptor.getAllValues().get(0), "SELECT COUNT(*) " + QUERY_SQL_WITH_OWNERID_EXC_DELETED);
+        assertEquals(queryCaptor.getAllValues().get(1), "SELECT * " + QUERY_SQL_WITH_OWNERID_EXC_DELETED);
+        
+        Map<String,Object> params = paramsCaptor.getValue();
+        assertEquals(params.get("appId"), APP_ID_VALUE);
```
alxdarksage, line 165: verify ownerId set

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: fixed

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentConfigController.java
alxdarksage: This has been changed for this controller.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
alxdarksage: This can use getOwnerId() like everywhere else, changed.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentResourceController.java
alxdarksage: Changed

src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
alxdarksage: removed

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDaoTest.java
alxdarksage: added

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentConfigController.java
alxdarksage: Changed so the configs can be read by everyone.

src/main/java/org/sagebionetworks/bridge/services/AssessmentConfigService.java
alxdarksage: Configs now do not check ownership.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
alxdarksage: We do, but they are not redundant. For example, some calls set the ownerId to null...so they must then check if the caller is a study designer, that they have access to the assessment. They complement each other and eliminate edge cases.

src/main/java/org/sagebionetworks/bridge/services/AssessmentResourceService.java
alxdarksage: Because it's a shared assessment with a compound ID, so we can't check it by ownerId. There's an AuthUtils check after this that does the correct permissions check.

src/main/java/org/sagebionetworks/bridge/services/AssessmentResourceService.java
alxdarksage: We do. They complement each other

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAssessmentDao.java
alxdarksage: We do use both, but they complement each other. In cases where this value is null, we then want to verify with the AuthUtils check.

nategbrown9 (approved):
Looks good.

PR #389, opened 2021-06-22
Study additions

alxdarksage:
- Adding list of sign in types to indicate to participant clients (if they have a study ID) what kind of authentication should be presented to the user;
- Converted study design types and diseases to lists;
- Annotations to reduce the number of queries to retrieve studies (to one, it appears);
- If the IRB decision type is "exempt," then an expiresOn value for the decision is no longer required.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
@@ -115,9 +115,8 @@ public class HibernateStudy implements Study {
     /**
      * For partial construction of object by Hibernate, excluding expensive fields like clientData.
      */
-    public HibernateStudy(String name, String identifier, String appId, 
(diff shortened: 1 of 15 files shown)


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
```diff
@@ -90,14 +117,15 @@ public HibernateStudy() {}
      */
     public HibernateStudy(String name, String identifier, String appId, 
             DateTime createdOn, DateTime modifiedOn, boolean deleted, 
-            StudyPhase phase, Long version) {
+            StudyPhase phase, List<SignInType> signInTypes, Long version) {
```
alxdarksage: Maybe this can come out.

syoung-smallwisdom (approved):
looks good

nategbrown9 (approved):
Looks good.

PR #391, opened 2021-06-24
Deleting previous S3 file during create request

jnelsonjava:
This is a minor fix to make sure a participant can not overwrite a ParticipantFile record while leaving the original file in S3.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
@@ -118,6 +118,10 @@ public class ParticipantFileService {
 
         participantFileDao.uploadParticipantFile(file);
 
+        // Deleting any previous object prevents a user from updating the ParticipantFile
+        // but leaving the previous object on S3.
+        s3Client.deleteObject(bucketName, getFilePath(file));
+
         file.setUploadUrl(generatePresignedRequest(file, PUT).toExternalForm());
         return file;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
```diff
@@ -118,6 +118,10 @@ public ParticipantFile createParticipantFile(String appId, String userId, Partic
 
         participantFileDao.uploadParticipantFile(file);
 
+        // Deleting any previous object prevents a user from updating the ParticipantFile
+        // but leaving the previous object on S3.
+        s3Client.deleteObject(bucketName, getFilePath(file));
```
DwayneJengSage, line 123: It's probably worth adding a `verify(mockS3Client).deleteObject()` in the unit tests.

Also, what happens if the object doesn't exist in S3? Does it do nothing or does it throw?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
jnelsonjava: That's good call, I'll update the unit tests.

If the object doesn't exist in S3, it just does nothing. The delete route uses that method as well and has a comment saying it returns as successful even if it didn't find anything to delete.

alxdarksage (changes requested):
As Dwayne mentions.

src/main/java/org/sagebionetworks/bridge/services/ParticipantFileService.java
jnelsonjava: Updated to verify that deleteObject was run through the mockS3Client.

alxdarksage (approved):

DwayneJengSage (approved):

PR #392, opened 2021-06-24
Upload study logo

alxdarksage:
BRIDGE-3012: add APIs to upload a study logo, which is then accessible via the study object. Uses the existing hosted files service to manage the file as a hosted file.


nategbrown9 (approved):
Looks good.

PR #393, opened 2021-06-25
New system events and endpoints to trigger the install link message

alxdarksage:
Implements BRIDGE-3023: there are two new system events (first_sign_in and sent_install_link) that can be used by MTB for certain administrative tasks (scheduling against sign in and following-up with participants if they don't sign in after the install link is sent).

Implements BRIDGE-3024: Adds two new endpoints to trigger the install link template via email or phone (depending on the user's account configuration).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventObjectType.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventObjectType.java
@@ -165,7 +165,7 @@ public enum ActivityEventObjectType {
      * An event to record that the install link message has been sent to a participant. Note
      * that this records the message was sent, and does not guarantee that it was received.
      */
-    SENT_INSTALL_LINK(FUTURE_ONLY) {
+    INSTALL_LINK_SENT(FUTURE_ONLY) {
         public String getEventId(String objectId, ActivityEventType eventType, String answerValue) {
             return this.name().toLowerCase();
         }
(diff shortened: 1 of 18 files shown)


src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventObjectType.java
```diff
@@ -149,6 +149,27 @@ public String getEventId(String objectId, ActivityEventType eventType, String an
...
+    
+    /**
+     * An event created the first time the participant signs in to the Bridge server via 
+     * any authentication method.
+     */
+    FIRST_SIGN_IN(IMMUTABLE) {
+        public String getEventId(String objectId, ActivityEventType eventType, String answerValue) {
+            return this.name().toLowerCase();
+        }
+    },
```
alxdarksage: Rename first_app_sign_in, OR email everyone and verify that someone wants this. Because as Nate notes, it could be confusing for schedulers in a multi-study scenario.

src/main/java/org/sagebionetworks/bridge/services/ActivityEventService.java
```diff
@@ -161,6 +163,42 @@ public void publishActivitiesRetrieved(App app, String healthCode, DateTime time
         }
     }
     
+    public void publishFirstSignIn(App app, String healthCode, DateTime timestamp) {
+        checkNotNull(app);
+        checkNotNull(healthCode);
+
```
alxdarksage: checkNotNull(timestamp) (here and below)

nategbrown9 (approved):
Looks good

src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventObjectType.java
alxdarksage: removed

PR #394, opened 2021-06-28
Fix creation of public buckets.

alxdarksage:
BRIDGE-3032

There were multiple issues here discovered during manual testing of a study logo upload feature:

- To download these files as public files, you need to enable S3 static website hosting;
- the buckets are named after the domain names in Route 53 and Cloud Front, however, I do not see any straightforward way to configure a domain name in the production AWS account for an S3 bucket in the development AWS account. So for local, development, and staging buckets, it is easier to use the hostname provided by S3 bucket static website hosting. Adjusted the configuration file so it is possible to override this for local/development
- To upload files to the public bucket from a web browser, you need to enable PUT CORS permissions.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
@@ -106,12 +106,19 @@ public class S3InitializerTest extends Mockito {
         verify(mockS3Client).createBucket(requestCaptor.capture());
         verify(mockS3Client).setBucketPolicy(BUCKET_NAME, resolvedPolicy);
         verify(mockS3Client).setBucketWebsiteConfiguration(eq(BUCKET_NAME), websiteConfigCaptor.capture());
+        verify(mockS3Client).setBucketCrossOriginConfiguration(eq(BUCKET_NAME), corsConfigCaptor.capture());
         
         assertEquals(requestCaptor.getValue().getBucketName(), BUCKET_NAME);
         
         // Not a lot to check here, but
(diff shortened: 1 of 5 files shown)


src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
```diff
@@ -101,8 +105,13 @@ public void bucketCreatedForPublicBucket() {
         
         verify(mockS3Client).createBucket(requestCaptor.capture());
         verify(mockS3Client).setBucketPolicy(BUCKET_NAME, resolvedPolicy);
+        verify(mockS3Client).setBucketWebsiteConfiguration(eq(BUCKET_NAME), websiteConfigCaptor.capture());
         
         assertEquals(requestCaptor.getValue().getBucketName(), BUCKET_NAME);
+        
+        // Not a lot to check here, but
+        BucketWebsiteConfiguration config = websiteConfigCaptor.getValue();
+        assertEquals(config.getIndexDocumentSuffix(), "index.html");
```
DwayneJengSage: Should we also check Cors configuration here?

DwayneJengSage (approved):
One minor comment about unit tests. Otherwise, looks good.

Also, is there a corresponding BridgeIntegTests change to verify the bucket config?

src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
alxdarksage: Added

alxdarksage:
Added integration test that downloads the file from the docs folder and verifies it (separate PR).

PR #396, opened 2021-06-30
New participant note

jnelsonjava:
For [Bridge-3034](https://sagebionetworks.jira.com/browse/BRIDGE-3034).

Allows notes to be added to Participants on account creation.


alxdarksage (approved):

alxdarksage (changes requested):
The same security check we apply on update, we should apply on create (participants signing themselves up should not be able to add a note, and it shouldn't be part of the SignUp API).

alxdarksage (approved):

PR #402, opened 2021-07-09
Delete adherence record

jnelsonjava:
For [https://sagebionetworks.jira.com/browse/BRIDGE-3009](https://sagebionetworks.jira.com/browse/BRIDGE-3009).

Creating API access to delete Adherence Records.

This is currently untested. Just checking if the design is on the right track.


src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -276,4 +276,28 @@ protected void addToMap(Map<String, DateTime> events,
             }
         }
     }
+
+    /* TODO: Add deleteAR method. Should delete AdherenceRecord but might also need
+    *   to update Assessments or Sessions. Also might need to validate input. */
+    public void deleteAdherenceRecords(String appId, AdherenceRecordList recordList) {
+        checkNotNull(recordList);
```
alxdarksage: This can be for a single record and it's a lot easier if it is. Since it's for testing, they will be manually deleting single records.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
```diff
@@ -91,4 +93,38 @@ public StatusMessage updateAdherenceRecords(@PathVariable String studyId, @PathV
         
         return service.getAdherenceRecords(session.getAppId(), search);
     }
+
+    @DeleteMapping("/v5/studies/{studyId}/participants/self/adherence")
```
alxdarksage: We don't need a "self" endpoint... this will only be done by people testing the system.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
```diff
@@ -91,4 +93,38 @@ public StatusMessage updateAdherenceRecords(@PathVariable String studyId, @PathV
...
+        }
+//        TODO: Build deletion method in service
+//        service.deleteAdherenceRecords(session.getAppId(), recordsList);
+        return DELETED_MSG;
+    }
+
+    @DeleteMapping("/v5/studies/{studyId}/participants/{userIdToken}/adherence")
+    public StatusMessage deleteAdherenceRecords(@PathVariable String studyId, @PathVariable String userIdToken) {
+//        TODO: Verify roles with access to Delete. Currently same as Update.
+        UserSession session = getAuthenticatedSession(RESEARCHER, STUDY_COORDINATOR);
```
alxdarksage, line 105: Yep, seems good. Some day we're going to have a way for testers to do this kind of thing too.

src/main/java/org/sagebionetworks/bridge/dao/AdherenceRecordDao.java
```diff
@@ -10,4 +10,6 @@
     
     PagedResourceList<AdherenceRecord> getAdherenceRecords(AdherenceRecordsSearch search);
 
+    void deleteAdherenceRecord(AdherenceRecord record);
```
alxdarksage: Maaaaybe name this "deleteAdherenceRecordPermanently"

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
```diff
@@ -91,4 +93,38 @@ public StatusMessage updateAdherenceRecords(@PathVariable String studyId, @PathV
...
+
+        String userId = accountService.getAccountId(session.getAppId(), userIdToken)
+                .orElseThrow(() -> new EntityNotFoundException(Account.class));
+
+        AdherenceRecordList recordsList = parseJson(AdherenceRecordList.class);
+        for (AdherenceRecord oneRecord : recordsList.getRecords()) {
+            oneRecord.setAppId(session.getAppId());
+            oneRecord.setUserId(userId);
+            oneRecord.setStudyId(studyId);
+        }
```
alxdarksage: In a delete method, there is no body, so the URL above is going to have to take the instance GUID and pass that along to find and delete the record: /v5/studies/{studyId}/participants/{userIdToken}/adherence/{instanceGuid}

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
```diff
@@ -91,4 +94,28 @@ public StatusMessage updateAdherenceRecords(@PathVariable String studyId, @PathV
...
+        String userId = accountService.getAccountId(session.getAppId(), userIdToken)
+                .orElseThrow(() -> new EntityNotFoundException(Account.class));
+
+        AdherenceRecord record = new AdherenceRecord();
+        record.setAppId(session.getAppId());
+        record.setUserId(userId);
+        record.setStudyId(studyId);
+        record.setInstanceGuid(instanceGuid);
+        record.setInstanceTimestamp(DateTime.parse(startedOn));
+        record.setEventTimestamp(DateTime.parse(eventTimestamp));
```
alxdarksage: Consider using BridgeUtils.getDateTimeOrDefault(eventTimestamp, null) for the timestamps. This will set these values to null, and then you should validate they are set in the service. Otherwise, the user with get a cryptic error related to DateTime parsing that doesn't explicitly tell them what is wrong.

jnelsonjava: I switched to BridgeUtils.getDateTimeOrDefault() but I just realized I haven't validated those fields in the service. I'll make a quick update and add unit tests for it.

alxdarksage (approved):

PR #403, opened 2021-07-15
Remove ExternalIdDao.

alxdarksage:
The functionality is moved to AccountDao, because external IDs are now part of enrollment, and thus manipulated through the account model, and we no longer use DynamoDB to record these.


jnelsonjava (approved):
Looks good!

PR #404, opened 2021-07-16
Simplify "isSelf()" check to remove the need to acquire identity

alxdarksage:
Request context gains a user ID on any call that requires a session, so by definition any call that does not have a userId in the context is a public or unauthenticated call. By adjusting the check in isSelf and allowing these calls to pass, we don't have to use acquireAccountIdentity to get it to pass, and can remove all that code. These are cases where we just need to ensure the alternative to our declarative security is present and working (e.g. password verification, or sending an email or SMS message).

Eventually the only two ways to retrieve an account will get AccountService.getAccount() (all permissions checked) and AccountService.editAccount() (no permissions checked). The latter is for internal calls that are updating a bit of information on the account, while the former is for any larger edit of the account where permissions matter.

Many tests had to be adjusted so they looked like authenticated rather than unauthenticated calls (we had to add caller user ID to the test setup). But otherwise everything behaves the same.

Changes were verified with integration tests and manual testing. No changes were required in any tests or client code.


src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
```diff
@@ -96,21 +96,23 @@ public AuthEvaluator isInOrg() {
...
         predicates.add((factMap) -> {
             String userId = factMap.get(USER_ID);
             String callerUserId = RequestContext.get().getCallerUserId();
-            // Calls like signUp happen without a session so there is no caller user ID in the 
-            // request context. In this case, if we’re also not comparing the user ID to a 
-            // known ID, allow this test to pass. This removes some special case code elsewhere
-            // in the system.
-            return (userId == null && callerUserId == null) ||
-                   (userId != null && userId.equals(callerUserId));
+            // return callerUserId == null || (userId != null && userId.equals(callerUserId));
```
alxdarksage: remove commented out code

src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
```diff
@@ -96,21 +96,23 @@ public AuthEvaluator isInOrg() {
...
     /**
-     * The caller is operating on their own account (the target user ID).
+     * The caller is operating on their own account (the target user ID), OR the call 
+     * is anonymous (and is thus also about operating on one’s own account). In the 
+     * latter case, we don’t verify access to the account through security rules, 
+     * because the caller is not directly selecting the account. Instead, we’ll use 
+     * something like password verification or sending email/SMS to verify the 
+     * operation is allowed. 
      */
     public AuthEvaluator isSelf() {
```
alxdarksage, line 106: isSelfOrPublic? or separate?

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -300,11 +300,11 @@ public void editAccount(String appId, String healthCode, Consumer<Account> accou
         checkNotNull(healthCode);
         
         AccountId accountId = AccountId.forHealthCode(appId, healthCode);
-        Account account = getAccount(accountId);
+        Account account = accountDao.getAccount(accountId).orElse(null);
         if (account != null) {
             accountEdits.accept(account);
             accountDao.updateAccount(account);
-        }        
+        }
```
alxdarksage, line 307: Can we roll this back and introduce in a later commit?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AuthenticationController.java
```diff
@@ -353,6 +355,8 @@ public JsonNode changeApp() {
             .withAppId(targetApp.getIdentifier())
             .build();
         
+        RequestContext.set(RequestContext.get().toBuilder().withCallerUserId(account.getId()).build());
```
alxdarksage, line 358: Would reset this to null be clearer?

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -1072,20 +1072,6 @@ public void authenticateAccountUnverifiedPhoneSucceedsForLegacy() throws Excepti
...
-
-        Account persistedAccount = mockGetAccountById(ACCOUNT_ID, false);
-        persistedAccount.setEnrollments(Sets.newHashSet(ACCOUNT_ENROLLMENTS));
-        when(mockAccountDao.getAccount(any())).thenReturn(Optional.of(persistedAccount));
-
-        service.editAccount(TEST_APP_ID, HEALTH_CODE, (account) -> fail("Should have thrown exception"));
-
-        verify(mockAccountDao, never()).updateAccount(any());
-        RequestContext.set(null);
-    }
```
alxdarksage: Could we restore this to a later check in?

src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
alxdarksage: We'll separate these in a later stage

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: Done

jnelsonjava (approved):

PR #405, opened 2021-07-19
Refactor editAccount() so it's more generally useful

alxdarksage:
Basically instead of healthCode, take AccountId, which allows us to use ID, healthCode, or whatever else is available.

Remove accesses of AccountDao that circumvent AccountService. Unfortunately due to the way we organized our code in the early days, we can't use Java packages to enforce this, but we want all access of accounts to go through the AccountService.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
@@ -28,15 +28,12 @@ import org.springframework.stereotype.Component;
 import org.sagebionetworks.bridge.RequestContext;
 import org.sagebionetworks.bridge.cache.CacheKey;
 import org.sagebionetworks.bridge.cache.CacheProvider;
-import org.sagebionetworks.bridge.dao.AccountDao;
 import org.sagebionetworks.bridge.dao.OrganizationDao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
 import org.sagebionetworks.bridge.exceptions.EntityAlreadyExistsException;
 import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.models.AccountSummarySearch;
 import org.sagebionetworks.bridge.models.PagedResourceList;
-import org.sagebionetworks.bridge.models.accounts.Account;
-import org.sagebionetworks.bridge.models.accounts.AccountId;
 import org.sagebionetworks.bridge.models.accounts.AccountSummary;
 import org.sagebionetworks.bridge.models.organizations.Organization;
 import org.sagebionetworks.bridge.validators.Validate;
@@ -45,7 +42,7 @@ import org.sagebionetworks.bridge.validators.Validate;
 public class OrganizationService {
 
     private OrganizationDao orgDao;
-    private AccountDao accountDao;
+    private AccountService accountService;
     private SessionUpdateService sessionUpdateService;
     private AssessmentDao assessmentDao;
     private CacheProvider cacheProvider;
@@ -55,8 +52,8 @@ public class OrganizationService {
         this.orgDao = orgDao;
     }
     @Autowired
-    final void setAccountDao(AccountDao accountDao) {
-        this.accountDao = accountDao;
+    final void setAccountService(AccountService accountService) {
+        this.accountService = accountService;
     }
     @Autowired
     final void setSessionUpdateService(SessionUpdateService sessionUpdateService) {
@@ -202,7 +199,7 @@ public class OrganizationService {
                 .withAdminOnly(null) 
                 .withOrgMembership(identifier).build();
         
-        return accountDao.getPagedAccountSummaries(appId, scopedSearch);
+        return accountService.getPagedAccountSummaries(appId, scopedSearch);
     }
     
     public PagedResourceList<AccountSummary> getUnassignedAdmins(String appId, AccountSummarySearch search) {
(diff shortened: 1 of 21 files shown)


src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -213,50 +211,45 @@ public void deleteOrganization(String appId, String identifier) {
...
-        sessionUpdateService.updateOrgMembership(account.getId(), null);
+
+        AccountId accountId = AccountId.forId(appId, userId);
+        accountService.editAccount(accountId, (acct) -> {
+            if (acct.getOrgMembership() == null || !acct.getOrgMembership().equals(identifier)) {
+                throw new BadRequestException("Account is not a member of organization " + identifier);
+            }
+            acct.setOrgMembership(null);
+        });
+        sessionUpdateService.updateOrgMembership(userId, null);
```
alxdarksage, line 252: So this no longer throws an EntityNotFoundException, and that seems problematic. We may want to throw an exception when you call editAccount.

src/main/java/org/sagebionetworks/bridge/spring/controllers/EmailController.java
```diff
@@ -58,11 +59,14 @@ public String unsubscribeFromEmail() {
                 throw new BadRequestException("Email not found.");
             }
             
-            // This should always return a healthCode under normal circumstances.
-            String healthCode = accountService.getAccountHealthCode(app.getIdentifier(), "email:"+email)
+            // This should always return an ID under normal circumstances. We could eliminate
+            // this and just call editAccount but there would be no bad request if the email
+            // is invalid.
+            String userId = accountService.getAccountId(app.getIdentifier(), "email:"+email)
                     .orElseThrow(() -> new BadRequestException("Email not found."));
```
alxdarksage: An example of why editAccount should probably throw ENFE.

src/main/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentController.java
```diff
@@ -98,13 +98,14 @@ public StatusMessage updateUserEnrollments(@PathVariable String userId) {
         List<EnrollmentMigration> migrations = parseJson(new TypeReference<List<EnrollmentMigration>>() {});
         
         AccountId accountId = BridgeUtils.parseAccountId(session.getAppId(), userId);
-        Account acct = accountService.getAccount(accountId);
-        if (acct == null) {
+        
+        Account account = accountService.getAccount(accountId);
+        if (account == null) {
             throw new EntityNotFoundException(Account.class);
         }
```
alxdarksage: And again here, throwing ENFE

src/main/java/org/sagebionetworks/bridge/spring/controllers/FPHSController.java
```diff
@@ -56,7 +56,7 @@ public StatusMessage registerExternalIdentifier() {
         UserSession session = getAuthenticatedSession();
         
         ExternalIdentifier externalId = parseJson(ExternalIdentifier.class);
-        fphsService.registerExternalIdentifier(session.getAppId(), session.getHealthCode(), externalId);
+        fphsService.registerExternalIdentifier(session.getAppId(), session.getId(), externalId);
```
alxdarksage: Check or revert - we're moving this over to ID, is that happening at every tier?

src/main/java/org/sagebionetworks/bridge/spring/controllers/MembershipController.java
```diff
@@ -51,8 +49,7 @@ public StatusMessage addMember(@PathVariable String orgId, @PathVariable String
         
         // organization membership checked in service
         
-        AccountId accountId = BridgeUtils.parseAccountId(session.getAppId(), userId);
-        organizationService.addMember(session.getAppId(), orgId, accountId);
+        organizationService.addMember(session.getAppId(), orgId, userId);
```
alxdarksage, line 52: These could be reverted I think and you could pass accountId.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityController.java
```diff
@@ -261,8 +262,8 @@ private ScheduleContext getScheduledActivitiesInternal(UserSession session, Date
     }
 
     DateTimeZone persistTimeZone(UserSession session, DateTimeZone timeZone) {
-        accountService.editAccount(session.getAppId(), session.getHealthCode(),
-                account -> account.setTimeZone(timeZone));
+        AccountId accountId = AccountId.forId(session.getAppId(), session.getId());
+        accountService.editAccount(accountId, account -> account.setTimeZone(timeZone));
```
alxdarksage, line 266: Revert healthCode?

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
```diff
@@ -213,50 +211,45 @@ public void deleteOrganization(String appId, String identifier) {
...
+        AccountId accountId = AccountId.forId(appId, userId);
+        accountService.editAccount(accountId, (acct) -> {
+            RequestContext context = RequestContext.get();
+            if (!context.isInRole(ADMIN) && acct.getOrgMembership() != null) {
+                throw new BadRequestException("Account already assigned to an organization.");
+            }
+            acct.setOrgMembership(identifier);
+        });
+        sessionUpdateService.updateOrgMembership(userId, identifier);
     }
```
alxdarksage, line 236: The use of userId in this last statement is why we switched to explicitly passing it in...accountId opens the door to another credential being used, but here, we really need userId.

src/main/java/org/sagebionetworks/bridge/services/OrganizationService.java
alxdarksage: We need userId here. we are also not throwing ENFE from editAccount()

src/main/java/org/sagebionetworks/bridge/spring/controllers/EmailController.java
alxdarksage: Cleaned up to use email directly with editAccount() throwing ENFE if warranted.

src/main/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentController.java
alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/spring/controllers/FPHSController.java
alxdarksage: reverted

src/main/java/org/sagebionetworks/bridge/spring/controllers/MembershipController.java
alxdarksage: No, must be kept as userId

src/main/java/org/sagebionetworks/bridge/spring/controllers/ScheduledActivityController.java
alxdarksage: reverted

jnelsonjava (approved):

PR #406, opened 2021-07-20
getAccount(): Account refactored to getAccount(): Optional<Account>

alxdarksage:
This makes getAccount() similar to getAccountNoFilter() and reflects the fact that we do different things after this call..sometimes we throw exceptions, sometimes we ignore a missing account. This is one step further to removing getAccountNoFilter() in favor of a getAccount() method that enforces all necessary authorization constraints.


src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -327,11 +327,10 @@ public void filterForStudyAccountNullReturnsNull() {
         assertNull(BridgeUtils.filterForStudy((Account)null));
     }
     
-    @Test
-    public void filterForStudyAccountNoContextNoStudyDoesNotReturnAccount() {
-        RequestContext.set(new RequestContext.Builder().withCallerUserId(TEST_USER_ID).build());
-        assertNull(BridgeUtils.filterForStudy(getAccountWithStudy()));
-    }
```
alxdarksage, line 334: Look at this again, you put a context here despite it saying this is a no context test.

src/test/java/org/sagebionetworks/bridge/hibernate/HibernateAccountDaoTest.java
```diff
@@ -575,7 +575,7 @@ public void getPagedReturnsQueryParams() throws Exception {
     @Test
     public void getPagedRemovesStudiesNotInCaller() throws Exception {
         RequestContext.set(new RequestContext.Builder()
-                .withCallerUserId(TEST_USER_ID) // it's not the same as ACCOUNT_ID
+                .withCallerUserId(TEST_USER_ID)
```
alxdarksage: restore comment

src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: Added back a test that *really* has a null context, and verified that it returns the account in that case.

jnelsonjava (approved):

PR #409, opened 2021-07-21
Permission changes

alxdarksage:
Some changes in security to accommodate the MTB designs:

- Any administrative account can get all organizations, but only admins will see all organizations, everyone else gets back only their own organization;
- Admins can create an account in any organization when they create an account (very useful for me when onboarding accounts);
- Everyone can see the members of their own organization (and sometimes they can edit these accounts, depending on their own role).
- Study designers can see the studies their organization sponsors

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -21,13 +21,17 @@ import java.util.Set;
  * All methods throw UnauthorizedException if they fail.
  */
 public class AuthUtils {
+    
+    public static final AuthEvaluator CAN_READ_ORGANIZATIONS = new AuthEvaluator()
+            .hasAnyRole(ADMIN);
+
     /**
      * Is this scoped to specific studies? It should have one of the study-scoped
      * roles, and no roles that are app scoped that we would allow wider latitude
      * when using the APIs.
      */
     public static final AuthEvaluator CAN_READ_ORG_SPONSORED_STUDIES = new AuthEvaluator()
-            .hasAnyRole(ORG_ADMIN, STUDY_COORDINATOR).hasNoRole(DEVELOPER, RESEARCHER, ADMIN, WORKER);
+            .hasAnyRole(ORG_ADMIN, STUDY_DESIGNER, STUDY_COORDINATOR).hasNoRole(DEVELOPER, RESEARCHER, ADMIN, WORKER);
 
     /**
      * Can the caller edit assessments? Must be a study designer in the organization that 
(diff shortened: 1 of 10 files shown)


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -21,20 +21,18 @@
...
      */
     public static final AuthEvaluator CAN_READ_ORG_SPONSORED_STUDIES = new AuthEvaluator()
-            .hasAnyRole(ORG_ADMIN, STUDY_COORDINATOR).hasNoRole(DEVELOPER, RESEARCHER, ADMIN, WORKER);
+            .hasAnyRole(ORG_ADMIN, STUDY_DESIGNER, STUDY_COORDINATOR).hasNoRole(DEVELOPER, RESEARCHER, ADMIN, WORKER);
 
-    /**
-     * Can the caller delete an organization?
-     */
-    public static final AuthEvaluator CAN_DELETE_ORG = new AuthEvaluator()
-            .hasAnyRole(ADMIN);
```
alxdarksage, line 36: Why add CAN_READ_ORGANIZATION and then remove this?

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
```diff
@@ -640,4 +641,20 @@ public void canEditSchedulesFailsForNonDeveloper() {
...
+    }
+
+    @Test(expectedExceptions = UnauthorizedException.class)
+    public void canReadOrganizationsFailsForOrgAdmin() {
+        RequestContext.set(new RequestContext.Builder()
+                .withCallerRoles(ImmutableSet.of(ORG_ADMIN)).build());
+                
+        CAN_READ_ORGANIZATIONS.checkAndThrow();
+    }
+}
```
alxdarksage, line 673: We added more than this, test all of them, even the trivial ones (ADMIN test).

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: It was removed because its one condition is easily verified by getAuthenticatedSession(ADMIN) in the controller. CAN_READ_ORGANIZATIONS is just an admin check so that's what I'm going to use there, too.

src/test/java/org/sagebionetworks/bridge/AuthUtilsTest.java
alxdarksage: Added more tests

nategbrown9 (approved):
Looks good.

PR #411, opened 2021-08-11
Client time zone

jnelsonjava:
For [BRIDGE-3047](https://sagebionetworks.jira.com/browse/BRIDGE-3047).

Adding clientTimeZone field to StudyParticipant as a reference for the participant's local time zone.

~~Still need to add unit tests so this is not ready for a final review. But the logic should be in place for initial checks.~~


src/main/java/org/sagebionetworks/bridge/hibernate/DateTimeZoneIdConverter.java
```diff
@@ -0,0 +1,21 @@
...
+        if (dateTimeZone == null) return null;
+        return dateTimeZone.getID();
+    }
+
+    @Override
+    public DateTimeZone convertToEntityAttribute(String s) {
+        if (s == null) return null;
+        return DateTimeZone.forID(s);
+    }
+}
```
alxdarksage: This already exists as DateTimeZoneAttributeConverter

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -609,6 +610,8 @@ private void updateAccountAndRoles(App app, Account account, StudyParticipant pa
             account.setNote(participant.getNote());
         }
         account.setMigrationVersion(MIGRATION_VERSION);
+        // TODO: check if this needs to retain previous value if it's getting a null from participant
+        account.setClientTimeZone(participant.getClientTimeZone());
```
alxdarksage, line 613: From all discussions, this is a field that can be set by the participant or a researchers, it can be added/changed/deleted, so no special anything required for it.

src/main/resources/db/changelog/changelog.sql
```diff
@@ -753,3 +753,8 @@ ADD COLUMN `logoGuid` varchar(255);
 
 ALTER TABLE `Substudies`
 MODIFY COLUMN `phase` enum('LEGACY', 'DESIGN', 'RECRUITMENT', 'IN_FLIGHT', 'ANALYSIS', 'COMPLETED', 'WITHDRAWN') DEFAULT 'LEGACY';
+
+-- changeset bridge:42
+
+ALTER TABLE `Accounts`
+ADD COLUMN `clientTimeZone` varchar(64) DEFAULT NULL;
```
alxdarksage: Everywhere else this is 255... which might have just been done on autopilot. Does 64 cover all possible values?

src/main/java/org/sagebionetworks/bridge/hibernate/DateTimeZoneIdConverter.java
jnelsonjava: I had an issue trying to get it to work with DateTimeZoneAttributeConverter. That seems to only work with offset strings because it uses bridge-base's DateUtils.parseZoneFromOffsetString(). So when it parses something like "America/Los_Angeles", it throws an error saying it needs to be in "+00:00" format.

src/main/resources/db/changelog/changelog.sql
jnelsonjava: I think 64 is safe. The longest zone I could find was 32 characters (America/Argentina/ComodRivadavia).

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Ah! You are right. So far we've stored this value as a string, and just validated that it can be converted to a valid DateTimeZone object in the validators. And you could do this here...the key thing is that in the database column, the value should be the region string, such as "America/Los_Angeles".

alxdarksage: When you get to testing JSON serialization, you may find it's easier to store it as a string...

jnelsonjava: I see what you mean. I'll definitely switch it to a string and just validate it.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAccount.java
```diff
@@ -543,4 +544,16 @@ public String getNote() {
     public void setNote(String note) {
         this.note = note;
     }
+
+    /** The client's time zone explicitly set through requests and used to calculate
+     * schedules for the user. Can be updated or deleted. */
```
jnelsonjava: Not used to calculate the schedule in the server, just for the client side as reference.

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -263,6 +265,7 @@ public void createAccount() {
         assertEquals(createdAccount.getStatus(), UNVERIFIED);
         assertEquals(createdAccount.getMigrationVersion(), MIGRATION_VERSION);
         assertEquals(createdAccount.getNote(), TEST_NOTE);
+        assertEquals(createdAccount.getClientTimeZone(), TEST_CLIENT_TIME_ZONE);
```
jnelsonjava, line 269: Need to add a check to the updateSuccess method to verify that this field is editable.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -20,6 +20,7 @@
 import static org.sagebionetworks.bridge.TestConstants.TEST_USER_ID;
```
jnelsonjava, line 20: Add a test to check that clientTimeZone can be updated.

src/main/java/org/sagebionetworks/bridge/validators/StudyParticipantValidator.java
```diff
@@ -128,6 +131,13 @@ public void validate(Object object, Errors errors) {
                 errors.rejectValue("attributes", messageForSet(app.getUserProfileAttributes(), attributeName));
             }
         }
+        if (participant.getClientTimeZone() != null) {
+            try {
+                DateTimeZone.forID(participant.getClientTimeZone());
```
jnelsonjava: Try using ZoneId type instead of DateTimeZone.

src/test/java/org/sagebionetworks/bridge/validators/StudyParticipantValidatorTest.java
```diff
@@ -5,10 +5,12 @@
 import static org.sagebionetworks.bridge.TestConstants.TEST_APP_ID;
 import static org.sagebionetworks.bridge.TestConstants.TEST_ORG_ID;
 import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY_ID;
+import static org.sagebionetworks.bridge.TestConstants.TEST_CLIENT_TIME_ZONE;
 import static org.sagebionetworks.bridge.TestUtils.assertValidatorMessage;
 import static org.sagebionetworks.bridge.validators.Validate.CANNOT_BE_BLANK;
 import static org.sagebionetworks.bridge.validators.Validate.INVALID_EMAIL_ERROR;
 import static org.sagebionetworks.bridge.validators.Validate.INVALID_PHONE_ERROR;
+import static org.sagebionetworks.bridge.validators.Validate.TIME_ZONE_ERROR;
```
jnelsonjava, line 13: Consider adding example format to the TIME_ZONE_ERROR constant.

alxdarksage (approved):

PR #413, opened 2021-08-12
design -> withdrawn publishes the schedule to freeze it.

alxdarksage:
BRIDGE-3064, if you take a study in design and withdraw it, you can currently still change the schedule. I think you want to freeze it entirely, so publish the schedule when transitioning to withdrawn.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -286,10 +286,18 @@ public class StudyService {
     /**
      * Moves a study to the withdrawn state. No further enrollments will be allowed,
      * and data collected for this study should not be uploaded as part of the data
-     * set.  
+     * set. If this was previously in design, the schedule will be published in order
+     * to freeze it from further changes.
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -286,10 +286,18 @@ public Study transitionToCompleted(String appId, String studyId) {
...
-        return phaseTransition(appId, studyId, WITHDRAWN, null);
+        return phaseTransition(appId, studyId, WITHDRAWN, (study) -> {
+            if (study.getScheduleGuid() != null) {
+                Schedule2 schedule = scheduleService.getSchedule(appId, study.getScheduleGuid());
+                if (!schedule.isPublished()) {
+                    scheduleService.publishSchedule(appId, study.getScheduleGuid());
+                }
+            }
+        });
     }
```
alxdarksage, line 301: Update the state diagram documentation.

nategbrown9 (approved):
Looks good.

PR #416, opened 2021-08-14
Adding custom events to study.

alxdarksage:


src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -179,10 +179,12 @@ public VersionHolder updateStudy(String appId, Study study) {
...
         }
-        if (!CAN_EDIT_STUDY_CORE.contains(existing.getPhase()) &&
-                !Objects.equals(study.getScheduleGuid(), existing.getScheduleGuid())) {
-            throw new BadRequestException("Study schedule cannot be changed or removed during phase " 
-                    + existing.getPhase().label() + ".");
+        if (!CAN_EDIT_STUDY_CORE.contains(existing.getPhase())) {
+            if(!Objects.equals(study.getScheduleGuid(), existing.getScheduleGuid())) {
+                throw new BadRequestException("Study schedule cannot be changed or removed during phase " 
+                        + existing.getPhase().label() + ".");
+            }
```
alxdarksage: This could arguably just be setting it back to the existing value, rather than throwing an error.

nategbrown9 (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: Changed this.

PR #417, opened 2021-08-17
Use custom events as they are defined on studies for study-scoped event APIs (not app-scoped custom events)

alxdarksage:
BRIDGE-3044. Automatic custom events will be replaced with some kind of support for study bursts, that will probably generate these events dynamically without persisting them.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/hibernate/HibernateStudyTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/hibernate/HibernateStudyTest.java
@@ -7,6 +7,7 @@ import org.testng.annotations.Test;
 
 import org.sagebionetworks.bridge.TestUtils;
 import org.sagebionetworks.bridge.json.BridgeObjectMapper;
+import org.sagebionetworks.bridge.models.activities.ActivityEventUpdateType;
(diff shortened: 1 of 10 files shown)


src/test/java/org/sagebionetworks/bridge/hibernate/HibernateStudyTest.java
```diff
@@ -44,6 +47,18 @@
...
+    public void getCustomEventsMap() {
+        Study study = Study.create();
+        study.setCustomEvents(ImmutableList.of(new StudyCustomEvent("event1", IMMUTABLE), 
+                new StudyCustomEvent("event2", FUTURE_ONLY)));
+        
+        Map<String, ActivityEventUpdateType> map = study.getCustomEventsMap();
+        assertEquals(map.size(), 2);
+        assertEquals(map.get("event1"), IMMUTABLE);
+        assertEquals(map.get("event2"), FUTURE_ONLY);
+    }
```
alxdarksage, line 64: test the empty case on a new object.

alxdarksage: Added

jnelsonjava (approved):
Looks good!

PR #418, opened 2021-08-18
Multiple events assignable to each schedule session.

alxdarksage:
BRIDGE-3059. In addition, the timeline structure changes so that the event ID for a specific session instance is in the ScheduledSession object, not the SessionInfo object.


jnelsonjava (approved):
Looks good!

PR #424, opened 2021-09-10
Give developer accounts access to test accounts

alxdarksage:
BRIDGE-3053

Developers can now do whatever researchers can do (and same with the study-scoped roles), but:

- Creating accounts, their accounts are always test accounts;
- Querying for accounts, they can only retrieve test accounts (allOfGroups = test_user is always added to queries);
- Operating on accounts, they can only access test accounts.

This effectively means that all the researcher/study_coordinator APIs are accessible to developers/study_coordinators, with these caveats.


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -235,4 +250,45 @@ public static boolean isInRole(Set<Roles> callerRoles, Set<Roles> requiredRoles)
...
+        }
+        // If after removing all enrollments that are not visible to the caller, there are no remaining 
+        // enrollments, the caller cannot access the account.
+        RequestContext context = RequestContext.get();
+        Set<String> callerStudies = context.getOrgSponsoredStudies();
+        Set<Enrollment> removals = account.getEnrollments().stream()
+                .filter(en -> !callerStudies.contains(en.getStudyId()))
+                .collect(toSet());
+        account.getEnrollments().removeAll(removals);
+        return !account.getEnrollments().isEmpty();
```
alxdarksage: An effect here would be that developers not associated to an organization would not be able to see test accounts, because they have no relationship to the study.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -266,7 +273,10 @@ public void updateAccount(Account account) {
         if (!RequestContext.get().isAdministrator()) {
             account.setNote(persistedAccount.getNote());
         }
-
+        if (IS_ONLY_DEVELOPER.check()) {
+            Set<String> newDataGroups = addToSet(account.getDataGroups(), TEST_USER_GROUP);
+            account.setDataGroups(newDataGroups);
+        }
```
alxdarksage: This is probably redundant and be removed because getAccount(...) was already called, or it should throw UnauthorizedException.

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: No, not true. A developer passes the AuthUtils check above as a developer. It never gets to that check if the account is not a test account. The end result is that developers can access all test accounts in an app. The is covered by test canAccessAccount_devSucceedsWithTestAccount().

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: This was a definite security hole as the account was being loaded via the DAO with no further security checks. This was changed to throw an UnauthorizedException if the caller is a dev and the account is not a test account, and I added six tests for both create and update to verify we're getting the right behavior.

src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
```diff
@@ -115,6 +133,15 @@ public AuthEvaluator isSelf() {
         return this;
     }
 
+    public AuthEvaluator isNotSelf() {
```
DwayneJengSage, line 136: Unit tests for isNotSelf?

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
DwayneJengSage: It looks like CAN_READ_PARTICIPANT already includes a canAccessStudy() check. Is this check for enrollments redundant with that check? In what cases would the enrollments check be different than the canAccessStudy() check?

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -251,6 +258,11 @@ public void updateAccount(Account account) {
         // Can't change app, email, phone, emailVerified, phoneVerified, createdOn, or passwordModifiedOn.
         Account persistedAccount = accountDao.getAccount(accountId)
                 .orElseThrow(() -> new EntityNotFoundException(Account.class));
+
+        // If a developer is trying to update a production account, that is not allowed.
+        if (IS_ONLY_DEVELOPER.check() && !account.getDataGroups().contains(TEST_USER_GROUP)) {
+            throw new UnauthorizedException();
+        }
```
DwayneJengSage: I don't think this comment is quite accurate. "account" is the updated version of the account that we're trying to save. "persistedAccount" is the existing version of the account from the database. This check would throw if the developer is trying to remove the test_user data group.

Also, do we still want to check if persistedAccount contains the test_user data group? updateAccount() is called in a lot of places, and instead of relying that the caller has already called accountService.getAccount(), it would probably be safer to do the check ourselves.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
```diff
@@ -425,6 +428,10 @@ public StudyParticipant getParticipant(App app, Account account, boolean include
         
         Validate.entityThrowingException(new AccountSummarySearchValidator(app.getDataGroups()), search);
         
+        if (IS_ONLY_DEVELOPER.check()) {
+            Set<String> newDataGroups = addToSet(search.getAllOfGroups(), TEST_USER_GROUP);
+            search = search.toBuilder().withAllOfGroups(newDataGroups).build();
+        }
```
DwayneJengSage, line 434: Add a unit test for this?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/AuthEvaluator.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: canAccessStudy() will only pass if a STUDY_ID is passed in, which it is not in that check, so as it stands canAccessStudy() will never return true for that criterion. Which makes little sense since CAN_READ_PARTICIPANTS is only used in this method call.

I have made these changes which I think is better design:

1) I removed the side effect of altering the enrollments. Since this method is called from one place, getAccount, it is better to do it there where a side effect might be expected;
2) I now iterate through the enrollments if they exist, returning true if any study matches, or false otherwise.
3) For accounts that are not enrolled in any study, they are returned or not based on the other checks of CAN_READ_PARTICIPANTS.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: You are right this needs to be checked against persistedAccount, so I changed that, and if the caller is a developer, I add the test flag so they cannot remove it (but others can still change this after the fact, as before). Added test to verify this in AccountServiceTest.

src/main/java/org/sagebionetworks/bridge/services/ParticipantService.java
alxdarksage: Added

DwayneJengSage (approved):

PR #427, opened 2021-09-14
Fixing how a couple of error conditions are reported back to the user.

alxdarksage:
1) enrolling someone in a non-existent study gave a strange exception, but now returns ENFE for a study (BRIDGE-3058);
2) deleting an account that was used to enroll another account threw a cryptic exception, it now explains that the account is being used that way and so cannot be deleted (found by me and fixed before filed).


jnelsonjava (approved):
Looks good to me!

PR #428, opened 2021-09-16
Making app delete more robust

alxdarksage:
BRIDGE-3065.

The first issue is that a number of models have been added that are not deleted when an app is deleted, and these need to be deleted in a certain order. Because app is in Dynamo and many of these objects are in MySQL, there is no transaction around these deletes. I moved the deletion of the app to last so I could retry until the delete was completely successfully.

I also fixed up /v1/apps to separate the "summary" and "includeDeleted" flags in order to support a reasonable UI for logically deleted apps.

AssessmentValidator did not return an invalid entity exception when the organization was wrong or missing (it threw an ENFE for the organization instead). This is now fixed.


nategbrown9 (approved):
Looks good.

PR #430, opened 2021-09-17
Allowing physical delete of studies for other roles than admin.

alxdarksage:
BRIDGE-3083. Logically deleted studies are accumulating in MTB, polluting the ID namespace. These can in many cases be physically deleted, so I have changed the logic to allow devs/study designers to do it, looking at the study phase to prevent the deletion of studies that are in use.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -230,8 +230,6 @@ public class StudyService {
             throw new BadRequestException("Study cannot be deleted during phase " 
                     + existing.getPhase().label());
         }
-        
-        // Throws exception if the element does not exist.
         String scheduleGuid = existing.getScheduleGuid();
         
         studyDao.deleteStudyPermanently(appId, studyId);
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -221,9 +223,16 @@ public void deleteStudyPermanently(String appId, String studyId) {
...
         
+        Study existing = getStudy(appId, studyId, true);
+        
+        RequestContext context = RequestContext.get();
+        if (!CAN_DELETE_STUDY.contains(existing.getPhase()) && !context.isInRole(ADMIN)) {
+            throw new BadRequestException("Study cannot be deleted during phase " 
+                    + existing.getPhase().label());
+        }
+        
         // Throws exception if the element does not exist.
```
nategbrown9: Does this comment apply to the call got getStudy? Which got moved up.

nategbrown9 (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
alxdarksage: I don't know what "element" referred to. Removing.

PR #436, opened 2021-09-28
Study bursts 2

alxdarksage:
Refactoring the way events are communicated through the system so they are more flexible and can deal with study bursts, which are another form of custom event. 

Responsibilities refactored:

- ActivityEventRequest - receive back events as they are communicated through the API, and parse those strings into the component parts of
- ActivityEventRequestBuilder - a construct that allows internal calls to work directly with a typed description of an event;
- ActivityEventRequest - the internal object we are persisting (with an always-correct event ID string).


src/main/java/org/sagebionetworks/bridge/models/activities/StudyActivityEvent.java
```diff
@@ -42,55 +50,54 @@
...
-    public StudyActivityEvent() {}
+    // Hibernate is happy with default (package) visibility constructor and setters, ensuring
+    // that objects can only be constructed in code through a builder.
     
-    public StudyActivityEvent(String eventId, DateTime timestamp) {
-        this.eventId = eventId;
-        this.timestamp = timestamp;
-        this.recordCount = 1;
-    }
+    StudyActivityEvent() {}
```
alxdarksage: What does Hibernate really need here? Remove more if possible.

src/test/java/org/sagebionetworks/bridge/models/activities/StudyActivityEventRequestTest.java
```diff
@@ -71,101 +48,76 @@ public void canDeserialize() throws Exception {
...
+    public static Object[][] parseRequestParams() {
+        return new Object[][] {
+            {"", null, m(), m()},
+            {"activities_retrieved", "activities_retrieved", m("foo"), m()},
+            {"ACTIVITIES_RETRIEVED", "activities_retrieved", m("foo"), m()},
+            {"custom:foo", "custom:foo", m("foo"), m()},
+            {"custom:foo", null, m(), m("foo")},
+            {"custom:foo", null, m(), m()},
+            {"custom:TIMELINE_RETRIEVED", null, m("timeline_retrieved"), m()},
+            {"custom:timeline_retrieved", null, m(), m()},
```
alxdarksage, line 65: Add custom event timeline retrieved that succeeds

src/main/java/org/sagebionetworks/bridge/models/activities/StudyActivityEvent.java
alxdarksage: No it doesn't need any of it. When you put the annotations on the field, it can set them even without accessors. All removed.

src/test/java/org/sagebionetworks/bridge/models/activities/StudyActivityEventRequestTest.java
alxdarksage: Added

nategbrown9 (approved):
Looks good.

PR #439, opened 2021-10-01
Make it a little easier to delete an unused account.

alxdarksage:
You can now delete an account enrolled in multiple studies as long as you have access to all the studies (and it hasn't been used). Also clarified this rule in the error message if you are not allowed to delete the account. Test accounts continue to be deletable no matter what.


nategbrown9 (approved):
Looks good.

PR #446, opened 2021-10-08
Bridge 3101

alxdarksage:
IS_ONLY_DEVELOPER was listing developer roles rather than listing roles with production participant access, which is brittle and was allowing org admin's access to accounts. Changed this, which picks up org admins, but then that role needs some help to still be able to access and edit administrative accounts. Tests in AuthUtils.canAccessAccount and AccountService need heuristics to determine an admin account (at least one role). Finally, the AccountController's verifyOrgAdminIsActingOnOrgMember method is almost entirely redundant with the more standard CAN_EDIT_ACCOUNTS check, so it was largely removed. Tests added or changed as appropriate.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -35,6 +35,9 @@ public class AuthUtils {
      */
     public static final AuthEvaluator IS_ONLY_DEVELOPER = new AuthEvaluator()
             .isNotSelf().hasOnlyRoles(DEVELOPER, STUDY_DESIGNER);
+
+    public static final AuthEvaluator CANNOT_SEE_PROD_PARTICIPANTS = new AuthEvaluator()
+            .isNotSelf().hasOnlyRoles(DEVELOPER, STUDY_DESIGNER, ORG_ADMIN);
     
     public static final AuthEvaluator CAN_TRANSITION_STUDY = new AuthEvaluator()
             .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
@@ -112,7 +115,7 @@ public class AuthUtils {
     public static final AuthEvaluator CAN_READ_PARTICIPANTS = new AuthEvaluator().isSelf().or()
             // This allows an org admin to see study participants, but we block this in the relevant API calls.
             .isInOrg().hasAnyRole(ORG_ADMIN).or()
-            .canAccessStudy().hasAnyRole(STUDY_DESIGNER, STUDY_COORDINATOR).or()
+            .canAccessStudy().hasAnyRole(STUDY_DESIGNER, STUDY_COORDINATOR, ORG_ADMIN).or()
             .hasAnyRole(DEVELOPER, RESEARCHER, WORKER, ADMIN);
     
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/org/sagebionetworks/bridge/AuthUtils.java
```diff
@@ -34,8 +34,8 @@
      * accounts, not production accounts.
      */
     public static final AuthEvaluator IS_ONLY_DEVELOPER = new AuthEvaluator()
```
alxdarksage: Name: this person does not have rights to access other accounts...which isn't quite the same as being a developer.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
```diff
@@ -235,23 +232,11 @@ public Account verifyOrgAdminIsActingOnOrgMember(UserSession session, String use
...
-        if (callerOrgId == null) {
-            throw new UnauthorizedException();
-        }
-        // The caller needs to be an administrator of the account's organization
-        if (!callerOrgId.equals(account.getOrgMembership())) {
+        
+        // The caller needs to have permissions to update the account
+        if (!CAN_EDIT_ACCOUNTS.check(ORG_ID, account.getOrgMembership(), USER_ID, account.getId())) {
             throw new UnauthorizedException();
         }
```
alxdarksage: This could just check and throw, remove if statement

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -432,7 +433,41 @@ public void updateAccountSucceedsForStudyDesignerUpdatingTestAccount() throws Ex
     }
     
     @Test
-    public void updateAccountNotFound() {
+    public void updateAccountSucceedsForOrgAdminUpdatingAdminAccount() throws Exception {
+        RequestContext.set(new RequestContext.Builder()
+                .withCallerUserId("id")
+                // .withOrgSponsoredStudies(ImmutableSet.of(TEST_STUDY_ID))
```
alxdarksage: Remove commented out code

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
```diff
@@ -432,7 +433,41 @@ public void updateAccountSucceedsForStudyDesignerUpdatingTestAccount() throws Ex
...
+                .withCallerRoles(ImmutableSet.of(ORG_ADMIN)).build());
+        
+        Account account = mockGetAccountById(ACCOUNT_ID, false);
+        // account.setRoles(ImmutableSet.of(STUDY_DESIGNER));
+        account.setDataGroups(ImmutableSet.of(TEST_USER_GROUP));
+        account.setEnrollments(ImmutableSet.of(Enrollment.create(TEST_APP_ID, TEST_STUDY_ID, TEST_USER_ID)));
+
+        service.updateAccount(account);
+        
+        verify(mockAccountDao).updateAccount(account);
```
alxdarksage: This succeeds, this doesn't fail. Is this correct?

src/main/java/org/sagebionetworks/bridge/AuthUtils.java
alxdarksage: Renamed to CANNOT_ACCESS_PROD_PARTICIPANTS

src/main/java/org/sagebionetworks/bridge/spring/controllers/AccountsController.java
alxdarksage: Cleaned up

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Removed

src/test/java/org/sagebionetworks/bridge/services/AccountServiceTest.java
alxdarksage: Yep the service needed to be changed. It is changed. Long-term as mentioned, the solution is to have an AccountService for admin accounts and a ParticipantService for participants, and not allow them to cross.

jnelsonjava (approved):
Looks good.

PR #447, opened 2021-10-11
BRIDGE-3036: Filter enrollments to return only studies the caller can access

alxdarksage:


src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -109,8 +113,10 @@ protected DateTime getWithdrawalDateTime() {
                 .orElseThrow(() -> new EntityNotFoundException(Account.class));
 
         CAN_EDIT_ENROLLMENTS.checkAndThrow(STUDY_ID, studyId, USER_ID, account.getId());
```
alxdarksage, line 119: If this account is not a test account, and the caller is "only a developer", then throw an error.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -109,8 +113,10 @@ protected DateTime getWithdrawalDateTime() {
...
-
-        return enrollmentDao.getEnrollmentsForUser(appId, account.getId());
+        
+        RequestContext context = RequestContext.get();
```
alxdarksage, line 117: If you are not associated to an organization, you are going to see all studies, and that is bad.

alxdarksage: And in the case of getting studies, what happens if the caller is not in an organization?

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -109,8 +113,10 @@ protected DateTime getWithdrawalDateTime() {
...
+        Set<String> studyIds = context.isInRole(ADMIN) ? ImmutableSet.of() : context.getOrgSponsoredStudies();
+        return enrollmentDao.getEnrollmentsForUser(appId, studyIds, account.getId());
```
alxdarksage, line 125: You can add researcher to see all studies, that's consistent with researchers and also, addresses the case of not being a member of the organization.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: This happens in the above check. Because this was complicated enough that I had to sit and puzzle it out during a code review, I've added some comments to the code explaining what each section is checking (security-wise).

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
alxdarksage: Added developer and researcher so they consistently can see all the studies (they can access all studies, so this works).

jnelsonjava (approved):

PR #448, opened 2021-10-11
Enrollment note

jnelsonjava:
For BRIDGE-3095

- Add note field to enrollment records for a study-scoped participant note
- Create route to update existing enrollments
- Allow note to be set on enrollment
- Prevent users from setting notes on own enrollments

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
@@ -93,4 +93,11 @@ public interface Enrollment extends BridgeEntity {
      */
     String getWithdrawalNote();
     void setWithdrawalNote(String note);
+
+    // TODO: add description of note field's use
+    /**
+     *
+     */
+    String getNote();
+    void setNote(String note);
 }
--- a/src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
@@ -94,9 +94,8 @@ public interface Enrollment extends BridgeEntity {
     String getWithdrawalNote();
     void setWithdrawalNote(String note);
 
-    // TODO: add description of note field's use
-    /**
-     *
+    /** Note field is accessible only by users with administrative roles. The enrolled participant
+     * can not get or set the note.
      */
     String getNote();
     void setNote(String note);
--- a/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
@@ -238,35 +238,24 @@ public class EnrollmentService {
         throw new EntityNotFoundException(Enrollment.class);
     }
 
-    // TODO: Add an update method that updates editable fields of an enrollment record
-    //       specifically in order to update the note field. (Could just make it specifically only
-    //       retrieve the previous record, update the note field, and save)
-    //       - Also note the method name could change to updateEnrollment but would conflict
-    //         with existing method above.
     public void editEnrollment(Enrollment enrollment) {
         checkNotNull(enrollment);
 
         Validate.entityThrowingException(INSTANCE, enrollment);
 
+        AccountId accountId = AccountId.forId(enrollment.getAppId(), enrollment.getAccountId());
+        Account account = accountService.getAccount(accountId)
+                .orElseThrow(() -> new EntityNotFoundException(Account.class));
+
         CAN_EDIT_OTHER_ENROLLMENTS.checkAndThrow(STUDY_ID, enrollment.getStudyId(), USER_ID, enrollment.getAccountId());
 
-        AccountId accountId = AccountId.forId(enrollment.getAppId(), enrollment.getAccountId());
-        Optional<Account> optionalAccount = accountService.getAccount(accountId);
-
-        if (optionalAccount.isPresent()) {
-            Account account = optionalAccount.get();
-
-            if (!account.getDataGroups().contains(TEST_USER_GROUP)) {
-                IS_ONLY_DEVELOPER.checkAndThrow();
+        for (Enrollment accountEnrollment : account.getEnrollments()) {
+            if (accountEnrollment.getStudyId().equals(enrollment.getStudyId())) {
+                accountEnrollment.setNote(enrollment.getNote());
+                break;
             }
-            for (Enrollment accountEnrollment : account.getEnrollments()) {
-                if (accountEnrollment.getStudyId().equals(enrollment.getStudyId())) {
-                    accountEnrollment.setNote(enrollment.getNote());
-                    break;
-                }
-            }
-
-            accountService.updateAccount(account);
         }
+
+        accountService.updateAccount(account);
     }
 }
--- a/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
@@ -238,8 +238,15 @@ public class EnrollmentService {
     // TODO: Add an update method that updates editable fields of an enrollment record
     //       specifically in order to update the note field. (Could just make it specifically only
     //       retrieve the previous record, update the note field, and save)
-    public void updateEnrollmentNote(Enrollment enrollment) {
+    //       - Also note the method name could change to updateEnrollment but would conflict
+    //         with existing method above.
+    public void editEnrollment(Enrollment enrollment) {
         checkNotNull(enrollment);
+
+        Validate.entityThrowingException(INSTANCE, enrollment);
+
+        // TODO: Check on these permissions, is isSelf() applicable
+        CAN_EDIT_ENROLLMENTS.checkAndThrow(STUDY_ID, enrollment.getStudyId(), USER_ID, enrollment.getAccountId());
 
         AccountId accountId = AccountId.forId(enrollment.getAppId(), enrollment.getAccountId());
         Optional<Account> optionalAccount = accountService.getAccount(accountId);
@@ -249,6 +256,7 @@ public class EnrollmentService {
             for (Enrollment accountEnrollment : account.getEnrollments()) {
                 if (accountEnrollment.getStudyId().equals(enrollment.getStudyId())) {
                     accountEnrollment.setNote(enrollment.getNote());
+                    break;
                 }
             }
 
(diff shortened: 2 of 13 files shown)


src/main/java/org/sagebionetworks/bridge/models/studies/Enrollment.java
```diff
@@ -93,4 +93,11 @@ static Enrollment create(String appId, String studyId, String accountId, String
      */
     String getWithdrawalNote();
     void setWithdrawalNote(String note);
+
+    // TODO: add description of note field's use
+    /**
+     *
```
jnelsonjava: Focus on the security constraints or who can read/write.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -5,16 +5,20 @@
 import static org.sagebionetworks.bridge.AuthEvaluatorField.USER_ID;
 import static org.sagebionetworks.bridge.AuthUtils.CAN_EDIT_STUDY_PARTICIPANTS;
 import static org.sagebionetworks.bridge.AuthUtils.CAN_EDIT_ENROLLMENTS;
+import static org.sagebionetworks.bridge.AuthUtils.CAN_EDIT_OTHER_ENROLLMENTS;
+import static org.sagebionetworks.bridge.AuthUtils.IS_ONLY_DEVELOPER;
```
jnelsonjava: IS_ONLY_DEVELOPER is not used anymore.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -5,16 +5,20 @@
...
 import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
 import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
 import static org.sagebionetworks.bridge.BridgeConstants.NEGATIVE_OFFSET_ERROR;
 import static org.sagebionetworks.bridge.BridgeConstants.PAGE_SIZE_ERROR;
+import static org.sagebionetworks.bridge.BridgeConstants.TEST_USER_GROUP;
 import static org.sagebionetworks.bridge.models.ResourceList.ENROLLMENT_FILTER;
 import static org.sagebionetworks.bridge.models.ResourceList.OFFSET_BY;
 import static org.sagebionetworks.bridge.models.ResourceList.PAGE_SIZE;
 import static org.sagebionetworks.bridge.validators.EnrollmentValidator.INSTANCE;
```
jnelsonjava: in updateEnrollment, allow the note to be set unless "self". Otherwise it should carry over from previous enrollment.

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
```diff
@@ -381,35 +386,39 @@ public void enrollInvalidEnrollment() {
...
-        assertNull(retValue.getWithdrawnBy());
-        assertEquals(retValue.getWithdrawalNote(), "Withdrawal reason");
 
-        verify(mockAccountService).editAccount(any(), any());
-        Enrollment captured = Iterables.getLast(account.getEnrollments(), null);
-        assertEquals(captured.getWithdrawnOn(), MODIFIED_ON.minusHours(1));
-        assertNull(captured.getWithdrawnBy());
-        assertEquals(captured.getWithdrawalNote(), "Withdrawal reason");        
-    }
+    // TODO: recomment in this test
```
jnelsonjava: Figure out why this was failing.

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
```diff
@@ -632,4 +641,90 @@ public void unenrollAccountStudyNotFound() {
...
+
+        service.editEnrollment(incomingEnrollment);
+
+        verify(mockAccountService).updateAccount(account);
+
+        assertEquals(account.getEnrollments().size(), 2);
+        assertTrue(account.getEnrollments().contains(targetEnrollment));
+        assertTrue(account.getEnrollments().contains(otherEnrollment));
+        assertEquals(targetEnrollment.getNote(), TEST_NOTE);
+        assertNull(otherEnrollment.getNote());
```
jnelsonjava: Retrieve the actual enrollments from the account to test this.

src/test/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentControllerTest.java
```diff
@@ -97,6 +100,8 @@ public void verifyAnnotations() throws Exception {
         assertGet(EnrollmentController.class, "getEnrollmentsForStudy");
         assertCreate(EnrollmentController.class, "enroll");
         assertDelete(EnrollmentController.class, "unenroll");
+        // TODO: should this include the updateUserEnrollments route?
```
jnelsonjava: No, don't include it.

src/test/java/org/sagebionetworks/bridge/spring/controllers/EnrollmentControllerTest.java
```diff
@@ -227,5 +232,56 @@ public void updateUserEnrollmentsRemovingAll() throws Exception {
...
+        assertEquals(captured.getAccountId(), TEST_USER_ID);
+        assertEquals(captured.getNote(), TEST_NOTE);
+    }
+
+    @Test
+    public void updateEnrollmentValidatesRoles() throws Exception {
+        UserSession session = new UserSession();
+        session.setAppId(TEST_APP_ID);
+        // TODO: Shouldn't this fail? With no roles in the session?
+        doReturn(session).when(controller).getAdministrativeSession();
```
jnelsonjava: Need to set roles in the RequestContext to trigger the AuthUtils.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -208,6 +209,9 @@ private void updateEnrollment(Account account, Enrollment newEnrollment, Enrollm
         String callerUserId = RequestContext.get().getCallerUserId();
         if (!account.getId().equals(callerUserId)) {
             existingEnrollment.setEnrolledBy(callerUserId);
+            existingEnrollment.setNote(newEnrollment.getNote());
+        } else {
+            newEnrollment.setNote(null);
```
alxdarksage: If the user later withdrew themselves, would this null out the note? If you're concerned about a user setting this field, maybe it needs to be nullified somewhere that clearly only applies to self-enrollment.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
```diff
@@ -258,4 +262,25 @@ public Enrollment unenroll(Account account, Enrollment enrollment) {
         }
         throw new EntityNotFoundException(Enrollment.class);
     }
+
+    public void editEnrollment(Enrollment enrollment) {
```
alxdarksage: It's kind of a drag that the private method is "updateEnrollment" and the public method is "editEnrollment" as elsewhere, we probably call the public method the "update" method. We could swap these.

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
```diff
@@ -382,34 +395,36 @@ public void enrollInvalidEnrollment() {
...
         account.setId(TEST_USER_ID);
         account.setEnrollments(Sets.newHashSet(otherStudy, existing));
         TestUtils.mockEditAccount(mockAccountService, account);
-        
+
         Enrollment enrollment = Enrollment.create(TEST_APP_ID, TEST_STUDY_ID, TEST_USER_ID);
         enrollment.setWithdrawnOn(MODIFIED_ON.minusHours(1));
         enrollment.setWithdrawalNote("Withdrawal reason");
-        
+
```
alxdarksage: Can you set your editor to not change these?

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
```diff
@@ -691,4 +706,96 @@ public void unenrollAccountStudyNotFound() {
...
+        assertTrue(capturedEnrollments.contains(targetEnrollment));
+        assertTrue(capturedEnrollments.contains(otherEnrollment));
+        for (Enrollment captureEnrollment : capturedEnrollments) {
+            if (captureEnrollment.getStudyId().equals(TEST_STUDY_ID)) {
+                assertEquals(captureEnrollment.getNote(), TEST_NOTE);
+            } else {
+                assertNull(captureEnrollment.getNote());
+            }
+        }
+    }
```
alxdarksage, line 793: You only need to test the enrollment you have edited, not all of them.

alxdarksage (changes requested):
Very close to done. Maybe a rename and one question about logic if an enrollment record is changed, will it have the side effect of potentially nullifying any note that is present?

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
jnelsonjava: I think I found the right setting. Should be fixed in the future.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
jnelsonjava: Pulled the note setting out of the conditional.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
jnelsonjava: Went ahead and swapped the names.

src/main/java/org/sagebionetworks/bridge/services/EnrollmentService.java
jnelsonjava: Changed this to accept the update regardless of "self" since the caller is an administrator.

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
jnelsonjava: done

src/test/java/org/sagebionetworks/bridge/services/EnrollmentServiceTest.java
jnelsonjava: Paired it down to find the appropriate enrollment.

alxdarksage (approved):

PR #449, opened 2021-10-12
Bridge to talk to Synapse Dev

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2803

Points Bridge dev and staging to talk to the Synapse dev stack.

Note: I'm still configuring and testing OAuth, so don't merge this yet.

### DIFF
--- a/src/main/resources/BridgeServer2.conf
+++ b/src/main/resources/BridgeServer2.conf
@@ -21,14 +21,18 @@ ses.notification.topic.arn = arn:aws:sns:us-east-1:649232250620:SNSBounces
 synapse.user = yours-synapse-user
 synapse.api.key = yours-synapse-api-key
 
+synapse.aws.account.id = 449435941126
+prod.synapse.aws.account.id = 325565585839
+
+synapse.endpoint = https://repo-dev.dev.sagebase.org/
+prod.synapse.endpoint = https://repo-prod.prod.sagebase.org/
+
 exporter.synapse.user = your-exporter-synapse-user
 exporter.synapse.api.key = your-exporter-synapse-api-key
-local.exporter.synapse.id = 3330889
-dev.exporter.synapse.id = 3330889
-uat.exporter.synapse.id = 3327942
+
+exporter.synapse.id = 3336429
 prod.exporter.synapse.id = 3325672
 test.synapse.user.id = 3348228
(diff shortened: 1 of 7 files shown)


src/main/resources/BridgeServer2.conf
```diff
@@ -78,14 +82,13 @@ local.webservices.url = http://localhost:9000
 route53.zone = ZP0HNVK1V670D
 
 # Synapse Team IDs, used by the BridgeStudyCreator when creating Synapse projects.
-team.bridge.admin = 3388390
-team.bridge.staff = 3388389
+team.bridge.admin = 3336438
+team.bridge.staff = 3336437
```
alxdarksage, line 86: It appears the dev server has all the same accounts and stuff. Why are these changing?

src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
```diff
@@ -50,6 +51,10 @@
     @BeforeMethod
     public void beforeMethod() {
         MockitoAnnotations.initMocks(this);
+
+        // Mock config.
+        when(mockBridgeConfig.get("bucket.prop")).thenReturn(BUCKET_NAME);
+        when(mockBridgeConfig.get(S3Initializer.CONFIG_KEY_SYNAPSE_AWS_ACCOUNT_ID)).thenReturn(SYNAPSE_AWS_ACCOUNT_ID);
     }
```
alxdarksage, line 58: Do we verify that SYNAPSE_AWS_ACCOUNT_ID is in the template?

alxdarksage (approved):
Just a note or two.

src/main/resources/BridgeServer2.conf
DwayneJengSage: These are changing because Dev and Staging used to talk to Synapse Prod. This pull request changes them to talk to Synapse Dev, which means we had to recreate all our resources, and they all have different IDs now.

src/test/java/org/sagebionetworks/bridge/S3InitializerTest.java
DwayneJengSage: Yes, on line 87.

PR #450, opened 2021-10-13
Do return errors if a resend verification fails, and prevent it if the credential is already verified

alxdarksage:


jnelsonjava (approved):
Looks good.

PR #455, opened 2021-10-14
Fix integration of v1 consents with v2 event system.

alxdarksage:
BRIDGE-3111

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
@@ -311,7 +311,8 @@ public class ParticipantServiceTest extends Mockito {
         account.setAppId(TEST_APP_ID);
         account.setHealthCode(HEALTH_CODE);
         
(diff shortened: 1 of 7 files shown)


jnelsonjava (approved):
Looks good. I'm surprised it is so simple.

src/test/java/org/sagebionetworks/bridge/services/ParticipantServiceTest.java
```diff
@@ -447,16 +448,30 @@ public void createParticipantDoesNotAlreadyExistThrowsInvalidEntity() {
     }
     
     @Test
-    public void createParticipantWithEnrollment() {
+    public void createParticipantWithEnrollment_anonymous() {
+        RequestContext.set(RequestContext.get().toBuilder().withCallerUserId(null).build());
```
alxdarksage, line 454: So, if this was done before mockHealthCodeAndAccountRetrieval, how did that clear the userId?!

alxdarksage: This is ok...added a comment explaining that the order of setting up the test (@Before, then the mock method) is correct.

PR #456, opened 2021-10-14
Add participant variables to consent email/sms template

alxdarksage:
BRIDGE-3106. This will allow us to put something in the email itself that we can use to route emails sent to the consent administrator. This appears to be enough to do BRIDGE-3107 too, since we send the exact email to the consent coordinator that we send to end users (there is no other template that needs to be added to the template system).


jnelsonjava (approved):
Looks good!

PR #458, opened 2021-10-20
Removing public enrollment using only external ID

alxdarksage:
- you cannot call sign up with an external ID and password and create an account;
- removed some obsolete properties on apps that controlled earlier versions of our external ID support, that are no longer referenced;

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
@@ -47,7 +47,6 @@ public class BridgeExceptionHandler {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity<String> handleException(HttpServletRequest request, Exception ex) throws JsonProcessingException {
-        ex.printStackTrace();
         logException(request, ex);
 
         return getResult(ex);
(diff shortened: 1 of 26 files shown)


src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
```diff
@@ -47,6 +47,7 @@ void setLogger(Logger log) {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity<String> handleException(HttpServletRequest request, Exception ex) throws JsonProcessingException {
+        ex.printStackTrace();
```
alxdarksage: REMOVE ME

src/test/java/org/sagebionetworks/bridge/models/schedules2/timelines/AssessmentInfoTest.java
```diff
@@ -52,6 +52,7 @@ public void canSerialize() throws Exception {
         AssessmentInfo info = AssessmentInfo.create(ref);
         
         JsonNode node = BridgeObjectMapper.get().valueToTree(info);
+        System.out.println(node.toString());
```
alxdarksage: REMOVE ME

src/main/java/org/sagebionetworks/bridge/spring/handlers/BridgeExceptionHandler.java
alxdarksage: Removed

src/test/java/org/sagebionetworks/bridge/models/schedules2/timelines/AssessmentInfoTest.java
alxdarksage: Removed.

PR #460, opened 2021-10-27
Removed startEventId from SessionInfo

alxdarksage:
This has already been removed from the SDK and from the integration tests, so this is the only change that is necessary. This cannot be merged until the client libraries have been updated.


alxdarksage:
I have no idea why this was closed. This was a mistake.

alxdarksage:
This was already merged. Whew.

PR #462, opened 2021-10-28
Errors for activity apis

alxdarksage:
Wouldn't it be nice if the activity event api told you when an event wasn't going to be updated? Now it can.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -247,7 +247,7 @@ public class AccountService {
                 .withObjectType(ENROLLMENT)
                 .withTimestamp(account.getCreatedOn());
         for (Enrollment en : account.getEnrollments()) {
-            studyActivityEventService.publishEvent(builder.withStudyId(en.getStudyId()).build());
+            studyActivityEventService.publishEvent(builder.withStudyId(en.getStudyId()).build(), false);
         }
     }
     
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -247,7 +247,7 @@ public void createAccount(App app, Account account) {
                 .withObjectType(ENROLLMENT)
                 .withTimestamp(account.getCreatedOn());
         for (Enrollment en : account.getEnrollments()) {
-            studyActivityEventService.publishEvent(builder.withStudyId(en.getStudyId()).build());
+            studyActivityEventService.publishEvent(builder.withStudyId(en.getStudyId()).build(), false);
```
alxdarksage, line 250: Should we log if we don't throw?

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
```diff
@@ -212,11 +237,12 @@ private void addIfPresent(List<StudyActivityEvent> events, Map<String, DateTime>
             events.add(event);
         }
     }
-
+    
     /**
-     * If the triggering event is mutable, these events can be created as well.
+     * If the triggering event is mutable, these events can be created as well. Returns true
+     * if any study burst failed to update, or false otherwise.
```
jnelsonjava: The comment addition might not be necessary since it doesn't return a boolean. Could change it to say it adds the failures to the given list, but you also covered that in the comment on 133.

jnelsonjava (approved):
Approved, only left a note about a comment.

PR #463, opened 2021-10-29
Using a utility method for the common operation of selection an item from a collection

alxdarksage:
Just a thing that has bugged me for a while.


jnelsonjava (approved):
Looks great!

PR #469, opened 2021-11-10
Events mutability enhancements

alxdarksage:
1. Deleting an event that triggers a study burst now also deletes the study burst events;
2. New query parameter introduced, `updateBursts` (default is true). If set to false, an update to a trigger event will not update the study burst events (a creation of the event will still create the study burst events however);
3. persisting `updateType` (once an event is created, its event type cannot be changed by redefining the event).

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/services/StudyActivityEventServiceTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/services/StudyActivityEventServiceTest.java
@@ -462,6 +462,7 @@ public class StudyActivityEventServiceTest extends Mockito {
         service.publishEvent(event, false, false);
         
         verify(mockDao, times(1)).publishEvent(eventCaptor.capture());
+        assertEquals(eventCaptor.getValue().getEventId(), "assessment:foo:finished");
     }
     
     @Test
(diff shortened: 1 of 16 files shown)


src/test/java/org/sagebionetworks/bridge/services/StudyActivityEventServiceTest.java
```diff
@@ -381,6 +431,68 @@ public void publishEvent_publishesStudyBursts() {
...
+        
+        when(mockScheduleService.getScheduleForStudy(TEST_APP_ID, study))
+                .thenReturn(Optional.of(schedule));
+        
+        when(mockDao.getRecentStudyActivityEvent(TEST_USER_ID, TEST_STUDY_ID, "assessment:foo:finished"))
+            .thenReturn(persistedEvent);
+        
+        service.publishEvent(event, false, false);
+        
+        verify(mockDao, times(1)).publishEvent(eventCaptor.capture());
```
alxdarksage, line 464: Verify that one event is the origin event.

alxdarksage: Added

jnelsonjava (approved):
Looks good!

PR #470, opened 2021-11-10
Model and Controllers for Participant Versions

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2605

Updating the account will create a participant version. Participant versions will be exported to Synapse as part of https://sagebionetworks.jira.com/browse/BRIDGE-2606


src/main/java/org/sagebionetworks/bridge/dao/ParticipantVersionDao.java
```diff
@@ -0,0 +1,23 @@
...
+
+public interface ParticipantVersionDao {
+    /** Creates a participant version. */
+    void createParticipantVersion(ParticipantVersion participantVersion);
+
+    /** Delete all participant versions for the given app and health code. This is called by integration tests. */
+    void deleteParticipantVersionsForHealthCode(String appId, String healthCode);
+
+    /** Get all participant versions for health code. */
+    List<ParticipantVersion> getAllParticipantVersionsForHealthCode(String appId, String healthCode);
```
alxdarksage, line 16: Does this need to be paged?

src/main/java/org/sagebionetworks/bridge/dao/ParticipantVersionDao.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /** Retrieves the latest participant version for health code. */
+    Optional<ParticipantVersion> getLatestParticipantVersionForHealthCode(String appId, String healthCode);
```
alxdarksage, line 19: Is this identical to the current state of the participant's records, or the last export before this state was changed?

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantVersionDao.java
```diff
@@ -0,0 +1,96 @@
...
+        DynamoParticipantVersion key = new DynamoParticipantVersion();
+        key.setAppId(appId);
+        key.setHealthCode(healthCode);
+
+        DynamoDBQueryExpression<DynamoParticipantVersion> query = new DynamoDBQueryExpression<DynamoParticipantVersion>()
+                .withHashKeyValues(key);
+        List<DynamoParticipantVersion> participantVersionList = queryHelper(query);
+
+        // Because of typing issues, we need to convert the list.
+        return ImmutableList.copyOf(participantVersionList);
```
alxdarksage, line 58: What if this exceeds one page of returned records? They are currently pretty small but if demographic information is added and they are exported/created with each change, there could be dozens or possibly hundreds of them?

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
```diff
@@ -0,0 +1,168 @@
...
+        if (account.getSharingScope() == SharingScope.NO_SHARING) {
+            // no_sharing means we don't export this to Synapse, which means we can skip making a Participant Version.
+            return;
+        }
+        Optional<Boolean> isConsented = consentService.isConsented(account);
+        if (!isConsented.isPresent() || !isConsented.get()) {
+            // If participant is not consented, don't create a participant version.
+            // If we're not sure whether they're consented, assume that they are not.
+            return;
+        }
```
alxdarksage, line 44: Because most studies don't include signing a consent, I think you'll need to check and see if the user is enrolled. Their records should be exported and made available to everyone who has access to the studies the user is enrolled in. For example in MTB, no participant would pass this test, I don't think.

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    // Compares non-key attributes for participant versions. Returns true if they are the same, false if they are
+    // different.
+    // Package-scoped for unit tests.
+    static boolean isIdenticalParticipantVersion(ParticipantVersion oldVersion, ParticipantVersion newVersion) {
+        Map<String, Object> oldAttrMap = getParticipantVersionAttributes(oldVersion);
+        Map<String, Object> newAttrMap = getParticipantVersionAttributes(newVersion);
+        return oldAttrMap.equals(newAttrMap);
+    }
```
alxdarksage, line 116: Even with EqualsVerifier this is easier?

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
DwayneJengSage: Dwayne read this comment and agrees to switch to Enrollment instead.

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
```diff
@@ -0,0 +1,168 @@
...
+        participantVersion.setLanguages(account.getLanguages());
+        participantVersion.setSharingScope(account.getSharingScope());
+        participantVersion.setTimeZone(account.getClientTimeZone());
+
+        // Convert study memberships into a map.
+        Map<String, String> studyMembershipMap = new HashMap<>();
+        for (Enrollment enrollment : account.getActiveEnrollments()) {
+            studyMembershipMap.put(enrollment.getStudyId(), enrollment.getExternalId());
+        }
+        participantVersion.setStudyMemberships(studyMembershipMap);
```
DwayneJengSage: Look for a utility in BridgeUtils. If not exists, make one.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantVersionController.java
```diff
@@ -0,0 +1,78 @@
...
+    @Autowired
+    public final void setParticipantVersionService(ParticipantVersionService participantVersionService) {
+        this.participantVersionService = participantVersionService;
+    }
+
+    /** Delete all participant versions for the given user. This is called by integration tests. */
+    @DeleteMapping(path="/v1/apps/{appId}/participants/{userIdToken}/versions")
+    public StatusMessage deleteParticipantVersionsForUser(@PathVariable String appId,
+            @PathVariable String userIdToken) {
+        getAuthenticatedSession(ADMIN);
```
DwayneJengSage: Get the appId from the Admin's session, not the URL.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantVersionDao.java
DwayneJengSage: In our discussion earlier, we decided that participants are _probably_ not changing all that often, so there shouldn't be that many versions. Currently, only the integ tests and the planned Backfill/Redrive Worker will use this API. The integ tests have about 5 versions, and the Backfill/Redrive worker isn't coming for a while. If it turns out there are a lot of participant versions, we can easily change this later.

src/main/java/org/sagebionetworks/bridge/dao/ParticipantVersionDao.java
DwayneJengSage: Every change to a participant will create a new participant version, so (barring no_sharing or withdrawn consent), it should match the current participant record.

src/main/java/org/sagebionetworks/bridge/dynamodb/DynamoParticipantVersionDao.java
DwayneJengSage: See above for notes about paging.

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
DwayneJengSage: We want to ignore fields like modifiedOn and participantVersion, which violates the .equals() contract.

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
DwayneJengSage: BridgeUtils.mapStudyMemberships() already exists. Switching to that.

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantVersionController.java
DwayneJengSage: This has been moved to /v3/participants/{userIdToken}/versions

src/main/java/org/sagebionetworks/bridge/spring/controllers/ParticipantVersionController.java
```diff
@@ -68,11 +70,18 @@ public StatusMessage deleteParticipantVersionsForUser(@PathVariable String appId
...
                 () -> new EntityNotFoundException(StudyParticipant.class));
-        return participantVersionService.getParticipantVersion(appId, healthCode, version);
+
+        int versionInt;
+        try {
+            versionInt = Integer.parseInt(version);
+        } catch (NumberFormatException ex) {
+            throw new BadRequestException("Invalid version " + version);
+        }
+        return participantVersionService.getParticipantVersion(appId, healthCode, versionInt);
```
alxdarksage, line 85: FWIW BridgeUtils.getIntOrDefault(...) does the same

alxdarksage (approved):

PR #473, opened 2021-11-16
BRIDGE-3131 Truncate UserAgent and ClientInfo to fit in our database

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-3131

MindKind (appId wellness) users in India are sending excessively long User-Agent strings. When we try to shove this in our RequestInfo, our database vomits because it's expecting a varchar(255). These User-Agent strings aren't in any format ClientInfo recognizes, so I'm okay with just truncating them.

While testing this, I also discovered that if you specify a User-Agent string that's just a long string of the same repeating character ("AAAAAAAAAAAAAAAAAAAAAAAA..."), ClientInfo will parse the whole thing as the app name. This is even less realistic than the above, but I figured while I'm in here, I'll have ClientInfo throw out any excessively long strings as well.

Note that the real fix is that MindKind app should send sensible User-Agent strings, but given that this is preventing users from signing in, and this is causing us to throw HTTP 500s, I figured we ought to fix it.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/BridgeConstants.java
+++ b/src/main/java/org/sagebionetworks/bridge/BridgeConstants.java
@@ -12,6 +12,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableList;
 
 public class BridgeConstants {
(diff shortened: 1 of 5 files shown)


alxdarksage (approved):
Looks good to merge.

PR #474, opened 2021-11-17
PasswordGeneratorBugFix

dephillipsmichael:
I am using the PasswordGenerator class in the project https://github.com/Sage-Bionetworks/DianUserMigration-Java and came across a minor bug in its implementation.

I noticed that the passwords that were generated from the code heavily favored the symbol being in the 6th or 7th character location. I thought that was odd, so I investigated deeper using the code in the screenshot below, and found that the distribution never put the symbol at index 0 and 1, which made me realize there was definitely a bug somewhere.

Screen Shot 2021-11-16 at 3 37 04 PM

The root cause of the bug ended up being that the `getFourUniqueIntegers` function stores the random ints in a `HashSet` which orders the values from lowest to highest. So, when the `nextPassword` function converts it to an iterator, the symbols always gets the highest value of the set, which explains why 0 and 1 are not possible.

I switched to using a `LinkedHashSet` which stores the Ints based on the order you added them instead of what the value of them in the set is. This fixed the bug.

I added a conditional check in the unit tests that the symbol must be placed in each index at least 1% of the time for the tests to pass. The old algorithm failed the test, but the new one passes every time.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/PasswordGenerator.java
+++ b/src/main/java/org/sagebionetworks/bridge/PasswordGenerator.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge;
 
 import java.security.SecureRandom;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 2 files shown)


dephillipsmichael:
@alxdarksage I wasn't able to edit reviewers for some reason, but tagging you here so you can take a look, no rush at all.

I also wasn't able to run the unit tests, is there a way to trigger the tests to run on a PR? If not, I'd suggest pulling my branch down and running tests yourself, I'll give you write access to my fork.

alxdarksage (approved):

PR #478, opened 2021-11-18
Fixing two bugs with study activity events:

alxdarksage:
- updates to study bursts were not carrying over the new study burst metadata;
- the query for most recent events was not using eventTimestamp for the key query portion;
- sorting the events by event ID after we add some stuff in the service;

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyActivityEventDao.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyActivityEventDao.java
@@ -31,8 +31,8 @@ public class HibernateStudyActivityEventDao implements StudyActivityEventDao {
     static final String GET_RECENT_SQL = "SELECT *, (SELECT count(*) as total FROM " +
             "StudyActivityEvents WHERE eventId = sae.eventId AND studyId = :studyId " +
             "AND userId = :userId GROUP BY eventId) FROM StudyActivityEvents AS sae " +
-            "WHERE userId = :userId AND studyId = :studyId AND createdOn = (SELECT " +
-            "createdOn FROM StudyActivityEvents WHERE userId = :userId AND studyId = " +
+            "WHERE userId = :userId AND studyId = :studyId AND eventTimestamp = (SELECT " +
+            "eventTimestamp FROM StudyActivityEvents WHERE userId = :userId AND studyId = " +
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyActivityEventDao.java
```diff
@@ -31,8 +31,8 @@
     static final String GET_RECENT_SQL = "SELECT *, (SELECT count(*) as total FROM " +
             "StudyActivityEvents WHERE eventId = sae.eventId AND studyId = :studyId " +
             "AND userId = :userId GROUP BY eventId) FROM StudyActivityEvents AS sae " +
-            "WHERE userId = :userId AND studyId = :studyId AND createdOn = (SELECT " +
-            "createdOn FROM StudyActivityEvents WHERE userId = :userId AND studyId = " +
+            "WHERE userId = :userId AND studyId = :studyId AND eventTimestamp = (SELECT " +
+            "eventTimestamp FROM StudyActivityEvents WHERE userId = :userId AND studyId = " +
             ":studyId AND eventId = sae.eventId ORDER BY createdOn DESC LIMIT 1) " +
             "ORDER BY eventId";
```
alxdarksage, line 37: Where's the test for this?

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
```diff
@@ -228,6 +234,7 @@ public void publishEvent(StudyActivityEvent event, boolean showError, boolean up
         for (String fieldName : GLOBAL_EVENTS_OF_INTEREST) {
             addIfPresent(events, map, fieldName, true);    
         }
+        events.sort(Comparator.comparing(StudyActivityEvent::getEventId));
         return new ResourceList<>(events); 
     }
```
alxdarksage, line 240: We need tests for the SQL query, also verify that a study burst event when updated copies over the study burst metadata to the event that is persisted.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudyActivityEventDao.java
alxdarksage: The test refers directly to this SQL constant so when this changes, the test always passes. Which makes sense.

PR #484, opened 2021-12-01
Adherence 2

alxdarksage:
- Add session name and symbol to TimelineMetadata for authoring and administrative tools;
- add method to get TimelineMetadata records for a schedule;
- add standalone enum and utility to describe adherence states.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
@@ -19,10 +19,11 @@ public class AdherenceUtils {
     public static SessionCompletionState calculateSessionState(
             AdherenceRecord record, int startDay, int endDay, Integer daysSinceEvent) {
 
-        // Not at all clear that this means it's not applicable.
+        // The origin event has not occurred for this user, so the session is not considered applicable.
         if (daysSinceEvent == null) {
             return NOT_APPLICABLE;
         }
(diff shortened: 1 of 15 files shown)


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
```diff
@@ -0,0 +1,59 @@
...
+        if (endDay < daysSinceEvent) {
+            if (record.getStartedOn() != null && record.getFinishedOn() != null) {
+                return COMPLETED;
+            }
+            if (record.getStartedOn() != null) {
+                return ABANDONED;
+            }
+            return EXPIRED;
+        }
+        // This record falls on the day and so it is available to the participant.
```
alxdarksage: Make a note that if it is on or after the window, then we're looking at the completion state (we are NOT hiding things that are in the future, if the client has told us they are started or finished, not because they should be available—they shouldn't be—but because we don't want to hide this fact from adherence monitors.

alxdarksage: Ask about edge cases around finishing sessions after they have expired. Should we prevent it? Allow it but document it? What is the consequences for adherence?

alxdarksage: Updated

jnelsonjava (approved):
Looks good!

PR #486, opened 2021-12-08
Adding validation of studies removing custom events currently in their schedule

jnelsonjava:
For Bridge-3135.

- Preventing removal of a custom event from a study if that event is being used by a schedule.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -24,8 +24,10 @@ import static org.sagebionetworks.bridge.models.studies.StudyPhase.IN_FLIGHT;
 import static org.sagebionetworks.bridge.models.studies.StudyPhase.RECRUITMENT;
 import static org.sagebionetworks.bridge.models.studies.StudyPhase.WITHDRAWN;
 
+import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import org.joda.time.DateTime;
 
@@ -158,7 +160,7 @@ public class StudyService {
         study.setAppId(appId);
         study.setPhase(DESIGN);
         
-        StudyValidator validator = new StudyValidator(scheduleService);
+        StudyValidator validator = new StudyValidator(getCustomEventIdsFromSchedule(appId, study.getScheduleGuid()));
         Validate.entityThrowingException(validator, study);
         
         study.setVersion(null);
@@ -206,7 +208,7 @@ public class StudyService {
         study.setModifiedOn(DateTime.now());
         study.setPhase(existing.getPhase());
     
-        StudyValidator validator = new StudyValidator(scheduleService);
+        StudyValidator validator = new StudyValidator(getCustomEventIdsFromSchedule(appId, study.getScheduleGuid()));
         Validate.entityThrowingException(validator, study);
         
         VersionHolder keys = studyDao.updateStudy(study);
@@ -352,5 +354,16 @@ public class StudyService {
         cacheProvider.removeObject(cacheKey);
         
         return study;
+    }
+    
+    private Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
+        Schedule2 schedule = scheduleService.getSchedule(appId, scheduleGuid);
+        Set<String> existingEventIds = new HashSet<>();
+        if (schedule != null) {
+            existingEventIds = schedule.getSessions().stream()
+                    .flatMap(session -> session.getStartEventIds().stream())
+                    .filter(s -> s.startsWith("custom")).collect(Collectors.toSet());
+        }
+        return existingEventIds;
     }
 }
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -157,7 +157,9 @@ public class StudyService {
         
         study.setAppId(appId);
         study.setPhase(DESIGN);
-        Validate.entityThrowingException(StudyValidator.INSTANCE, study);
+        
+        StudyValidator validator = new StudyValidator(scheduleService);
+        Validate.entityThrowingException(validator, study);
         
         study.setVersion(null);
         study.setDeleted(false);
@@ -203,8 +205,9 @@ public class StudyService {
         study.setCreatedOn(existing.getCreatedOn());
         study.setModifiedOn(DateTime.now());
         study.setPhase(existing.getPhase());
-
-        Validate.entityThrowingException(StudyValidator.INSTANCE, study);
+    
+        StudyValidator validator = new StudyValidator(scheduleService);
+        Validate.entityThrowingException(validator, study);
         
         VersionHolder keys = studyDao.updateStudy(study);
         
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyService.java
@@ -357,12 +357,14 @@ public class StudyService {
     }
     
     private Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
-        Schedule2 schedule = scheduleService.getSchedule(appId, scheduleGuid);
         Set<String> existingEventIds = new HashSet<>();
-        if (schedule != null) {
-            existingEventIds = schedule.getSessions().stream()
-                    .flatMap(session -> session.getStartEventIds().stream())
-                    .filter(s -> s.startsWith("custom")).collect(Collectors.toSet());
+        if (scheduleGuid != null) {
+            Schedule2 schedule = scheduleService.getSchedule(appId, scheduleGuid);
+            if (schedule != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -350,4 +355,15 @@ private Study phaseTransition(String appId, String studyId, StudyPhase targetPha
         
         return study;
     }
+    
+    private Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
+        Schedule2 schedule = scheduleService.getSchedule(appId, scheduleGuid);
```
alxdarksage: A study can not have a scheduleGuid, so you can't guarantee it won't be null here.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -350,4 +355,15 @@ private Study phaseTransition(String appId, String studyId, StudyPhase targetPha
...
+        Set<String> existingEventIds = new HashSet<>();
+        if (schedule != null) {
+            existingEventIds = schedule.getSessions().stream()
+                    .flatMap(session -> session.getStartEventIds().stream())
+                    .filter(s -> s.startsWith("custom")).collect(Collectors.toSet());
```
alxdarksage: I would add the colon: s.startsWith("colon:")

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -141,10 +146,15 @@ public void validate(Object object, Errors errors) {
...
+            scheduledCustomEventIds.remove(customEvent.getEventId());
             errors.popNestedPath();
         }
         if (uniqueIds.size() > 0 && (uniqueIds.size() != study.getCustomEvents().size())) {
-            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplidate event IDs");
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplicate event IDs");
+        }
+        if (!scheduledCustomEventIds.isEmpty()) {
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, 
+                    String.format("cannot remove custom events currently used in a schedule: %s", scheduledCustomEventIds.toString()));
```
alxdarksage: You need to verify that the existing scheduleCustomEventIds are all in the study...but there could be none and that's okay (schedules can be saved with no sessions). Adding is okay here. Just not removing.

alxdarksage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -350,4 +355,17 @@ private Study phaseTransition(String appId, String studyId, StudyPhase targetPha
         
         return study;
     }
+    
+    private Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
+        Set<String> existingEventIds = new HashSet<>();
+        if (scheduleGuid != null) {
+            Schedule2 schedule = scheduleService.getSchedule(appId, scheduleGuid);
+            if (schedule != null) {
```
jnelsonjava: This null check is unnecessary.

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -141,10 +146,16 @@ public void validate(Object object, Errors errors) {
...
             errors.popNestedPath();
         }
         if (uniqueIds.size() > 0 && (uniqueIds.size() != study.getCustomEvents().size())) {
-            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplidate event IDs");
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplicate event IDs");
+        }
+        // Any EventIds not removed from the set are on a schedule but missing from the validated study
+        if (!protectedCustomEventIds.isEmpty()) {
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, 
+                    String.format("cannot remove custom events currently used in a schedule: %s", protectedCustomEventIds.toString()));
```
jnelsonjava: Look up comma space joiner

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -350,4 +355,17 @@ private Study phaseTransition(String appId, String studyId, StudyPhase targetPha
         
         return study;
     }
+    
+    private Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
```
jnelsonjava: Make it protected to test in service tests.

src/test/java/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
```diff
@@ -285,7 +295,22 @@ public void customEvents_duplicated() {
         study.getCustomEvents().add(event);
         study.getCustomEvents().add(event);
         
-        assertValidatorMessage(INSTANCE, study, CUSTOM_EVENTS_FIELD, "cannot contain duplidate event IDs");
+        assertValidatorMessage(validator, study, CUSTOM_EVENTS_FIELD, "cannot contain duplicate event IDs");
+    }
+
```
jnelsonjava, line 300: Add test for empty cases just to make sure it's not going to break anything.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
jnelsonjava: Done

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
jnelsonjava: Removed

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
jnelsonjava: Swapped to joiner instead of toString

src/test/java/org/sagebionetworks/bridge/validators/StudyValidatorTest.java
jnelsonjava: Added tests for empty set and null

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
jnelsonjava: Added null check on scheduleGuid before calling getSchedule.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
jnelsonjava: Added colon to "custom:"

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
jnelsonjava: Changed to set comparison operations.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
```diff
@@ -350,4 +355,22 @@ private Study phaseTransition(String appId, String studyId, StudyPhase targetPha
...
         return study;
     }
+    
+    /**
+     * Retrieves the custom event IDs from a schedule. The returned event IDs will 
+     * have the "custom:" prefix removed. If scheduleGuid is null, will return
+     * an empty set.
+     */
+    protected Set<String> getCustomEventIdsFromSchedule(String appId, String scheduleGuid) {
+        Set<String> existingEventIds = new HashSet<>();
```
alxdarksage, line 365: very minor: could use ImmutableSet.of() here to avoid creating an object instance

src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -135,16 +141,21 @@ public void validate(Object object, Errors errors) {
...
             errors.popNestedPath();
         }
         if (uniqueIds.size() > 0 && (uniqueIds.size() != study.getCustomEvents().size())) {
-            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplidate event IDs");
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, "cannot contain duplicate event IDs");
+        }
+        if (protectedCustomEventIds != null && !uniqueIds.containsAll(protectedCustomEventIds)) {
+            protectedCustomEventIds.removeAll(uniqueIds);
+            errors.rejectValue(CUSTOM_EVENTS_FIELD, 
+                    String.format("cannot remove custom events currently used in a schedule: [%s]",
```
alxdarksage: Minor: you can remove the brackets in the string

jnelsonjava: Sounds good, done.

src/main/java/org/sagebionetworks/bridge/services/StudyService.java
jnelsonjava: I have to leave it as mutable because the validator removes items from the set.

PR #500, opened 2021-12-21
Move some of this config into the default file (pull password from environment).

alxdarksage:
Once this is deployed, we should be able to clean up the infrastructure scripts, which will remove the values from the environment.

### DIFF
--- a/src/main/resources/BridgeServer2.conf
+++ b/src/main/resources/BridgeServer2.conf
@@ -149,9 +149,12 @@ prod.usersigned.consents.bucket = bridgepf-prod-awss3usersignedconsentsdownloadb
 # Bootstrap user for integration tests (will only be used when first initializing your database).
 # This should match the synapse.test.user.* credentials set for your integration test properties 
 # in your bridge-sdk-test.properties file.
-admin.email = dummy-value
+admin.email = bridgetest
+admin.synapse.user.id = 3336429
 admin.password = dummy-value
-admin.synapse.user.id = dummy-value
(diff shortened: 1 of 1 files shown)


src/main/resources/BridgeServer2.conf
```diff
@@ -149,9 +149,12 @@ prod.usersigned.consents.bucket = bridgepf-prod-awss3usersignedconsentsdownloadb
 # Bootstrap user for integration tests (will only be used when first initializing your database).
 # This should match the synapse.test.user.* credentials set for your integration test properties 
 # in your bridge-sdk-test.properties file.
-admin.email = dummy-value
+admin.email = bridgetest
```
DwayneJengSage, line 152: I think this is an email address, which could vary by developer.

DwayneJengSage (changes requested):

src/main/resources/BridgeServer2.conf
alxdarksage: Yes, in my local file, local.admin.email is an email address. Are you saying change this to bridgetest@synapse.org (which I am guessing will also work)?

DwayneJengSage: I believe this is the email address of the Bridge account (eg dwayne.jeng@sagebase.org), so it would need to be overridden by each developer.

alxdarksage:
I am going to close this PR, Dwayne just updated this and it's correct as it stands.

PR #501, opened 2021-12-22
BRIDGE-2606 Export Participant Versions

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2606


src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
```diff
@@ -61,10 +69,82 @@
...
+        // instead of reserving potentially 600+ characters that will be sparsely used, we will keep to the
+        // 250 character limit.
+        ColumnModel dataGroupsColumn = new ColumnModel();
+        dataGroupsColumn.setName("dataGroups");
+        dataGroupsColumn.setColumnType(ColumnType.STRING);
+        dataGroupsColumn.setMaximumSize(250L);
+        listBuilder.add(dataGroupsColumn);
+
+        // These are 2-character language codes (eg "en", "es"). Let's generously allow participants to export up to
+        // 10 languages, which is 20 characters at most, and should be more than enough for nearly all participants.
```
alxdarksage: I don't think this changes your calculations, but at some point we'll probably go to the five character version that specifies region, e.g. en-US and en-UK.

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
```diff
@@ -61,10 +69,82 @@
...
+        sharingScopeColumn.setMaximumSize(25L);
+        listBuilder.add(sharingScopeColumn);
+
+        // Synapse doesn't have maps as a type in their Table model. For now, we serialize it as a string. In 2.0, we
+        // set a max string length of 250 characters, and that seems to be fine.
+        ColumnModel studyMembershipsColumn = new ColumnModel();
+        studyMembershipsColumn.setName("studyMemberships");
+        studyMembershipsColumn.setColumnType(ColumnType.STRING);
+        studyMembershipsColumn.setMaximumSize(250L);
+        listBuilder.add(studyMembershipsColumn);
```
alxdarksage, line 135: Terminology: study enrollments? Will multiple studies share the same table, and if so how will that work if studies eventually go to separate projects/tables? It also turns out that a coordinator for study A doesn't have permission to see that the participant is in study B (if they don't have access to study B). This is only partially implemented on Bridge Server (some identifiers are app-scoped so they might leak information about a participant's involvement in other studies).

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
```diff
@@ -61,10 +69,82 @@
...
+
+        // The longest time zone I could find was 28 characters long. Let's add a bit of a buffer and say 40
+        // characters. As long as Llanfair PG, Wales doesn't have its own time zone, this should be fine.
+        ColumnModel timeZoneColumn = new ColumnModel();
+        timeZoneColumn.setName("timeZone");
+        timeZoneColumn.setColumnType(ColumnType.STRING);
+        timeZoneColumn.setMaximumSize(40L);
+        listBuilder.add(timeZoneColumn);
```
alxdarksage, line 143: We call this client time zone, vs. study time zone which is the time zone in which the study is being run or in which they coordinators want to sees times.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
DwayneJengSage: > Terminology: study enrollments?

I called it study memberships, because that's what it's called in Exporter 2.0 and in [BridgeUtils](https://github.com/Sage-Bionetworks/BridgeServer2/blob/develop/src/main/java/org/sagebionetworks/bridge/BridgeUtils.java#L118). I can change the name if you think it's more clear that way, but this is the studyId to externalId mapping, which is different from StudyParticipants.getEnrollments.

> Will multiple studies share the same table, and if so how will that work if studies eventually go to separate projects/tables?

The app-scoped Synapse project will contain all the identifiers. The study-scoped Synapse project will only contain the current study and that study's external ID, if any.

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
DwayneJengSage: Renamed to clientTimeZone.

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
DwayneJengSage: Bumped this up to 50 characters.

PR #504, opened 2022-01-05
Refactor how adherence report generator gets the data to operate on.

alxdarksage:
The next report needs all the same information, so changed this a bit so we can collect all the state and pass it in to the appropriate report generator.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceState.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceState.java
@@ -1,0 +1,188 @@
+package org.sagebionetworks.bridge.models.schedules2.adherence;
+
+import static java.util.stream.Collectors.counting;
+import static java.util.stream.Collectors.toMap;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.Days;
+import org.joda.time.LocalDate;
+import org.sagebionetworks.bridge.models.activities.StudyActivityEvent;
+import org.sagebionetworks.bridge.models.schedules2.adherence.eventstream.EventStream;
+import org.sagebionetworks.bridge.models.schedules2.adherence.eventstream.EventStreamDay;
+import org.sagebionetworks.bridge.models.schedules2.timelines.TimelineMetadata;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+public final class AdherenceState {
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceState.java
```diff
@@ -0,0 +1,188 @@
...
+            eventStreamDay.setSessionName(meta.getSessionName());
+            eventStreamDay.setSessionSymbol(meta.getSessionSymbol());
+            eventStreamDay.setWeek(startDay / 7);
+            eventStreamDay.setStudyBurstId(meta.getStudyBurstId());
+            eventStreamDay.setStudyBurstNum(meta.getStudyBurstNum());
+            streamsByStreamKey.put(streamKey, eventStreamDay);
+            getEventStreamById(eventId).addEntry(startDay, eventStreamDay);
+        }
+        return eventStreamDay;
+    }
```
alxdarksage, line 111: Mixes two things: an immutable parameter object, and an object for collecting the calculated state of streams. This is confusing and the stream calculation should go back into the generator (which would not be an instance in that case, as before).

alxdarksage: After discussing I decided to leave this. From the perspective of the generators, these do not appear mutable, it just appears that the state has a stream, for every ID. We could create them all up front, but there's just no reason.

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceStateTest.java
```diff
@@ -0,0 +1,224 @@
...
+        } catch(IllegalArgumentException e) {
+        }
+        try {
+            emptyState.getDaysSinceEventById("event1");    
+        } catch(IllegalArgumentException e) {
+        }
+        try {
+            emptyState.getEventTimestampById("event1");    
+        } catch(IllegalArgumentException e) {
+        }
```
jnelsonjava: I think it would be worth it to add fail(..) statements in the try blocks. Just to prove that the getters aren't hiding the exceptions and returning nulls.

jnelsonjava (approved):
Looks good. One comment in the tests to check out.

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceStateTest.java
alxdarksage: Good catch. This test is wrong (I removed the illegal argument exception checks) and fails caught that. Changed the test to reflect the fact that these methods now return null in most cases.

PR #508, opened 2022-01-07
Add weekly adherence report APIs

alxdarksage:


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamDay.java
```diff
@@ -83,4 +84,10 @@ public void setTimeWindows(List<EventStreamWindow> timeWindows) {
     public void addTimeWindow(EventStreamWindow timeWindowEntry) {
         this.timeWindows.put(timeWindowEntry.getTimeWindowGuid(), timeWindowEntry);
     }
+    public String getLabel() {
+        return label;
+    }
+    public void setLabel(String label) {
+        this.label = label;
+    }
```
alxdarksage, line 92: Do we need this label? Double-check.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/weekly/NextActivity.java
```diff
@@ -0,0 +1,55 @@
...
+    public String getStudyBurstId() {
+        return studyBurstId;
+    }
+    public Integer getStudyBurstNum() {
+        return studyBurstNum;
+    }
+    public LocalDate getStartDate() {
+        return startDate;
+    }
+}
```
alxdarksage, line 60: Don't we need the label here too?

alxdarksage:
Labels are also a cue to the UI as to how many rows there need to be in this column-major report. So don't hide them, show them.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamDay.java
alxdarksage: Yes, it's obviously important in the JSON output....

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/weekly/NextActivity.java
alxdarksage: Added

alxdarksage:
Also adding all labels for a user Understanding the "rowLabels" property.

jnelsonjava (approved):
Looks good.

PR #513, opened 2022-01-25
Fix leap year min age consent bug & Add Test

chuguoxipo:
Fix leap year min age consent bug & Add Test

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
+++ b/src/main/java/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
@@ -20,6 +20,9 @@ public class ConsentSignatureValidator implements Validator {
     private static final String TOO_YOUNG = "too recent (the study requires participants to be %s years of age or older).";
     private static final String CANNOT_BE_BLANK = "cannot be missing, null, or blank";
     private static final String CANNOT_BE_EMPTY_STRING = "cannot be an empty string";
+
+    private static final int LEAP_YEAR_MONTH = 2;
+    private static final int LEAP_YEAR_DAY = 29;
     
     private final int minAgeOfConsent;
 
@@ -54,6 +57,15 @@ public class ConsentSignatureValidator implements Validator {
                 Period period = new Period(birthdate, now);
 
                 if (period.getYears() < minAgeOfConsent) {
+                    String message = String.format(TOO_YOUNG, minAgeOfConsent);
+                    errors.rejectValue("birthdate", message);
+                }
+
+                // Handle participants' birthdays on 2/29 of a leap year.
+                if (period.getYears() == minAgeOfConsent && now.getMonthOfYear() == LEAP_YEAR_MONTH &&
+                        birthdate.getMonthOfYear() == LEAP_YEAR_MONTH &&
+                        birthdate.getDayOfMonth() == LEAP_YEAR_DAY)
+                {
                     String message = String.format(TOO_YOUNG, minAgeOfConsent);
                     errors.rejectValue("birthdate", message);
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
```diff
@@ -57,6 +60,15 @@ public void validate(Object target, Errors errors) {
...
                 }
+
+                // Handle participants' birthdays on 2/29 of a leap year.
+                if (period.getYears() == minAgeOfConsent && now.getMonthOfYear() == LEAP_YEAR_MONTH &&
+                        birthdate.getMonthOfYear() == LEAP_YEAR_MONTH &&
+                        birthdate.getDayOfMonth() == LEAP_YEAR_DAY)
+                {
+                    String message = String.format(TOO_YOUNG, minAgeOfConsent);
+                    errors.rejectValue("birthdate", message);
+                }
```
alxdarksage, line 62: When we calculate dates like this it is easy to get it wrong, and the leaper issue might only be one special case where we're getting date calculations wrong. It is safer to use a date and time library (we use Joda). Something like this:

        int minAgeOfConsent = 18;
        LocalDate birthday = LocalDate.parse("2004-01-24"); // the value entered
        LocalDate today = LocalDate.now(); // not exact, it'll be today in UTC and the user probably isn't in UTC
        boolean oldEnough = birthday.plusYears(minAgeOfConsent).isBefore(today);
        assertTrue(oldEnough);

You could also test if they are equal (in addition to before) as when we say you are 18, we aren't specifying what time zone you're in. So it may be enough that it's your birthday today somewhere in the world.

chuguoxipo: Hi Alx, this is to handle the bug listed here: https://sagebionetworks.jira.com/browse/BRIDGE-2136

```
int minAgeOfConsent = 18;
LocalDate birthday = LocalDate.parse("2004-02-29"); // the value entered
LocalDate now = LocalDate.parse("2022-02-28"); // not exact, it'll be today in UTC and the user probably isn't in UTC
boolean oldEnough = birthday.plusYears(minAgeOfConsent).isEqual(now) || birthday.plusYears(minAgeOfConsent).isBefore(now);
assertTrue(oldEnough);
```

The above test passes. However, in reality, participants with birthdate on 2/29/2004 are not turning 18 years old until 3/1/2022.

DwayneJengSage: > When we calculate dates like this it is easy to get it wrong, and the leaper issue might only be one special case where we're getting date calculations wrong. It is safer to use a date and time library (we use Joda).

Hi Alx,

The issue here is specifically that Joda thinks there are 18 years between 2004-02-29 and 2022-02-28. If you use Period.getYears(), it will return 18 years for this result. Additionally, if you do LocalDate.parse("2004-02-29").plusYears(18), you will get 2022-02-28, so that solution won't work either.

The only option to fix this bug, that I'm aware of, is to special case it.

DwayneJengSage (approved):
Changes look good. Will go over next steps and next project with you after stand-up on Wednesday.

src/main/java/org/sagebionetworks/bridge/validators/ConsentSignatureValidator.java
alxdarksage: Then let's note it in the code because this looks pretty hacky. I can see someone else trying to remove it. I see two issues filed with Joda about this calculation that are closed as "won't fix" for backwards compatibility reasons:

https://github.com/JodaOrg/joda-time/issues/295
https://github.com/JodaOrg/joda-time/issues/132

chuguoxipo:
> Changes look good. Will go over next steps and next project with you after stand-up on Wednesday.

I've extracted the code block to handle the leap year bug to a separate helper method so that the code/logic is cleaner and easier to read. Should I do another pull request since this one is approved?

DwayneJengSage (approved):

PR #517, opened 2022-02-02
Schedule metadata

alxdarksage:


src/main/java/org/sagebionetworks/bridge/models/schedules2/timelines/TimelineMetadata.java
```diff
@@ -50,7 +50,6 @@ public static final TimelineMetadata copy(TimelineMetadata meta) {
     
     public final Map<String,String> asMap() {
         Map<String,String> map = new HashMap<>();
-        map.put("guid", guid);
         map.put("assessmentInstanceGuid", assessmentInstanceGuid);
```
alxdarksage, line 53: Where is session instance guid?

jnelsonjava (approved):
Looks good!

PR #518, opened 2022-02-04
Adherence fixes

alxdarksage:
- searching for labels and anything else returned no results (a saga of rewriting HQL ensued, but it works now);
- removed colons from label search (these are visible in searchableLabels and if the caller wants to match a label exactly, they can supply the colons in their search string);
- labelFilter renamed to labelFilters, as it was turned into an array late in the process;
- adding "no schedule" progress state for cases where someone calls the weekly adherence report API and generates a report for a participant in a study that has—no schedule. Otherwise we report it as "done" which seems weird.


jnelsonjava (approved):
Looks good.

PR #519, opened 2022-02-04
Refactor validation code for better encapsulation

alxdarksage:
I realized with recent work I was making validator stuff even messier by putting constants into BridgeConstants, and was motivated to do a cleanup.

- **Validate** now only exposes two methods, entityThrowingException and nonEntityThrowingException. Rewrote some older tests so they no longer depend on Validate's internal methods;
- **ValidatorUtils** is now package private and all its utilities are available to validators only. One or two methods were moved to BridgeUtils that were less about validation and more about checking state;
- Moved a lot of constants that had crept into BridgeConstants back to their validators or, if they were used across validators, to Validate which already had some similar constants;
- Removed the presence of "%s" from the errors as the validator utilities don't require it, and removed a lot of places where we were referencing these constants for checkNull() - we don't need the message in these precondition assertions, they are there for developers and it's always clear from the stack trace what is wrong.
- AppAndUsers validation was a bit of a mess so I cleaned it up (this was required to hide Validate's implementation details, and a prime example of why this refactor should help down the road);

None of this changes any API; all unit tests pass and integration tests pass without modification.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
@@ -19,15 +19,14 @@ import java.util.Set;
 
 import org.springframework.validation.Errors;
 import org.testng.annotations.Test;
-
+import org.mockito.Mockito;
 import org.sagebionetworks.bridge.models.appconfig.AppConfig;
-import org.sagebionetworks.bridge.validators.Validate;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 
-public class CriteriaUtilsTest {
+public class CriteriaUtilsTest extends Mockito {
     
     private static final String KEY = "key";
     private static final Set<String> EMPTY_SET = ImmutableSet.of();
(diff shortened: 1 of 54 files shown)


src/test/java/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
```diff
@@ -266,74 +268,71 @@ public void validateDataGroupSetsCannotBeNull() {
...
                 .noneOfGroups(ImmutableSet.of("group2", "group3")).build();
         
-        Errors errors = Validate.getErrorsFor(criteria);
+        Errors errors = mock(Errors.class);
         CriteriaUtils.validate(criteria, ImmutableSet.of("group1","group2","group3","group4"), EMPTY_SET, errors);
         // It's a set so validate without describing the order of the groups in the error message
-        assertTrue(errors.getFieldErrors("allOfGroups").get(0).getCode().contains("includes these excluded data groups: "));
-        assertTrue(errors.getFieldErrors("allOfGroups").get(0).getCode().contains("group2"));
-        assertTrue(errors.getFieldErrors("allOfGroups").get(0).getCode().contains("group3"));
+        verify(errors).rejectValue("allOfGroups", "includes these excluded data groups: group2, group3");
```
jnelsonjava, line 298: If this comes from a set, can you guarantee that it will order "group2" before "group3" in the error message?

jnelsonjava (dismissed):
This looks great. But check out the commented test. I think the test could fail depending on the order items are visited in the set.

src/test/java/org/sagebionetworks/bridge/models/CriteriaUtilsTest.java
alxdarksage: ImmutableSet (from Google's Guava library) is unusual for a set because the implementation guarantees that elements will be returned in the same order they are specified in the factory method.

jnelsonjava (approved):
Good to go!

PR #520, opened 2022-02-04
Add Developer APIs & Tests for Health Records in Exporter 3.0

chuguoxipo:


src/test/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3ControllerTest.java
```diff
@@ -5,8 +5,7 @@
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-import static org.sagebionetworks.bridge.TestConstants.HEALTH_CODE;
-import static org.sagebionetworks.bridge.TestConstants.TEST_USER_ID;
+import static org.sagebionetworks.bridge.TestConstants.*;
```
DwayneJengSage: We generally try to avoid using wildcards in imports, to make sure we don't accidentally import something we don't intend to.

There's a setting in IntelliJ. Preferences -> Editor -> Code Style -> Java -> Imports -> Classes count to use import with '*' and Names count to use static import with '*'. I have those set to 999.

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
```diff
@@ -4,6 +4,7 @@
 import static org.sagebionetworks.bridge.Roles.WORKER;
 
 import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.models.accounts.UserSession;
```
DwayneJengSage: In our IntelliJ settings, we have Bridge imports separate from the rest, so we can organize our imports from other imports.

You can find this in IntelliJ through Preferences -> Editor -> Code Style -> Java -> Imports

At the bottom, there's Import Layout, add an extra line at the end for import org.sagebionetworks.bridge.*

DwayneJengSage (changes requested):
Changes look good. One comment about imports.

I can send you the IntelliJ code style XML if that helps.

DwayneJengSage (approved):

PR #523, opened 2022-02-08
Completion flag event stream report

alxdarksage:
Now both reports have the participant study progress constant calculated for them.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGeneratorTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGeneratorTest.java
@@ -382,10 +382,12 @@ public class EventStreamAdherenceReportGeneratorTest {
     
     @Test
     public void progression_done() { 
-        AdherenceState state = new AdherenceState.Builder().withNow(NOW).build();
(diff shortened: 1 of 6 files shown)


src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGeneratorTest.java
```diff
@@ -371,6 +372,41 @@ public void handleNulls() {
...
+        assertEquals(report.getProgression(), ParticipantStudyProgress.NO_SCHEDULE);
+    }
+    
+    @Test
+    public void progression_done() { 
+        AdherenceState state = new AdherenceState.Builder().withNow(NOW).build();
+        
+        EventStreamAdherenceReport report = INSTANCE.generate(state);
+        assertEquals(report.getProgression(), ParticipantStudyProgress.NO_SCHEDULE);
+    }
```
jnelsonjava: This is testing "no schedule" instead of "done" like the name indicates.

jnelsonjava (approved):
Looks good but there's one test to update before merging.

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGeneratorTest.java
alxdarksage: Thanks for catching this. I updated the test to verify the done state (we already have a test to verify the "no schedule" state).

PR #524, opened 2022-02-12
Prevent link to a schedule that is invalid or owned by another organization

alxdarksage:
Some strange errors can occur when you include a schedule GUID in a study that is linked to an organization the caller is not a member of. Prevent this.


jnelsonjava (approved):
Looks good to me.

PR #525, opened 2022-02-17
New schedule API for mobile apps

alxdarksage:


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/participantschedule/ParticipantScheduleGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+                SessionCompletionState asmtState = AdherenceUtils.calculateSessionState(
+                        asmtRecord, schSession.getStartDay(), schSession.getEndDay(), days);
+                
+                DateTime finishedOn = null;
+                String clientTimeZone = null;
+                if (asmtRecord != null) {
+                    clientTimeZone = asmtRecord.getClientTimeZone();
+                    if (clientTimeZone == null) {
+                        clientTimeZone = state.getClientTimeZone();
+                    }
```
alxdarksage: Client time zone is defined at the top of the document. Only use another time zone if it was recorded in the assessment record, and we want to preserve that information (don't fall back to other time zones...just use Z if it doesn't exist).

alxdarksage: This now only uses client time zone to adjust the time if it was set on the adherence record in the first place.

jnelsonjava (approved):
Looks good.

PR #528, opened 2022-02-22
Adding studyStartEventId to study (BRIDGE-3218)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/validators/StudyValidator.java
```diff
@@ -105,6 +109,14 @@ public void validate(Object object, Errors errors) {
         if (isBlank(study.getName())) {
             errors.rejectValue(NAME_FIELD, CANNOT_BE_BLANK);
         }
+        if (study.getStudyStartEventId() != null) {
```
alxdarksage, line 112: So, this will always come back timeline_retrieved, which will always pass...

jnelsonjava (approved):
Looks good.

PR #530, opened 2022-02-24
Revising participant schedule (BRIDGE-3221)

alxdarksage:
In the e-tag caching work I realized a couple of things:

1) A full schedule with adherence information is 100% uncacheable because the state depends on the time of the request;
2) It still wasn't accurate to the moment because our adherence information currently only calculates to the day.

Consequently I suspect the mobile clients will have to pull adherence records, which they are creating while caching the adherence information locally. So I removed all this information and added cache setting/invalidating code to cache the ParticipantSchedule for the long-term. The ParticipantSchedule is a Timeline + concrete date information for a specific participant given their time zone and events.

This call no longer allows the mobile client to submit a clientTimeZone value...it might be possible to add this to the e-tag in order to bust the cache however, so I will discuss with the mobile team when they get back (it would need to be mandatory to the request).


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -31,9 +31,14 @@
 
 import com.google.common.collect.Sets;
 
+import io.jsonwebtoken.lang.Objects;
```
DwayneJengSage: Minor: I think you want java.util.Objects.

DwayneJengSage (approved):
Looks good. One minor comment about imports.

PR #533, opened 2022-03-02
BRIDGE-3228 Permissions for authorization restructure

jnelsonjava:
[https://sagebionetworks.jira.com/browse/BRIDGE-3228](https://sagebionetworks.jira.com/browse/BRIDGE-3228)

- Adding model and API to manage permissions per entity


src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
```diff
@@ -0,0 +1,96 @@
...
+@Entity
+@Table(name = "Permissions")
+public class Permission implements BridgeEntity {
+    
+    @Id
+    private String guid;
+    @JsonIgnore
+    private String appId;
+    private String userId;
+    private String role; // TODO: rename?
```
alxdarksage: Should be an enum, of type Role or PermRole or PermissionRole, etc.

src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
```diff
@@ -0,0 +1,96 @@
...
+    public String toString() {
+        return "Permission{" +
+                "guid='" + guid + '\'' +
+                ", appId='" + appId + '\'' +
+                ", userId='" + userId + '\'' +
+                ", role='" + role + '\'' +
+                ", permissionType=" + permissionType +
+                ", objectId='" + objectId + '\'' +
+                '}';
+    }
```
alxdarksage: By convention, object methods (hashCode, toString(), equals) come last.

src/main/resources/db/changelog/changelog.sql
```diff
@@ -923,3 +923,33 @@ ADD COLUMN `assessmentGuid` varchar(60),
...
 ADD CONSTRAINT FOREIGN KEY (`assessmentGuid`) REFERENCES `Assessments` (`guid`) ON DELETE CASCADE,
 ADD CONSTRAINT FOREIGN KEY (`sessionGuid`) REFERENCES `Sessions` (`guid`) ON DELETE CASCADE;
+
+-- changeset bridge:63
+
+CREATE TABLE IF NOT EXISTS `Permissions` (
+  `guid` varchar(60) NOT NULL,
+  `appId` varchar(255) NOT NULL,
+  `userId` varchar(255) NOT NULL,
+  `role` varchar(255) NOT NULL, -- Should this be an enum?
```
jnelsonjava: Yes, should be enum. Can rename to accessLevel.

src/main/resources/db/changelog/changelog.sql
```diff
@@ -923,3 +923,33 @@ ADD COLUMN `assessmentGuid` varchar(60),
...
+  `permissionType` enum('STUDY') NOT NULL,
+  `objectId` varchar(255) NOT NULL,
```
jnelsonjava: Rename to entityType/Id

src/main/java/org/sagebionetworks/bridge/dao/PermissionDao.java
```diff
@@ -0,0 +1,20 @@
...
+import org.sagebionetworks.bridge.models.permissions.Permission;
+import org.sagebionetworks.bridge.models.permissions.PermissionType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public interface PermissionDao {
+    // TODO: add method descriptions
+    Set<Permission> getPermissions(String appId, String userId, Map<PermissionType, List<String>> objectIdFilters);
```
jnelsonjava: Can just separate this to getPermissionsForUser and ForObject.

src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
jnelsonjava: Change name and switch to enum.

src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
```diff
@@ -0,0 +1,96 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Permission that = (Permission) o;
+        return Objects.equals(guid, that.guid) && Objects.equals(appId, that.appId) && Objects.equals(userId, that.userId) && Objects.equals(role, that.role) && permissionType == that.permissionType && Objects.equals(objectId, that.objectId);
+    }
+    
+    @Override
+    public int hashCode() {
+        return Objects.hash(guid, appId, userId, role, permissionType, objectId);
+    }
```
jnelsonjava: Bump these and toString to end of class.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,87 @@
...
+    
+    @GetMapping("v1/permissions/{permissionType}/{objectId}")
+    public Set<Permission> getPermissionsForObject(
+            @PathVariable String permissionType, @PathVariable String objectId) {
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        String appId = session.getAppId();
+        
+        return permissionService.getPermissionsForObject(appId, permissionType, objectId);
+    }
```
jnelsonjava: Need to make a PermissionDetail object so that an account ref can be included in these responses.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,87 @@
...
+    
+    @PostMapping("v1/permissions")
+    public StatusMessage createPermission() {
```
jnelsonjava: Return PermissionDetail so guid is included.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,87 @@
...
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        String appId = session.getAppId();
+    
+        Permission permission = parseJson(Permission.class);
+        permission.setAppId(appId);
+        
+        permissionService.createPermission(appId, permission);
+        
+        return new StatusMessage("Permission created.");
```
jnelsonjava: Make sure to set to 201 return type for created.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    @PostMapping("v1/permissions/{guid}")
+    public StatusMessage updatePermission(@PathVariable String guid) {
```
jnelsonjava: Can return PermissionDetail here as well.

src/test/java/org/sagebionetworks/bridge/models/permissions/PermissionTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY_ID;
+import static org.sagebionetworks.bridge.TestConstants.TEST_USER_ID;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+public class PermissionTest {
+    
+    @Test
+    public void equalsHashCode() {
+        EqualsVerifier.forClass(Permission.class).allFieldsShouldBeUsed().verify();
```
jnelsonjava, line 21: Will need to add an exception for non-final fields.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public Set<Permission> getPermissionsForUser(String appId, String userId) {
+        checkArgument(isNotBlank(appId));
+        checkArgument(isNotBlank(userId));
+        
+        QueryBuilder builder = new QueryBuilder();
+        builder.append(GET_BY_USER, APP_ID, appId, USER_ID, userId);
+        
+        return ImmutableSet.copyOf(hibernateHelper.queryGet(
+                builder.getQuery(), builder.getParameters(), null, null, Permission.class));
```
DwayneJengSage: Permissions are mutable. If this doesn't need to be a Set, I recommend using a List instead.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,107 @@
...
+        return getPermissionDetail(appId, createdPermission);
+    }
+    
+    public PermissionDetail updatePermission(String appId, Permission permission) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(permission);
+    
+        Validate.entityThrowingException(INSTANCE, permission);
+        
+        Permission updatedPermission = permissionDao.updatePermission(appId, permission);
```
DwayneJengSage, line 123: Get the permission by Guid. If it doesn't exist, throw EntityNotFound.

Overwrite existing.accessLevel with permission.accessLevel to prevent them from changing the user ID or entity. (Alternatively, check if these are changed and throw a BadRequest if they are.)

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
```diff
@@ -0,0 +1,62 @@
...
+public final class PermissionDetail {
+    
+    private final String guid;
+    private final String userId;
+    private final AccessLevel accessLevel;
+    private final EntityType entityType;
+    private final String entityId;
+    private final AccountRef userAccountRef;
+    
+    public PermissionDetail(Permission permission, AccountRef userAccountRef) {
```
DwayneJengSage: Consider adding entityName, so that users can figure out what the permissions are for without looking up a guid.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,107 @@
...
+    
+    public PermissionDetail createPermission(String appId, Permission permission) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(permission);
+        
+        permission.setGuid(generateGuid());
+        
+        Validate.entityThrowingException(INSTANCE, permission);
+        
+        Permission createdPermission = permissionDao.createPermission(appId, permission);
```
DwayneJengSage, line 101: If the user or entity doesn't exist, what error does it throw? Status code? Is the error message reasonable?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: Switched over to a List.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Updated to throw EntityNotFound if the guid does not exist and overwrite all fields except for accessLevel with previous values.

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
jnelsonjava: Added an EntityRef to PermissionDetail that includes type, id, and name in case further entity details are needed later.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Added exception handling through the MySQLHibernatePersistenceExceptionConverter to handle foreign keys to non-existent entities. It throws a 409 ConstraintViolationException now. The message reads: "This permission cannot be created or updated because the referenced user account does not exist." (user account is replaced with study, assessment, or organization if the foreign key issue was based on the entityId)

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,101 @@
...
+    public Optional<Permission> getPermission(String appId, String guid) {
+        checkArgument(isNotBlank(appId));
+        checkArgument(isNotBlank(guid));
+        
+        Permission permission = hibernateHelper.getById(Permission.class, guid);
+        if (permission == null) {
+            return Optional.empty();
+        }
+        
+        return Optional.of(permission);
```
alxdarksage: Optional.ofNullable(permission) can do the null vs. value check for you in one line

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public List<Permission> getPermissionsForUser(String appId, String userId) {
+        checkArgument(isNotBlank(appId));
+        checkArgument(isNotBlank(userId));
+        
+        QueryBuilder builder = new QueryBuilder();
+        builder.append(GET_BY_USER, APP_ID, appId, USER_ID, userId);
+        
+        return hibernateHelper.queryGet(
+                builder.getQuery(), builder.getParameters(), null, null, Permission.class);
```
alxdarksage: This is simple enough you might consider dropping QueryBuilder and just using the string constant and an ImmutableMap.of(...)

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    
+    @Override
+    public List<Permission> getPermissionsForEntity(String appId, EntityType entityType, String entityId) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(entityType);
+        checkArgument(isNotBlank(entityId));
```
alxdarksage: Dwayne convinced me that a null check for these values is probably sufficient...if you're checking to see if the caller snuck an empty string into this field, that should be caught by validation.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,101 @@
...
+        return hibernateHelper.queryGet(
+                builder.getQuery(), builder.getParameters(), null, null, Permission.class);
+    }
+    
+    @Override
+    public Permission createPermission(String appId, Permission permission) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(permission);
+    
+        hibernateHelper.create(permission);
```
alxdarksage, line 68: What happens if the permission already exists with the same guid? Does the database throw an error that you translate correctly into an EntityAlreadyExists exception?

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
```diff
@@ -0,0 +1,101 @@
...
+        
+        return permission;
+    }
+    
+    @Override
+    public Permission updatePermission(String appId, Permission permission) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(permission);
+    
+        hibernateHelper.update(permission);
```
alxdarksage, line 78: What happens if the permission does not exist under that guid? We want an EntityNotFoundException (I'll look to see if this is in the service, but better is if we handle a database exception in the exception handler for this hibernateHelper and return ENFE).

src/main/java/org/sagebionetworks/bridge/hibernate/MySQLHibernatePersistenceExceptionConverter.java
```diff
@@ -82,15 +95,19 @@ public RuntimeException convert(PersistenceException exception, Object entity) {
...
-                displayMessage = selectMsg(
-                        rawMessage, FOREIGN_KEY_CONSTRAINTS, FK_CONSTRAINT_MSG, name, displayMessage);
+                if (rawMessage.contains("Cannot add or update a child row")) {
+                    displayMessage = selectMsg(
+                            rawMessage, FOREIGN_KEY_MISSING_PARENT_CONSTRAINTS, MISSING_PARENT_INVALID_FK_CONSTRAINT_MSG, name, displayMessage);
+                } else {
+                    displayMessage = selectMsg(
+                            rawMessage, FOREIGN_KEY_CONSTRAINTS, FK_CONSTRAINT_MSG, name, displayMessage);
+                }
             }
```
alxdarksage: I think instead of nesting these rawMessage string checks, they should be in a flat if/else but ordered correctly (in this case, your message check needs to come before "rawMessage.contains("a foreign key constraint fails")"

src/main/java/org/sagebionetworks/bridge/models/permissions/AccessLevel.java
```diff
@@ -0,0 +1,11 @@
+package org.sagebionetworks.bridge.models.permissions;
+
+public enum AccessLevel {
+    
+    LIST,
+    READ,
+    EDIT,
+    DELETE,
+    ADMIN
```
alxdarksage, line 34: We should document, as in much detail as we can, what each of these permissions means, and we'll need to copy it all over to the public documentation website.

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
```diff
@@ -0,0 +1,76 @@
...
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import org.sagebionetworks.bridge.models.accounts.AccountRef;
+
+import java.util.Objects;
+
+@JsonPropertyOrder({ "guid", "userId", "accessLevel", "entityType", "entityId", "entityRef", "userAccountRef" })
+public final class PermissionDetail {
+    
+    private final String guid;
+    private final String userId;
```
alxdarksage: It seems like userId should be covered by AccountRef, which I would call "account" to be consistent with EnrollmentDetail where we use the property "participant".

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
```diff
@@ -0,0 +1,76 @@
...
+    private final AccessLevel accessLevel;
+    private final EntityType entityType;
+    private final String entityId;
+    private final EntityRef entityRef;
```
alxdarksage: Same here... entityType and entityId are part of entityRef which should be exposed under the property "entity"

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
+package org.sagebionetworks.bridge.services;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.sagebionetworks.bridge.models.permissions.EntityType.*;
```
alxdarksage: Expand imports... we don't use .* as a rule

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
...
+        this.assessmentService = assessmentService;
+    }
+    
+    // For mock testing
+    protected String generateGuid() {
+        return BridgeUtils.generateGuid();
+    }
+    
+    public PermissionDetail createPermission(String appId, Permission permission) {
+        checkArgument(isNotBlank(appId));
```
alxdarksage: A null check is probably sufficient here. We're just looking for an internal error by a programmer

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
alxdarksage: I don't see a @Version field for this entity which we normally provide. It will give you optimistic locking behavior: if someone updates a permission and then you update it, you'll get an optimistic lock exception and you can examine the new object state before deciding to proceed with your edit. In any event, version would also need to be copied over.

alxdarksage: We also have started adding createdOn/modifiedOn timestamps to database objects. It's not a great form of auditing but it's better than nothing and would be consistent.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        
+        return permissionDetails;
+    }
+    
+    public void deletePermission(String appId, String guid) {
+        checkArgument(isNotBlank(appId));
+        checkArgument(isNotBlank(guid));
+        
+        permissionDao.deletePermission(appId, guid);
```
alxdarksage, line 169: What happens if the GUID doesn't exist? (Should throw ENFE)

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
...
+    
+        return permissionDetails;
+    }
+    
+    public List<PermissionDetail> getPermissionsForEntity(String appId, String entityType, String entityId) {
+        checkArgument(isNotBlank(appId));
+        checkNotNull(entityType);
+        checkArgument(isNotBlank(entityId));
+        
+        List<Permission> permissions = permissionDao.getPermissionsForEntity(appId, EntityType.valueOf(entityType), entityId);
```
alxdarksage, line 153: What happens if the entity does not exist? (Should throw ENFE)

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
...
+        Permission updatedPermission = permissionDao.updatePermission(appId, permission);
+    
+        return getPermissionDetail(appId, updatedPermission);
+    }
+    
+    public List<PermissionDetail> getPermissionsForUser(String appId, String userId) {
+        checkArgument(isNotBlank(appId));
+        checkArgument(isNotBlank(userId));
+    
+        List<Permission> permissions = permissionDao.getPermissionsForUser(appId, userId);
```
alxdarksage, line 135: What happens if the user does not exist? (Should throw ENFE)

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -0,0 +1,169 @@
...
+            Study study = studyService.getStudy(appId, permission.getEntityId(), false);
+            entityName = study.getName();
+        } else if (assessmentTypes.contains(permission.getEntityType())) {
+            Assessment assessment = assessmentService.getAssessmentByGuid(appId, null, permission.getEntityId());
+            entityName = assessment.getTitle();
+        }
+    
+        EntityRef entityRef = new EntityRef(permission.getEntityType(), permission.getEntityId(), entityName);
+        
+        return new PermissionDetail(permission, entityRef, userAccountRef);
```
alxdarksage: See comments about not duplicating some information about the permission alongside the entity and account references.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,89 @@
...
+        this.permissionService = permissionService;
+    }
+    
+    @GetMapping("v1/permissions/{userId}")
+    public List<PermissionDetail> getPermissionsForUser(@PathVariable String userId) {
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        String appId = session.getAppId();
+        
+        return permissionService.getPermissionsForUser(appId, userId);
```
alxdarksage: Return a resource list.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    
+    @GetMapping("v1/permissions/{entityType}/{entityId}")
+    public List<PermissionDetail> getPermissionsForEntity(
+            @PathVariable String entityType, @PathVariable String entityId) {
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        String appId = session.getAppId();
+        
+        return permissionService.getPermissionsForEntity(appId, entityType, entityId);
```
alxdarksage: Return a resource list.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    @PostMapping("v1/permissions")
+    @ResponseStatus(HttpStatus.CREATED)
+    public PermissionDetail createPermission() {
+        UserSession session = getAuthenticatedSession(ADMIN);
+        
+        String appId = session.getAppId();
+    
+        Permission permission = parseJson(Permission.class);
+        permission.setGuid(null);
```
alxdarksage, line 59: This gets reset by the service, but is harmless to null out here.

src/main/java/org/sagebionetworks/bridge/validators/PermissionValidator.java
```diff
@@ -0,0 +1,50 @@
...
+        return Permission.class.isAssignableFrom(clazz);
+    }
+    
+    @Override
+    public void validate(Object target, Errors errors) {
+        Permission permission = (Permission) target;
+        
+        if (isBlank(permission.getGuid())) {
+            errors.rejectValue(GUID_FIELD, CANNOT_BE_NULL_OR_EMPTY);
+        }
```
alxdarksage, line 33: Why doesn't validation fail in the service's create method, since you are nulling out GUID in the controller, and new objects won't actually have a guid?

alxdarksage (changes requested):
Looking very good. A few comments to consider (nothing must-do except that controllers should return ResourceLists for non-paginated lists, to be consistent across the whole API).

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: I didn't look at the guid duplication before because it's generated in the service. But it returns a ConstraintViolationException that says "Cannot update this permission because it has duplicate primary keys". If the guid is unique but there's already a permission with matching user, accessLevel, and entity, then the message changes to "Cannot update this permission because it has duplicate permission records". That is from the existing unique key constraint message in the MySQL hibernate converter.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: The service update method does catch this and throws an EntityNotFound Exception. I wasn't able to differentiate between "wrong guid" and "wrong version" in the hibernate converter, unfortunately.

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
jnelsonjava: Makes sense, removed userId and renamed the account ref to "account".

src/main/java/org/sagebionetworks/bridge/models/permissions/PermissionDetail.java
jnelsonjava: Removed entityType and entityId, changed entityRef name to "entity".

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Noted and corrected!

src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
jnelsonjava: This was changed to an enum called AccessLevel.

src/main/java/org/sagebionetworks/bridge/models/permissions/Permission.java
jnelsonjava: Noted and updated.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: That is handy. It's updated.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: That makes sense. Dropped the QueryBuilder.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: Done.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
jnelsonjava: An EntityAlreadyExists exception would be a bit clearer. Should I just check to see if the permission already exists here or in the service before actually making the create() call?

src/main/java/org/sagebionetworks/bridge/hibernate/MySQLHibernatePersistenceExceptionConverter.java
jnelsonjava: Changed to flat if/else.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Updated to just null checks.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Added a get call to check if the permission exists first.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: I separated a chunk of the getPermissionDetail() method so I could call it here. It checks that the entity exists and throws ENFE now.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Added a check for the user account before getting the permissions. Now throws ENFE.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Removed the userId, entityType, and entityId fields that are duplicated by the reference types.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
jnelsonjava: Done.

src/main/java/org/sagebionetworks/bridge/spring/controllers/PermissionController.java
jnelsonjava: Done.

src/main/java/org/sagebionetworks/bridge/validators/PermissionValidator.java
jnelsonjava: I'm generating and overwriting the GUID field before validating in the service's create method. Would it be better to add an "isNew" check to the validator and make validation the first step in the service call?

src/main/java/org/sagebionetworks/bridge/models/permissions/AccessLevel.java
jnelsonjava: I added some light descriptions here. Definitely going to need a lot more. The combination of accessLevel and entityType is going to need a lot of attention.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: I added createdOn, modifiedOn, and version to the permission model. Didn't copy over the existing version in the update method so the optimistic lock exception does come through if the permission is submitted with the wrong version. There are integration tests verifying that behavior now as well.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
alxdarksage: Yes it needs to return an EAEE but you don't have to do it here. Either the persistence exception converter can do it, or you can check and throw in the service. As long as the API behavior is the right exception returned, you are good.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernatePermissionDao.java
alxdarksage: Then this is good

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
alxdarksage: This still needs to be a ResourceList, not a list.

src/main/java/org/sagebionetworks/bridge/validators/PermissionValidator.java
alxdarksage: No I think that's fine, sometimes I set values like this and then check them in the validator...it seems robust in case of future changes somewhere.

alxdarksage (approved):

PR #536, opened 2022-03-04
Study adherence report 2 (BRIDGE-3219)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/SessionCompletionState.java
```diff
@@ -22,8 +22,8 @@
     DECLINED;
 
     // These include all sessions except sessions that are not yet available, and thus cannot be included
-    // in a calculation of adherence. 
+    // in a calculation of adherence.
     public static final EnumSet<SessionCompletionState> NONCOMPLIANT = EnumSet.of(ABANDONED, EXPIRED, DECLINED);
     public static final EnumSet<SessionCompletionState> COMPLIANT = EnumSet.of(COMPLETED);
-    public static final EnumSet<SessionCompletionState> UNKNOWN = EnumSet.of(UNSTARTED, STARTED);
+    public static final EnumSet<SessionCompletionState> UNKNOWN = EnumSet.of(UNSTARTED, STARTED, NOT_YET_AVAILABLE);
```
alxdarksage, line 28: So this makes sense for *weekly* adherence, but what about *study-scoped* adherence?

alxdarksage, line 28: So this makes sense for *weekly* adherence, but what about *study-scoped* adherence?

alxdarksage: 0-100 searches of weekly adherence reports would not return records that have NULL, so don't test >0 and <100.

alxdarksage: 0-100 searches of weekly adherence reports would not return records that have NULL, so don't test >0 and <100.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStream.java
```diff
@@ -30,6 +30,9 @@
     public EventStream() {
         byDayEntries = new TreeMap<>(); // maintain the order of the days
     }
+    public EventStream(Map<Integer, List<EventStreamDay>> byDayEntries) {
+        this.byDayEntries = byDayEntries;
```
alxdarksage: This could copy the map.

alxdarksage: This could copy the map.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGenerator.java
```diff
@@ -58,13 +67,37 @@ public EventStreamAdherenceReport generate(AdherenceState state) {
...
+        }
         EventStreamAdherenceReport report = new EventStreamAdherenceReport();
-        report.setActiveOnly(state.showActive());
         report.setTimestamp(state.getNow());
         report.setClientTimeZone(state.getClientTimeZone());
         report.setAdherencePercent(state.calculateAdherencePercentage());
+        report.setDayRangeOfAllStreams(dayRange);
+        report.setDateRangeOfAllStreams(dateRange);
+        report.setEarliestEventId(earliestEventId);
         for (String eventId : state.getStreamEventIds()) {
```
alxdarksage, line 101: There are tests for all this?

alxdarksage, line 101: There are tests for all this?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                        if (oneDay.getStartDate().isEqual(todayLocal)) {
+                            days.forEach(day -> day.setToday(true));
+                        }
+                    }
+                }
+            }
+            if (BridgeUtils.isLocalDateInRange(oneWeek.getStartDate(), null, todayLocal)) {
+                int weekAdh = AdherenceUtils.calculateAdherencePercentage(oneWeek.getByDayEntries());
+                oneWeek.setAdherencePercent(weekAdh);
+            }
```
alxdarksage, line 146: Why does this work?

alxdarksage, line 143: Why does this work?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                // <Session Name>
+                // <Session Name>:Week 1
+                // Week 1
+                String searchableLabel = (oneDay.getStudyBurstId() != null) ?
+                    String.format(":%s:%s %s:Week %s:%s:", oneDay.getStudyBurstId(), oneDay.getStudyBurstId(), oneDay.getStudyBurstNum(), oneWeek.getWeek(), oneDay.getSessionName()) :
+                    String.format(":%s:Week %s:", oneDay.getSessionName(), oneWeek.getWeek());
+                String displayLabel = (oneDay.getStudyBurstId() != null) ?
+                        String.format("%s %s / Week %s / %s", oneDay.getStudyBurstId(), oneDay.getStudyBurstNum(), oneWeek.getWeek(), oneDay.getSessionName()) :
+                        String.format("%s / Week %s", oneDay.getSessionName(), oneWeek.getWeek());
+                labels.add(searchableLabel);
```
alxdarksage, line 207: Check the real world search terms, like "Study Burst #3" which is in the label, but doesn't appear to pull up the record with one of these values.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                .findFirst();
+        if (oneDay.isPresent()) {
+            return oneDay.get();
+        }
+        return new EventStreamDay();
+    }
+    
+    private NextActivity getNextActivity(Collection<StudyReportWeek> weeks, LocalDate localToday) {
+        for (StudyReportWeek oneWeek : weeks) {
+            if (oneWeek.getStartDate().isAfter(localToday)) {
```
alxdarksage, line 261: Wasn't this startDate + 7 days at one point?

alxdarksage, line 258: Wasn't this startDate + 7 days at one point?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeek.java
```diff
@@ -0,0 +1,103 @@
...
+    private LocalDate endDate;
+    private Integer adherencePercent;
+    private Map<Integer, List<EventStreamDay>> byDayEntries;
+    @CollectionTable(name = "WeeklyAdherenceReportLabels", joinColumns = {
+        @JoinColumn(name = "appId"), @JoinColumn(name = "studyId"), @JoinColumn(name = "userId")})
+    @Column(name = "label")
+    @ElementCollection(fetch = FetchType.EAGER)
+    private Set<String> searchableLabels;
+    @Convert(converter = WeeklyAdherenceReportRowListConverter.class)
+    private List<WeeklyAdherenceReportRow> rows;
```
alxdarksage, line 38: These Hibernate annotations can be removed, this object is not directly persisted, these are all already in Weekly AdherenceReport (which is how this information is persisted).

alxdarksage, line 24: These Hibernate annotations can be removed, this object is not directly persisted, these are all already in Weekly AdherenceReport (which is how this information is persisted).

src/main/resources/db/changelog/changelog.sql
```diff
@@ -928,3 +928,8 @@ ADD CONSTRAINT FOREIGN KEY (`sessionGuid`) REFERENCES `Sessions` (`guid`) ON DEL
 
 ALTER TABLE `Substudies`
 ADD COLUMN `studyStartEventId` varchar(255);
+
+-- changeset bridge:64
+
+ALTER TABLE `WeeklyAdherenceReports`
+ADD COLUMN `week` int(4);
```
alxdarksage: Could call is weekInStudy or studyWeek....

alxdarksage: Could call is weekInStudy or studyWeek....

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
```diff
@@ -0,0 +1,603 @@
...
+import org.sagebionetworks.bridge.models.schedules2.adherence.weekly.WeeklyAdherenceReportRow;
+import org.sagebionetworks.bridge.models.schedules2.timelines.Scheduler;
+import org.sagebionetworks.bridge.models.schedules2.timelines.Timeline;
+import org.sagebionetworks.bridge.models.schedules2.timelines.TimelineMetadata;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+public class StudyAdherenceReportGeneratorTest extends Mockito {
```
alxdarksage, line 48: Test that negative weeks due to event ID mismatch work

alxdarksage, line 48: Test that negative weeks due to event ID mismatch work

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeek.java
alxdarksage: Removed

alxdarksage: Removed

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/SessionCompletionState.java
alxdarksage: Yes according to design, in the week and for the study as a whole, adherence starts at 0% and works its way up to 100%.

alxdarksage: On the second, code is adjusted so that adherence range search does not search on this value if the two fields are null, searches on a range unbounded in one direction if one is specified, or both if both are specified. If a record has no min/max values and min/max values are specified in the search, the record is not returned.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStream.java
alxdarksage: Not worth the complexity it would introduce.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/eventstream/EventStreamAdherenceReportGenerator.java
alxdarksage: Yes

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
alxdarksage: Because the range of startDate- matches every week up to the current week, but after that, todayLocal is BEFORE the range, it doesn't match and, so adherence is not calculated for the weeks after the current week. Added comment and test to verify this. It's the correct behavior.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
alxdarksage: This works, but that doesn't guarantee that any label shown by a client app will work as is; in this case you'd have to look for ":Study Burst 3:" if you wanted an exact match for the 3rd study burst.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
alxdarksage: This works; test added (one week's start date is after the local date, so it starts searching at the week after the current week).

src/main/resources/db/changelog/changelog.sql
alxdarksage: Updated to WeekInStudy here and in other places where week is being reported, but not EventStreamDay because the week value there is only set for the event stream report.

src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
```diff
@@ -92,6 +92,20 @@ public void after() {
...
+        LocalDate end = LocalDate.parse("2022-02-17");
+        
+        assertTrue(BridgeUtils.isLocalDateInRange(start, end, LocalDate.parse("2022-02-10")));
+        assertTrue(BridgeUtils.isLocalDateInRange(start, end, LocalDate.parse("2022-02-17")));
+        assertTrue(BridgeUtils.isLocalDateInRange(null, end, LocalDate.parse("2022-02-01")));
+        assertTrue(BridgeUtils.isLocalDateInRange(start, null, LocalDate.parse("2022-02-17")));
+        assertFalse(BridgeUtils.isLocalDateInRange(start, end, LocalDate.parse("2022-02-09")));
+        assertFalse(BridgeUtils.isLocalDateInRange(start, end, LocalDate.parse("2022-02-18")));
+        assertFalse(BridgeUtils.isLocalDateInRange(start, end, null));
+    }
```
jnelsonjava: Could optionally add:
assertFalse(BridgeUtils.isLocalDateInRange(null, end, LocalDate.parse("2022-02-18")));
assertFalse(BridgeUtils.isLocalDateInRange(start, null, LocalDate.parse("2022-02-09")));

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeek.java
```diff
@@ -0,0 +1,90 @@
...
+import org.sagebionetworks.bridge.models.schedules2.adherence.eventstream.EventStreamDay;
+import org.sagebionetworks.bridge.models.schedules2.adherence.weekly.WeeklyAdherenceReportRow;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder({ "startDate", "endDate", "week", "adherencePercent", "rows", "byDayEntries", "type" })
+public class StudyReportWeek {
+    
+    private int weekInStudy;
```
jnelsonjava: "week" looks like it should change to "weekInStudy".

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeekTest.java
```diff
@@ -0,0 +1,46 @@
...
+        JsonNode node = BridgeObjectMapper.get().valueToTree(week);
+        assertEquals(node.size(), 7);
+        assertEquals(node.get("startDate").textValue(), "2022-02-10");
+        assertEquals(node.get("endDate").textValue(), "2022-05-11");
+        assertEquals(node.get("adherencePercent").intValue(), 36);
+        assertEquals(node.get("rows").get(0).get("label").textValue(), "labelRow");
+        assertEquals(node.get("rows").get(0).get("type").textValue(), "WeeklyAdherenceReportRow");
+        assertEquals(node.get("byDayEntries").get("5").get(0).get("label").textValue(), "dayLabel");
+        assertEquals(node.get("byDayEntries").get("5").get(0).get("type").textValue(), "EventStreamDay");
+        assertEquals(node.get("type").textValue(), "StudyReportWeek");
```
jnelsonjava: Should this include "week" (or "weekInStudy")?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/weekly/WeeklyAdherenceReport.java
```diff
@@ -37,7 +37,7 @@
 @Entity
 @Table(name = "WeeklyAdherenceReports")
 @IdClass(WeeklyAdherenceReportId.class)
-@JsonPropertyOrder({ "participant", "testAccount", "progression", "weeklyAdherencePercent", "clientTimeZone",
+@JsonPropertyOrder({ "participant", "testAccount", "progression", "weeklyAdherencePercent", "week", "clientTimeZone",
```
jnelsonjava: "week" -> "weekInStudy"

jnelsonjava (approved):
Looks good. Check out the few comments, mostly minor JsonPropertyOrder edits.

src/test/java/org/sagebionetworks/bridge/BridgeUtilsTest.java
alxdarksage: added

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeek.java
alxdarksage: Good catch, changed

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyReportWeekTest.java
alxdarksage: Yes, added

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/weekly/WeeklyAdherenceReport.java
alxdarksage: yikes. changed here too

PR #541, opened 2022-03-11
Export for Studies

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2857


src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -174,19 +175,28 @@ public Study finishStudyLogo(@PathVariable String id, @PathVariable("createdOn")
...
+        } else {
+            CacheKey key = CacheKey.publicStudy(appId, studyId);
+            String json = cacheProvider.getObject(key, String.class);
+            if (json == null) {
+                appService.getApp(appId);
+                Study study = service.getStudy(appId, studyId, true);
+                json = Study.STUDY_SUMMARY_WRITER.writeValueAsString(study);
+                cacheProvider.setObject(key, json, ONE_DAY_IN_SECONDS);
+            }
+            return json;
```
alxdarksage: Does anything use this other than a worker? I wonder if we created this then never used it as a public endpoint like this...

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
```diff
@@ -247,7 +251,27 @@ public String getDetails() {
...
+
+    @Override
+    public void setExporter3Configuration(Exporter3Configuration exporter3Configuration) {
+        this.exporter3Configuration = exporter3Configuration;
+    }
+
+    @Override
+    public boolean isExporter3Enabled() {
+        return exporter3Enabled;
+    }
```
DwayneJengSage, line 285: Add an AuthUtils check in updateStudy so that only ORG_ADMIN can turn this on/off. (We can later refactor this as part of the permissions stuff Jack is working on.)

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -174,19 +175,28 @@ public Study finishStudyLogo(@PathVariable String id, @PathVariable("createdOn")
     }
 
     @GetMapping(path = "/v1/apps/{appId}/studies/{studyId}", produces = { APPLICATION_JSON_UTF8_VALUE })
```
DwayneJengSage, line 179: Separate this out into a /v2/ URL for the Worker API.

Add a comment about why v1 exists. This is used to show app participants some basic study info before they create their accounts.

src/main/java/org/sagebionetworks/bridge/spring/controllers/StudyController.java
```diff
@@ -174,19 +175,28 @@ public Study finishStudyLogo(@PathVariable String id, @PathVariable("createdOn")
...
+    public String getStudyForApp(@PathVariable String appId, @PathVariable String studyId,
+            @RequestParam(required = false) String full)
             throws JsonProcessingException {
-        CacheKey key = CacheKey.publicStudy(appId, studyId);
-        String json = cacheProvider.getObject(key, String.class);
-        if (json == null) {
+        boolean fullBool = Boolean.parseBoolean(full);
+        if (fullBool) {
+            getAuthenticatedSession(WORKER);
             appService.getApp(appId);
```
DwayneJengSage: Add comment: This will throw if the app doesn't exist, with an error that says the app doesn't exist.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateStudy.java
DwayneJengSage: StudyController.updateStudy() uses `CAN_UPDATE_STUDIES.checkAndThrow(STUDY_ID, studyId);`, so I added that to Exporter3Controller.initForStudy().

Also, Schedule2Controller.createOrUpdateSchedule() uses `getAuthenticatedSession(STUDY_DESIGNER, DEVELOPER);`, so I updated Exporter3Controller to use that as well.

src/main/java/org/sagebionetworks/bridge/spring/controllers/Exporter3Controller.java
```diff
@@ -35,4 +39,13 @@ public Exporter3Configuration initExporter3() throws BridgeSynapseException, IOE
         UserSession session = getAuthenticatedSession(DEVELOPER);
         return exporter3Service.initExporter3(session.getAppId());
     }
+
+    @PostMapping("/v5/studies/{studyId}/exporter3")
+    @ResponseStatus(HttpStatus.CREATED)
+    public Exporter3Configuration initExporter3ForStudy(@PathVariable String studyId) throws BridgeSynapseException, IOException,
+            SynapseException {
+        UserSession session = getAuthenticatedSession(STUDY_DESIGNER, DEVELOPER);
```
jnelsonjava, line 47: Just noting the roles here don't match the SDK. Looking at it now, I think this is correct and the SDK should just change to reflect these.

jnelsonjava (approved):
Looks great!

PR #542, opened 2022-03-22
Carry over unfinished and unexpired activities from prior weeks into the weekly adherence report (BRIDGE-3262)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/BridgeUtils.java
```diff
@@ -77,6 +78,7 @@
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper.FailedBatch;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.amazonaws.util.Throwables;
+import com.fasterxml.jackson.core.JsonProcessingException;
```
alxdarksage, line 81: Remove unused imports

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
```diff
@@ -272,4 +273,83 @@ private NextActivity getNextActivity(Collection<StudyReportWeek> weeks, LocalDat
...
+        for (StudyReportWeek oneWeek : weeks) {
+            if (isLocalDateInRange(weekReport.getStartDate(), null, oneWeek.getStartDate())) {
+                break;
+            }
+            oneWeek.visitDays((day, i) -> {
+                boolean match = day.getTimeWindows().stream()
+                        .anyMatch(window -> UNKNOWN.contains(window.getState()));
+                if (match) {
+                    EventStreamDay dayCopy = day.copy();
+                    dayCopy.setToday(weekReport.getStartDate().isEqual(localToday));
```
alxdarksage: You should be able to remove line 313

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
```diff
@@ -248,80 +256,691 @@ public static Schedule2 createSchedule() {
...
+        assertEquals(report.getUnsetEventIds(), ImmutableSet.of("custom:event1", "timeline_retrieved", 
+                "study_burst:Study Burst:01", "study_burst:Study Burst:02", "study_burst:Study Burst:03"));
+        assertEquals(report.getUnscheduledSessions(), ImmutableSet.of("Supplemental Survey", "Initial Survey",
+                "Baseline Tapping Test", "Study Burst Tapping Test", "Final Survey"));
+        assertTrue(report.getEventTimestamps().isEmpty());
+
+        assertEquals(report.getWeeks().size(), 0);
+        
+        // Verify this too
+        assertNull(report.getWeekReport().getAdherencePercent());
```
alxdarksage, line 598: Is there more to verify about weekly report in this and other cases?

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
```diff
@@ -248,80 +256,691 @@ public static Schedule2 createSchedule() {
...
+        AdherenceState.Builder builder = createAdherenceState();
+        // remove timeline_retrieved
+        builder.withEvents(createEvents().subList(1,  createEvents().size()));
+        builder.withAdherenceRecords(createAdherenceRecords());
+        builder.withClientTimeZone("America/Chicago");
+        StudyAdherenceReport report = INSTANCE.generate(builder.build());
+        
+        // testAccount is set in the service, not the generator
+        assertEquals(report.getAdherencePercent(), Integer.valueOf(0));
+        // assertEquals(report.getProgression(), ParticipantStudyProgress.IN_PROGRESS);
```
alxdarksage: why is this commented out?

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
```diff
@@ -248,80 +256,691 @@ public static Schedule2 createSchedule() {
...
+        assertEquals(week1Day2.get(0).getSessionGuid(), "baselineGuid");
+        assertEquals(week1Day2.get(0).getStartEventId(), "timeline_retrieved");
+        assertTimeWindow(week1Day2.get(0).getTimeWindows().get(0), "xyvAcmEYAVAzCMfGhf187g", 
+                "win2", COMPLETED, "2022-03-03");
+        
+        assertEquals(week1Day2.get(0).getSessionGuid(), "baselineGuid");
+        assertEquals(week1Day2.get(0).getStartEventId(), "timeline_retrieved");
+        assertEquals(week1Day2.get(0).getStartDate(), LocalDate.parse("2022-03-03"));
+        assertTimeWindow(week1Day2.get(0).getTimeWindows().get(0), "xyvAcmEYAVAzCMfGhf187g", 
+                "win2", COMPLETED, "2022-03-03");
```
alxdarksage, line 779: This block is a duplicate of the block above it, remove

alxdarksage:
In addition:

- Are we doing too much work when studyStartEvent timestamp is null? This user is basically unstarted, so most of the report generator can be circumvented? [ANSWER: No, it's not too much work. There's some iteration we have to do in this case to collect the unused event IDs and sessions, and in the current code we do that, then skip all the more expensive loops where we're calculating week and day data, because none exist. Leave as is.]
- When calculating the weekly report that's persisted (just one week), should we include the progression state? Right now that's only in the full study adherence report, but for the current week it would be much easier to report unstarted, done, etc.  [ANSWER: No, I was confused in this review. The week is used to generate the weekly adherence report, but that report contains more than just the one current week. Progression is in the final weekly report and we use it in the BSM.]
- NextActivity: could be set even if the current week has carry-overs. What do we want here? Both? Should we not calculate next activity if there are carry-overs? In this case, the test needs to be moved down in the code and needs to look at something else to determine if a next activity is needed.

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGenerator.java
alxdarksage: No, we need it, and it doesn't work here. I now track these carry over days and reset the value after calling the unused method, where it is set to the startDate of this day, which we already know isn't in the week (on purpose).

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
alxdarksage: No, this test is just about looking at the adherence percent and ensuring it is not generated where it should not be generated. This week is empty so there's no percent.

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
alxdarksage: added it back (it passes)

src/test/java/org/sagebionetworks/bridge/models/schedules2/adherence/study/StudyAdherenceReportGeneratorTest.java
alxdarksage: removed

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -424,15 +424,16 @@ protected WeeklyAdherenceReport deriveWeeklyAdherenceFromStudyAdherenceReport(St
...
-            weeklyReport.setRows(week.getRows());
-            weeklyReport.setByDayEntries(week.getByDayEntries());
-            weeklyReport.setWeeklyAdherencePercent(week.getAdherencePercent());
-            weeklyReport.setWeekInStudy(week.getWeekInStudy());
-            weeklyReport.setStartDate(week.getStartDate());
+        
+        StudyReportWeek week = report.getWeekReport();
+        weeklyReport.setSearchableLabels(week.getSearchableLabels());
+        weeklyReport.setRows(week.getRows());
+        weeklyReport.setByDayEntries(week.getByDayEntries());
```
jnelsonjava, line 431: These setters were removed from StudyReportWeek.

alxdarksage, line 437: Those setters are there. weeklyReport is a WeeklyAdherenceReport, not a StudyReportWeek.

jnelsonjava (approved):
Looks good!

PR #552, opened 2022-03-30
Add API & Unit tests to get S3 pre-signed URL for a record ID

chuguoxipo:
Jira: bridge-2609

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
@@ -39,7 +39,6 @@ public class HealthDataEx3Service {
     private HealthDataEx3Dao healthDataEx3Dao;
     private UploadService uploadService;
     private String S3bucketName;
-    private String S3Key;
     private AmazonS3 s3Client;
 
     @Autowired
@@ -87,10 +86,9 @@ public class HealthDataEx3Service {
             throw new BadRequestException("ID must be specified");
         }
 
-        Optional<HealthDataRecordEx3> record = Optional.ofNullable(healthDataEx3Dao.getRecord(id))
-                .orElseThrow(() -> new EntityNotFoundException(HealthDataRecordEx3.class));
+        Optional<HealthDataRecordEx3> record = healthDataEx3Dao.getRecord(id);
 
-        if (download) {
+        if (record != null && download) {
             record.get().setDownloadUrl(generatePresignedUrl(record.get(), GET).toExternalForm());
         }
         return record;
@@ -105,7 +103,7 @@ public class HealthDataEx3Service {
         long expiration = DateTime.now().plusMinutes(EXPIRATION_IN_MINUTES).getMillis();
         record.setDownloadExpiration(expiration);
 
-        S3Key = UploadUtil.getRawS3KeyForUpload(record.getAppId(), record.getStudyId(),
+        String S3Key = UploadUtil.getRawS3KeyForUpload(record.getAppId(),
                 this.uploadService.getUpload(record.getId()), record);
 
         GeneratePresignedUrlRequest request = new GeneratePresignedUrlRequest(S3bucketName , S3Key, method);
--- a/src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
@@ -1,18 +1,30 @@
 package org.sagebionetworks.bridge.services;
 
+import java.net.URL;
+import java.util.Date;
 import java.util.Optional;
 
+import com.amazonaws.HttpMethod;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.Headers;
+import static com.amazonaws.HttpMethod.GET;
+import com.amazonaws.services.s3.model.GeneratePresignedUrlRequest;
+import com.amazonaws.services.s3.model.ObjectMetadata;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+
 import org.sagebionetworks.bridge.BridgeConstants;
+import org.sagebionetworks.bridge.config.BridgeConfig;
 import org.sagebionetworks.bridge.dao.HealthDataEx3Dao;
 import org.sagebionetworks.bridge.exceptions.BadRequestException;
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
 import org.sagebionetworks.bridge.exceptions.InvalidEntityException;
 import org.sagebionetworks.bridge.models.ForwardCursorPagedResourceList;
 import org.sagebionetworks.bridge.models.healthdata.HealthDataRecordEx3;
+import org.sagebionetworks.bridge.upload.UploadUtil;
 import org.sagebionetworks.bridge.validators.HealthDataRecordEx3Validator;
 import org.sagebionetworks.bridge.validators.Validate;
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
```diff
@@ -23,14 +34,34 @@
 @Component
 public class HealthDataEx3Service {
     static final int MAX_DATE_RANGE_DAYS = 60;
+    static final int EXPIRATION_IN_MINUTES = 60;
 
     private HealthDataEx3Dao healthDataEx3Dao;
+    private UploadService uploadService;
+    private String S3bucketName;
+    private String S3Key;
```
DwayneJengSage: Minor: Looks like s3Key is only used in generatePresignedUrl(). Instead of making a class member, we should make this local to that specific method.

src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
```diff
@@ -50,13 +81,37 @@ public void deleteRecordsForHealthCode(String healthCode) {
...
         }
 
-        return healthDataEx3Dao.getRecord(id);
+        Optional<HealthDataRecordEx3> record = Optional.ofNullable(healthDataEx3Dao.getRecord(id))
+                .orElseThrow(() -> new EntityNotFoundException(HealthDataRecordEx3.class));
+
+        if (download) {
+            record.get().setDownloadUrl(generatePresignedUrl(record.get(), GET).toExternalForm());
+        }
+        return record;
```
DwayneJengSage: This changes a lot of the behavior of the method. Previously, if the record doesn't exist, it would return an empty Optional. Now it throws. Also, it looks like you are wrapping an Optional inside of an Optional.

You should instead check that the Optional that dao.getRecord() returns is present, and if it is and if the download flag is present, you can generate and set the download link.

src/test/java/org/sagebionetworks/bridge/services/HealthDataEx3ServiceTest.java
```diff
@@ -33,16 +43,43 @@
...
+        when(mockConfig.getProperty(Exporter3Service.CONFIG_KEY_RAW_HEALTH_DATA_BUCKET)).thenReturn(RECORD_BUCKET);
+        service.setConfig(mockConfig);
+
+        when(mockS3Client.generatePresignedUrl(any())).thenAnswer(i -> {
+            GeneratePresignedUrlRequest request = i.getArgument(0);
+            String filePath = request.getKey();
+            return new URL("https://" + RECORD_BUCKET + "/" + filePath);
+        });
+
+        DateTimeUtils.setCurrentMillisFixed(TestConstants.TIMESTAMP.getMillis());
```
DwayneJengSage, line 84: We should add a method with @AfterMethod and call DateTimeUtils.setCurrentMillisSystem(). This will reset DateTimeUtils to use the system clock.

src/main/java/org/sagebionetworks/bridge/upload/UploadUtil.java
```diff
@@ -77,6 +80,7 @@
 
     // Misc constants
     private static final int DEFAULT_MAX_LENGTH = 100;
+    public static final DateTimeZone LOCAL_TIME_ZONE = DateTimeZone.forID("America/Los_Angeles");
```
DwayneJengSage: This already exists in BridgeConstants. There's no need to create your own.

src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
```diff
@@ -50,13 +81,37 @@ public void deleteRecordsForHealthCode(String healthCode) {
...
+     * Returns the url path of the given record.
+     * @param record the record
+     * @return the url path of the given record.
+     */
+    private URL generatePresignedUrl(HealthDataRecordEx3 record, HttpMethod method) {
+        long expiration = DateTime.now().plusMinutes(EXPIRATION_IN_MINUTES).getMillis();
+        record.setDownloadExpiration(expiration);
+
+        S3Key = UploadUtil.getRawS3KeyForUpload(record.getAppId(), record.getStudyId(),
+                this.uploadService.getUpload(record.getId()), record);
```
DwayneJengSage: Digging through the code, it looks like we no longer use record.getStudyId(). Instead, we get the studyId from the Participant, and the health record can be exported to multiple studies.

For now, just assume that the uploaded data will always be under the app and ignore the study for creating the S3Key.

Similarly, we can remove studyId from UploadUtils and from the expected download URL in the tests.

src/main/java/org/sagebionetworks/bridge/upload/UploadUtil.java
```diff
@@ -756,4 +760,32 @@ public static DateTime parseIosTimestamp(String timestampStr) {
         }
         return sanitizedFieldMap;
     }
+
+    public static String getRawS3KeyForUpload(String appId, String studyId, Upload upload, HealthDataRecordEx3 record) {
```
DwayneJengSage: Can we add a unit test to UploadUtilTest to make sure that this method works?

src/main/java/org/sagebionetworks/bridge/spring/controllers/HealthDataEx3Controller.java
```diff
@@ -178,10 +178,10 @@ public HealthDataRecordEx3 getRecord(@PathVariable String appId, @PathVariable S
 
     /** Retrieves the record for the given ID for self. */
     @GetMapping(path="/v3/participants/self/exporter3/healthdata/{recordId}")
-    public HealthDataRecordEx3 getRecordForSelf(@PathVariable String recordId) {
+    public HealthDataRecordEx3 getRecordForSelf(@PathVariable String recordId, @RequestParam(required = false) String download) {
         UserSession session = getAuthenticatedAndConsentedSession();
 
-        HealthDataRecordEx3 record = healthDataEx3Service.getRecord(recordId).orElseThrow(() ->
+        HealthDataRecordEx3 record = healthDataEx3Service.getRecord(recordId, "true".equalsIgnoreCase(download)).orElseThrow(() ->
```
DwayneJengSage, line 184: Add unit tests for the true and null case, to make sure we are parsing the boolean flag correctly.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/HealthDataEx3Service.java
```diff
@@ -50,13 +80,36 @@ public void deleteRecordsForHealthCode(String healthCode) {
...
         if (StringUtils.isBlank(id)) {
             throw new BadRequestException("ID must be specified");
         }
 
-        return healthDataEx3Dao.getRecord(id);
+        Optional<HealthDataRecordEx3> record = healthDataEx3Dao.getRecord(id);
+
+        if (record != null && download) {
+            record.get().setDownloadUrl(generatePresignedUrl(record.get(), GET).toExternalForm());
+        }
```
DwayneJengSage: record is guaranteed to not be null, record.get() will throw if the record doesn't exist. You should check record.isPresent() instead.

Also, add a unit test for Optional.empty() to make sure this code path works correctly.

src/test/java/org/sagebionetworks/bridge/upload/UploadUtilTest.java
```diff
@@ -482,7 +488,7 @@ public void validFieldNameAndAnswerChoice() {
                 "foo bar",
                 "foo-bar_baz.qwerty asdf",
                 "select",
-                "where",
+                "whervre",
```
DwayneJengSage: Is this a typo?

DwayneJengSage (changes requested):

DwayneJengSage (approved):

PR #559, opened 2022-04-05
Mark admin account (BRIDGE-3266)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -302,10 +303,13 @@ public void updateAccount(Account account) {
         account.setOrgMembership(persistedAccount.getOrgMembership());
         // Update modifiedOn.
         account.setModifiedOn(DateUtils.getCurrentDateTime());
+        account.setAdmin(persistedAccount.isAdmin());
```
alxdarksage: You will need to duplicate the above logic until this flag is set by the particular API being called.

jnelsonjava (approved):
Looks good.

PR #560, opened 2022-04-08
Refactor: move authentication and workflow code out of participant and account services (BRIDGE-3269)

alxdarksage:


jnelsonjava (approved):
Looks good.

PR #566, opened 2022-04-13
Further session validation (BRIDGE-3265)

alxdarksage:
Interval and occurrence work together. You can't have one without the other (and expiration is needed if either is set).


jnelsonjava (approved):
Looks good.

PR #567, opened 2022-04-14
Allowing clientTimeZone to be submitted when requesting the participant schedule (BRIDGE-3258)

alxdarksage:
This required specifying the property and looking for it in the ETag implementation, busting the cache if the value is going to change. The code assumes the method it is wrapping will update the value, so it recalculates the tag after execution.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
@@ -6,6 +6,7 @@ import static org.sagebionetworks.bridge.TestConstants.ACCOUNT_ID;
 import static org.sagebionetworks.bridge.TestConstants.CREATED_ON;
 import static org.sagebionetworks.bridge.TestConstants.MODIFIED_ON;
 import static org.sagebionetworks.bridge.TestConstants.TEST_APP_ID;
+import static org.sagebionetworks.bridge.TestConstants.TEST_CLIENT_TIME_ZONE;
 import static org.sagebionetworks.bridge.TestConstants.TEST_ORG_ID;
 import static org.sagebionetworks.bridge.TestConstants.TEST_STUDY_ID;
 import static org.sagebionetworks.bridge.TestConstants.TEST_USER_ID;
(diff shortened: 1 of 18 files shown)


src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
```diff
@@ -264,4 +298,112 @@ public void nullBridgeSessionHeader() throws Throwable {
...
+        };
+        // This just gets ignored
+        doReturn(ImmutableList.of(badTimeZoneAnn)).when(mockContext).getCacheKeys();
+
+        when(mockRequest.getHeader(IF_NONE_MATCH)).thenReturn(ETAG);
+        
+        component.checkEtag(mockJoinPoint);
+        
+        verify(mockCacheProvider, never()).removeObject(any());
+    }
```
alxdarksage, line 329: Maybe this should throw an exception-this would be a code configuration error, in fact, if this is not a key in the argument map—it's not a query parameter, and that is ALSO a configuration exception.

src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
```diff
@@ -264,4 +298,112 @@ public void nullBridgeSessionHeader() throws Throwable {
...
+        when(mockCacheProvider.getObject(
+                CacheKey.etag(DateTimeZone.class, TEST_USER_ID), DateTime.class)).thenReturn(MODIFIED_ON);
+        when(mockRequest.getHeader(IF_NONE_MATCH)).thenReturn(ETAG);
+        
+        when(mockMd5DigestUtils.digest(MODIFIED_ON.toString().getBytes())).thenReturn("ETAG".getBytes());
+        
+        component.checkEtag(mockJoinPoint);
+        
+        verify(mockCacheProvider).removeObject(CacheKey.etag(DateTimeZone.class, TEST_USER_ID));
+        verify(mockMd5DigestUtils).digest((byte[])any());
```
alxdarksage, line 388: so does this run twice, it would have to recalculate the tag...in which case, isn't it the same case as below?

src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
alxdarksage: This now throws an IllegalArgumentException

src/test/java/org/sagebionetworks/bridge/spring/util/EtagComponentTest.java
alxdarksage: Added some notes on why these are different, and a couple more assertions to each.

jnelsonjava (approved):
Looks good

PR #568, opened 2022-04-15
Security

alxdarksage:
This works and brings all our versions up-to-date, however, it requires Tomcat 9. If you try and force use of Tomcat 8.5, there are errors on start-up. I'm not clear if we have to stay on Tomcat 8.5 because of EBS (if so, this can't be merged and we can't update to address BRIDGE-3279 until we do).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,32 +11,26 @@
     <properties>
         <apache.http.version>4.5.13</apache.http.version>
         <aws.version>1.11.944</aws.version>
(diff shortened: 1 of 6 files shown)


DwayneJengSage:
As per comment on Jira, this will require changes to Infra, which could be potentially up to an entire Sprint's worth of work.

DwayneJengSage:
Closing obsolete pull request.

PR #570, opened 2022-04-18
Deprecations and warnings cleaned up

alxdarksage:
Updated code to remove deprecations and adjusted the pom.xml file to use the dependencies introduced by Spring Boot when we're not actually overriding them (although we do override a fair number of these dependencies).

I added a comment to the pom.xml file that hides useless pom warnings in Eclipse; I left in lot of @SuppressWarnings annotations that seem to be about suppressing warnings in IntelliJ.

All unit and integration tests pass.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
@@ -27,9 +27,7 @@ import org.springframework.stereotype.Component;
 
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.AmazonServiceException;
-import com.amazonaws.regions.Region;
-import com.amazonaws.regions.Regions;
-import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClient;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailService;
 import com.amazonaws.services.simpleemail.model.RawMessage;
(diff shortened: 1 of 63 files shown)


jnelsonjava (approved):
Looks good.

src/main/java/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
```diff
@@ -103,7 +100,6 @@ private void sendEmail(String senderEmail, String recipient, MimeTypeEmail email
         SendRawEmailRequest req = new SendRawEmailRequest(sesRawMessage);
         req.setSource(senderEmail);
         req.setDestinations(Collections.singleton(recipient));
-        emailClient.setRegion(REGION);
         SendRawEmailResult result = emailClient.sendRawEmail(req);
```
DwayneJengSage, line 103: Can we manually verify that Bridge still sends email even if setRegion() is not set?

DwayneJengSage (approved):
Overall looks good. One comment about setRegion() and SES.

src/main/java/org/sagebionetworks/bridge/services/SendMailViaAmazonService.java
alxdarksage: Yes, it sends email. The deprecation warning on the field says to use the "withRegion()" method of the client builder so I feel pretty confident this is correct.

PR #574, opened 2022-04-20
For users who have not signed in, adherence reports are always report as UNSTARTED (BRIDGE-3299)

alxdarksage:


jnelsonjava (approved):
Looks good.

PR #580, opened 2022-04-22
Fixing duplicate key error (BRIDGE-3179)

alxdarksage:
The duplication of keys in this list, was causing a Collector to throw an exception elsewhere in the code due to duplicate keys. The CREATED_ON timestamp can be persisted in the study-specific events table and when that happens, we can ignore the value in the app-scoped table. The service should not return two entries for this value.

Also I added a couple of real-world UA strings to that test to double-check that some proposed UA strings from the mobile team will be parsed correctly (they will).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
@@ -1,7 +1,6 @@
 package org.sagebionetworks.bridge.services;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static java.util.stream.Collectors.toSet;
 import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
 import static org.sagebionetworks.bridge.BridgeConstants.API_MINIMUM_PAGE_SIZE;
 import static org.sagebionetworks.bridge.BridgeConstants.NEGATIVE_OFFSET_ERROR;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
```diff
@@ -310,7 +313,8 @@ public PagedResourceList<StudyActivityEvent> getStudyActivityEventHistory(
     }
     
     private void addIfPresent(List<StudyActivityEvent> events, Map<String, DateTime> map, String field, boolean addCount) {
-        if (map.containsKey(field)) {
+        Set<String> existingFields = events.stream().map(StudyActivityEvent::getEventId).collect(toSet());
+        if (map.containsKey(field) && !existingFields.contains(field)) {
```
DwayneJengSage: Minor optimization: Put the existingFields check in a nested if statement. This way, if the map doesn't contain the field, we don't have to walk the event stream to get all the existing fields. This could save some CPU cycles if the event stream is very long.

Also, consider using anyMatch() instead of collect(toSet()), which will short-circuit if it finds any, instead of walking the whole list.

DwayneJengSage (approved):
Looks good. One comment about a minor optimization.

src/main/java/org/sagebionetworks/bridge/services/StudyActivityEventService.java
alxdarksage: Changed this to only create a stream if the first check passes, and switched to anyMatch.

DwayneJengSage (approved):

PR #582, opened 2022-04-27
Fix for BRIDGE-3283

alxdarksage:
Version attributes that use the Long object will accept null values via the API, which lead to null pointer errors in Hibernate code. Primitive long values at least deserialize to 0L when this occurs, and that correctly throws a concurrent modification exception. Switching these fields to use primitives (this issue does not seem to exist for DynamoDB, only for Hibernate-backed SQL entities).


jnelsonjava (approved):
Looks good.

PR #583, opened 2022-04-27
Fix for BRIDGE-3290

alxdarksage:
Sign out accounts on a change of the synapse user ID. When an ID is changed, we want that user to authenticate so we know they control that account.

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
@@ -889,6 +889,75 @@ public class AdminAccountServiceTest extends Mockito {
     }
     
     @Test
+    public void updateAccount_noChangeToSynapseUserId() {
+        App app = App.create();
+        when(mockAppService.getApp(TEST_APP_ID)).thenReturn(app);
+        
+        Account persistedAccount = Account.create();
+        persistedAccount.setAdmin(TRUE);
+        persistedAccount.setId(TEST_USER_ID);
+        persistedAccount.setAppId(TEST_APP_ID);
+        persistedAccount.setSynapseUserId(SYNAPSE_USER_ID);
+
+        when(mockAccountDao.getAccount(any())).thenReturn(Optional.of(persistedAccount));
+
+        Account account = Account.create();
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
```diff
@@ -888,6 +888,75 @@ public void updateAccount_workerCanChangeStatus() {
         assertEquals(retValue.getStatus(), ENABLED);
     }
     
+    @Test
+    public void updateAccount_noChangeToSynapseUserId() {
```
jnelsonjava: This test name indicates no change, but the Synapse user ID does change.

src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
```diff
@@ -888,6 +888,75 @@ public void updateAccount_workerCanChangeStatus() {
...
+        account.setEmail(EMAIL);
+        
+        Account retValue = service.updateAccount(TEST_APP_ID, account);
+        assertNull(retValue.getSynapseUserId());
+        
+        verify(mockCacheProvider, never()).removeSessionByUserId(TEST_USER_ID);
+    }
+    
+    @Test
+    public void updateAccount_changeSynapseUserIdSignsOutUser() {
```
jnelsonjava: Opposite of the test above, name indicates ID changes but test checks "no change" behavior.

jnelsonjava (approved):
Approved but check out the tests with comments. Can just swap the names or logic between them.

src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
alxdarksage: Thanks for catching this, renamed

src/test/java/org/sagebionetworks/bridge/services/AdminAccountServiceTest.java
alxdarksage: Thanks for catching this, renamed

PR #584, opened 2022-04-28
Bridge 3273: Schedule-dependent APIs return 404 and not empty artifacts

alxdarksage:
This is a lot cleaner and more consistent for the API.

### DIFF
--- a/.zschr
+++ b/.zschr
@@ -1,2 +1,0 @@
-
-export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk
--- a/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
@@ -19,6 +19,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.models.schedules2.Schedule2;
 import org.sagebionetworks.bridge.models.schedules2.adherence.eventstream.EventStream;
 import org.sagebionetworks.bridge.models.schedules2.adherence.eventstream.EventStreamDay;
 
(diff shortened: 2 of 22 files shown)


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
```diff
@@ -73,7 +75,7 @@ public static int calculateAdherencePercentage(Collection<EventStream> streams)
     
     public static ParticipantStudyProgress calculateProgress(AdherenceState state, List<EventStream> eventStreams) {
         if (state.getMetadata().isEmpty()) {
-            return ParticipantStudyProgress.NO_SCHEDULE;
+            throw new EntityNotFoundException(Schedule2.class);
         }
```
alxdarksage: So could you create an empty schedule (yes). Then is it still 404?

src/main/java/org/sagebionetworks/bridge/models/schedules2/participantschedules/ParticipantSchedule.java
```diff
@@ -25,11 +27,11 @@ public class ParticipantSchedule {
...
-    private List<ScheduledSession> scheduledSessions;
-    private List<AssessmentInfo> assessments;
-    private List<SessionInfo> sessions;
-    private List<StudyBurstInfo> studyBursts;
-    private Map<String, DateTime> eventTimestamps;
+    private List<ScheduledSession> scheduledSessions = ImmutableList.of();
+    private List<AssessmentInfo> assessments = ImmutableList.of();
+    private List<SessionInfo> sessions = ImmutableList.of();
+    private List<StudyBurstInfo> studyBursts = ImmutableList.of();
+    private Map<String, DateTime> eventTimestamps = ImmutableMap.of();
```
alxdarksage: Revert this

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceUtils.java
alxdarksage: Removed this, as I don't believe it is necessary. It'll just calculate that an empty schedule is unstarted, and elsewhere in the code you don't get here if there's no schedule, because you never generate the report. Adjusted test to show this is what happens (UNSTARTED...an empty schedule can't really be started or completed).

src/main/java/org/sagebionetworks/bridge/models/schedules2/participantschedules/ParticipantSchedule.java
alxdarksage: reverted

.zschr
```diff
@@ -0,0 +1,2 @@
+
+export SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk
```
jnelsonjava: This looks like it snuck in.

jnelsonjava (approved):
Looks good.

PR #586, opened 2022-04-29
migrate from AWS keys to AWS Role

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3188

Passing in AWS keys as env vars is a security risk and is also unnecessary. By not passing in credentials at all, Elastic Beanstalk will automatically pick them up from the AWS Role. See https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html

I also removed the UploadSessionsCredentialsService. This was a carry-over from how we used to do permissions back on Heroku, and with our current set up, it doesn't actually do anything. Furthermore, it only works with IAM Users, not with IAM Roles. So I deleted it.

I tested this on my laptop, and I also tested this on our Dev instance, by pointing my laptop's integration tests at our Dev instance.

Requires https://github.com/Sage-Bionetworks/BridgeServer2-infra/pull/56


alxdarksage:
Locally, does this then pick up credentials from the ~/.aws/credentials file?

alxdarksage (approved):

PR #589, opened 2022-05-05
BRIDGE-3268: consolidate account deletion methods

alxdarksage:
There is now one canonical delete method in AccountService that handles clearing all the caches, the non-transactional DynamoDB tables, etc.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -13,7 +13,6 @@ import static org.sagebionetworks.bridge.BridgeUtils.collectStudyIds;
 import static org.sagebionetworks.bridge.dao.AccountDao.MIGRATION_VERSION;
 import static org.sagebionetworks.bridge.models.activities.ActivityEventObjectType.ENROLLMENT;
 
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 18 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -94,7 +84,7 @@ protected String generateGUID() {
      * Search for all accounts across apps that have the same Synapse user ID in common, 
      * and return a list of the app IDs where these accounts are found.
      */
-    public List<String> getAppIdsForUser(String synapseUserId) {
+    public List<String> getAppIdsForUser2(String synapseUserId) {
```
jnelsonjava: Just curious what this rename is for? Are you differentiating from the same name method in AdminAccountService?

jnelsonjava: Actually should this method just be deleted?

jnelsonjava (approved):
Approved but check out the commented method in AccountService. I'm pretty sure it can just be removed now.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: Yes this was a step to ensuring all code was pointed at the same method in AdminAccountService; I will delete it.

PR #593, opened 2022-05-09
Search for declined (BRIDGE-3303)

alxdarksage:


src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAdherenceRecordDao.java
```diff
@@ -70,7 +70,9 @@ protected QueryBuilder createQuery(AdherenceRecordsSearch search) {
         builder.append(BASE_QUERY);
         
         WhereClauseBuilder where = builder.startWhere(AND);
-        where.appendRequired("ar.userId = :userId", "userId", search.getUserId());
+        if (search.getUserId() != null) {
```
jnelsonjava, line 73: Might be worth adding a test in HibernateAdherenceRecordDaoTest for a null userId.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAdherenceRecordDao.java
```diff
@@ -91,6 +93,8 @@ protected QueryBuilder createQuery(AdherenceRecordsSearch search) {
             where.append("tm.timeWindowGuid IN :timeWindowGuids",
                     "timeWindowGuids", search.getTimeWindowGuids());
         }
+        where.appendBoolean("declined", search.isDeclined());
```
jnelsonjava, line 96: This could use a test as well.

jnelsonjava (approved):
Approved. Could add unit tests referenced in comments. But I see that the integration tests seem to cover that as well.

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAdherenceRecordDao.java
alxdarksage: Added

src/main/java/org/sagebionetworks/bridge/hibernate/HibernateAdherenceRecordDao.java
alxdarksage: Added

PR #595, opened 2022-05-12
Preview accounts (BRIDGE-3255)

alxdarksage:
1) Accounts can be tagged with "preview_user" data group (if you want to use this, it needs to be added to the app's configuration where it will be used);
2) These accounts will be deleted if the study transitions from design to recruitment (no other transition, and you aren't prevented from creating these when the study is in other states);
3) An account marked with "preview_user" can only be enrolled in one study (so when it is deleted, it doesn't disappear from another study).

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -7,6 +7,8 @@ import static org.sagebionetworks.bridge.AuthEvaluatorField.USER_ID;
 import static org.sagebionetworks.bridge.AuthUtils.CAN_READ_PARTICIPANTS;
 import static org.sagebionetworks.bridge.AuthUtils.CANNOT_ACCESS_PARTICIPANTS;
 import static org.sagebionetworks.bridge.AuthUtils.canAccessAccount;
+import static org.sagebionetworks.bridge.BridgeConstants.API_MAXIMUM_PAGE_SIZE;
+import static org.sagebionetworks.bridge.BridgeConstants.PREVIEW_USER_GROUP;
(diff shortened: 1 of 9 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -286,6 +289,34 @@ public void deleteAccount(AccountId accountId) {
...
+        do {
+            searchBuilder.withOffsetBy(offset);
+            page = getPagedAccountSummaries(appId, searchBuilder.build());
+            for (AccountSummary summary : page.getItems()) {
+                // It is too slow to use deleteAccount because it cleans up a ton of
+                // DynamoDB resources. So... we leave all the non-relational data behind.
+                accountDao.deleteAccount(summary.getId());
+            }
+            offset += API_MAXIMUM_PAGE_SIZE;
+        } while(offset < page.getTotal());
```
jnelsonjava: I'm not sure about this, but is it possible to miss preview accounts if there are more to delete than the API_MAXIMUM_PAGE_SIZE? For instance, if the max page size is 100 and there are 125 preview accounts, it would delete 100 and set the offset to 100. Then the next time through the getPagedAccountSummaries() call would have an offset of 100, bypassing the remaining 25 records since the first 100 were already deleted.

jnelsonjava (approved):
Approved but check out the comment on AccountService.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
alxdarksage: GOOD CATCH. All we have to do here is keep calling until the list comes back empty, from offset 0 each time. I had initially been collecting all the IDs, then switched late in the writing of the code!

PR #599, opened 2022-05-19
BRIDGE-3307 Add SQS and SNS Initializers

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3307

Depends on https://github.com/Sage-Bionetworks/bridge-base/pull/92

Since these changes don't impact integ tests, I tested these changes manually.

PR #600, opened 2022-06-11
Get queue URLs from configs dynamically

DwayneJengSage:
Build break due to https://github.com/Sage-Bionetworks/BridgeServer2/pull/599

We changed how we initialize SQS queues. However, SqsInitializer runs after Spring, which caused some classes to pull the queue URL from the configs before SqsInitializer initialized those configs.

The fix is to pull SQS Queue URLs dynamically.


DwayneJengSage:
Merging this to un-break the build. Will have Jack review this retroactively.

jnelsonjava (commented):
Looks good.

PR #604, opened 2022-06-29
Validate MD5 file upload hash

imranmaj:
[https://sagebionetworks.jira.com/browse/BRIDGE-1897](https://sagebionetworks.jira.com/browse/BRIDGE-1897)

- Ensures that MD5 hashes associated with file uploads are valid (not that the hash necessarily matches the file) by checking base64 decoded byte length

Also included:

- Adds missing tests for cases where the MD5 hash is missing
- Adds failing for all tests when the validator fails to recognize invalid uploads (previously, all tests would pass silently even if the validator incorrectly failed to detect anything)


jnelsonjava (approved):
Looks good!

PR #608, opened 2022-07-20
BRIDGE-3150 Backfill Participant Versions

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3150

Create API to backfill a participant version. If a participant version doesn't exist, but should, this API forces the creation of a participant version. (Does nothing if the participant version already exists, or if Exporter 3.0 is not enabled.)


jnelsonjava (approved):
This looks great!

PR #609, opened 2022-07-20
Fix permissions for emailRoster for study

DwayneJengSage:
Bridge Study Manager only uses /v5/studies/{studyId}/participants/emailRoster. However, there is a bug in the permissions check that app-scoped accounts (even superadmins) cannot get the participant roster through this API.

This change adds ADMIN to CAN_EXPORT_PARTICIPANTS, so app admins (and superadmins) can get the participant roster. There is a subsequent check CAN_DOWNLOAD_PARTICIPANT_ROSTER, which was coded correctly.

Workaround is to log in using HTTP and call /v3/participants/emailRoster directly, which does not use CAN_EXPORT_PARTICIPANTS and uses CAN_DOWNLOAD_PARTICIPANT_ROSTER.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -173,11 +173,12 @@ public class AuthUtils {
 
     public static final AuthEvaluator CAN_DELETE_PARTICIPANTS = new AuthEvaluator()
             .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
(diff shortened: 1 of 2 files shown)


DwayneJengSage:
Consider instead replacing Email Roster permissions with CAN_READ_PARTICIPANTS

jnelsonjava:
I thought about it some more and you're probably right to keep this as a separate permission check. CAN_READ_PARTICIPANTS includes access for developer and study designer roles. But those roles are restricted to test users elsewhere. So the validator is a bit deceiving about current state.

jnelsonjava (approved):

PR #610, opened 2022-07-20
Updating permissions on role change (BRIDGE-3231)

jnelsonjava:
- Permissions are updated to match edits to roles on admin accounts
- Mapping to convert roles to equivalent permissions

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -74,8 +74,6 @@ public class AccountService {
     private UploadService uploadService;
     @Autowired
     private RequestInfoService requestInfoService;
-    @Autowired
-    private PermissionService permissionService;
     
     // Provided to override in tests
     protected String generateGUID() {
@@ -105,11 +103,6 @@ public class AccountService {
 
         // Create account. We don't verify studies because this is handled by validation
         accountDao.createAccount(account);
-        
-        // If roles are provided then permissions also need to be created
-        if (!account.getRoles().isEmpty()) {
-            permissionService.updatePermissionsFromRoles(account, Account.create());
-        }
         
         if (!account.getEnrollments().isEmpty()) {
             activityEventService.publishEnrollmentEvent(
@@ -169,11 +162,6 @@ public class AccountService {
 
         // Update. We don't verify studies because this is handled by validation
         accountDao.updateAccount(account);
-        
-        // If roles have changed, then permissions need to be adjusted as well
-        if (!persistedAccount.getRoles().equals(account.getRoles())) {
-            permissionService.updatePermissionsFromRoles(account, persistedAccount);
-        }
         
         // If any enrollments have been added, then create an enrollment event for that enrollment.
         // We want to create these events only after we're sure the account has been updated to 
--- a/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AccountService.java
@@ -74,6 +74,8 @@ public class AccountService {
     private UploadService uploadService;
     @Autowired
     private RequestInfoService requestInfoService;
+    @Autowired
+    private PermissionService permissionService;
     
     // Provided to override in tests
     protected String generateGUID() {
@@ -103,6 +105,11 @@ public class AccountService {
 
         // Create account. We don't verify studies because this is handled by validation
         accountDao.createAccount(account);
+        
+        // If roles are provided then permissions also need to be created
+        if (!account.getRoles().isEmpty()) {
+            permissionService.updatePermissionsFromRoles(account, Account.create());
+        }
         
         if (!account.getEnrollments().isEmpty()) {
             activityEventService.publishEnrollmentEvent(
@@ -162,6 +169,11 @@ public class AccountService {
 
         // Update. We don't verify studies because this is handled by validation
         accountDao.updateAccount(account);
+        
+        // If roles have changed, then permissions need to be adjusted as well
+        if (!persistedAccount.getRoles().equals(account.getRoles())) {
+            permissionService.updatePermissionsFromRoles(account, persistedAccount);
+        }
         
         // If any enrollments have been added, then create an enrollment event for that enrollment.
         // We want to create these events only after we're sure the account has been updated to 
--- a/src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
@@ -174,10 +174,13 @@ public class PermissionService {
     
     public void updatePermissionsFromRoles(Account account, Account persistedAccount) {
         checkNotNull(account);
-        checkNotNull(persistedAccount);
         
         Map<EntityType, Map<String, Set<AccessLevel>>> newPermissions = getPermissionsForRoles(account);
-        Map<EntityType, Map<String, Set<AccessLevel>>> oldPermissions = getPermissionsForRoles(persistedAccount);
+        Map<EntityType, Map<String, Set<AccessLevel>>> oldPermissions = new HashMap<>();
+        
+        if (persistedAccount != null) {
+            oldPermissions = getPermissionsForRoles(persistedAccount);
+        }
         
         Map<EntityType, Map<String, Set<AccessLevel>>> additionalPermissions = new HashMap<>();
         Map<EntityType, Map<String, Set<AccessLevel>>> removablePermissions = new HashMap<>();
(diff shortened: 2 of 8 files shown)


src/main/java/org/sagebionetworks/bridge/services/AccountService.java
```diff
@@ -104,6 +106,11 @@ public void createAccount(App app, Account account) {
         // Create account. We don't verify studies because this is handled by validation
         accountDao.createAccount(account);
         
+        // If roles are provided then permissions also need to be created
```
jnelsonjava: Can remove, only participant accounts are managed through here.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                            }
+                        }
+                        Map<String, Set<AccessLevel>> entityTypeMap = additionalPermissions.getOrDefault(entityType, new HashMap<>());
+                        entityTypeMap.put(entityId, additionalAccessLevels);
+                        additionalPermissions.put(entityType, entityTypeMap);
+                    } else {
+                        Map<String, Set<AccessLevel>> entityTypeMap = additionalPermissions.getOrDefault(entityType, new HashMap<>());
+                        entityTypeMap.put(entityId, newPermissionEntityTypeMap.get(entityId));
+                        additionalPermissions.put(entityType, entityTypeMap);
+                    }
```
DwayneJengSage: Is this else block covered by unit tests? (You can check by looking at the Jacoco report in target/site/jacoco/index.html.)

It looks like this block is hit if the account already has permissions to the EntityType, but not the specific Entity. However, this block only manages permissions set by Roles. Role changes can't grant permissions to new Entities, and updateAccount() forbids changing Org ID. I'm not sure if there's any way for this block to actually be hit.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+        }
+    }
+    
+    private Map<EntityType, Map<String, Set<AccessLevel>>> getPermissionsForRoles(Account account) {
+        String orgId = account.getOrgMembership();
+        if (orgId == null) {
+            return new HashMap<>();
+        }
+        
+        Set<String> studyIds = sponsorService.getSponsoredStudyIds(account.getAppId(), orgId);
```
DwayneJengSage, line 274: Edge case: If a new study is created for the org, will existing accounts be granted permissions to that study? This is probably a separate changelist, but it's something we'll need to think about, and we'll need to handle it in a performant way.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                }
+            } else {
+                // the old permissions do not have any of the new so add them all
+                additionalPermissions.put(entityType, newPermissions.get(entityType));
+            }
+        }
+        
+        for (EntityType entityType : oldPermissions.keySet()) {
+            if (newPermissions.containsKey(entityType)) {
+                Map<String, Set<AccessLevel>> oldPermissionEntityTypeMap = newPermissions.get(entityType);
```
DwayneJengSage: This is labeled oldPermissions, but it gets it from the newPermissions map. Do the unit tests catch this case, and if not, does that mean we need to add another test case?

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                            }
+                        }
+                        Map<String, Set<AccessLevel>> entityTypeMap = removablePermissions.getOrDefault(entityType, new HashMap<>());
+                        entityTypeMap.put(entityId, removableAccessLevels);
+                        removablePermissions.put(entityType, entityTypeMap);
+                    } else {
+                        Map<String, Set<AccessLevel>> entityTypeMap = removablePermissions.getOrDefault(entityType, new HashMap<>());
+                        entityTypeMap.put(entityId, oldPermissionEntityTypeMap.get(entityId));
+                        removablePermissions.put(entityType, entityTypeMap);
+                    }
```
DwayneJengSage: Similarly, what would hit this code block?

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                Map<String, Set<AccessLevel>> oldPermissionEntityTypeMap = newPermissions.get(entityType);
+                for (String entityId : oldPermissionEntityTypeMap.keySet()) {
+                    if (newPermissions.get(entityType).containsKey(entityId)) {
+                        Set<AccessLevel> newAccessLevels = newPermissions.get(entityType).get(entityId);
+                        Set<AccessLevel> removableAccessLevels = new HashSet<>();
+                        for (AccessLevel accessLevel : oldPermissionEntityTypeMap.get(entityId)) {
+                            if (!newAccessLevels.contains(accessLevel)) {
+                                removableAccessLevels.add(accessLevel);
+                            }
+                        }
```
DwayneJengSage: We are missing unit tests where the new permission type has the same entity type and entity, but removes permissions from that entity (eg, an entity going from ListThroughDelete to ListRead, which removes Edit and Delete).

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                }
+            } else {
+                removablePermissions.put(entityType, oldPermissions.get(entityType));
+            }
+        }
+        
+        List<Permission> currentPermissions = permissionDao.getPermissionsForUser(account.getAppId(), account.getId());
+        for (EntityType entityType : removablePermissions.keySet()) {
+            for (String entityId : removablePermissions.get(entityType).keySet()) {
+                for (AccessLevel accessLevel : removablePermissions.get(entityType).get(entityId)) {
```
DwayneJengSage: It might be wortwhile storing removablePermissions.get(entityType) in a temporary variable, so that you only have to call get() once. (Similarly in the other nested loops as well.)

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                    for (Permission permission : currentPermissions) {
```
DwayneJengSage: Looping through all existing permissions for each entity type, id, and access level is very inefficient. It might be simplier to loop through permissions as the outermost loop, and then getting the entity type, id, and access level from the permission and calling the getters in the map. This way, we don't need nested loops to do this operation.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
```diff
@@ -158,6 +172,120 @@ public void deletePermission(String appId, String guid) {
...
+                }
+            }
+        }
+    }
+    
+    private Map<EntityType, Map<String, Set<AccessLevel>>> getPermissionsForRoles(Account account) {
+        String orgId = account.getOrgMembership();
+        if (orgId == null) {
+            return new HashMap<>();
+        }
```
DwayneJengSage, line 272: If orgId is null, do we still call getSponsoredStudyIds()? What happens if we pass in a null orgId to that call?

Do we need a unit test for this case? Are there any permissions that get generated or changed if the account is not in an org?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: I didn't run coverage! I'm doing that now. And you're right that this can't be hit. I'm counting on the orgId staying the same in other places. So I'll remove this block.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Same as above, can't hit it. Removing.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: This should be getting it from oldPermissions. I edited the unit test to make sure there's coverage.

src/main/java/org/sagebionetworks/bridge/services/AccountService.java
jnelsonjava: Removed

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Yep, adjusted the main delete test to go from ADMIN to ORG_ADMIN to cover this as well.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: If we pass a null orgId to getSponsoredStudyIds(), it will throw an exception. If an account is missing an orgId, then this will prevent getSponsoredStudyIds() from being called and just return an empty map of permissions. Since all the permissions granted through roles are in the context of the account's org, there would be no permissions anyways.

A new account can't be missing an orgId because it's enforced in the service. And it's not possible to update the orgId of an existing account. So the only case this covers is if an old account is missing an org and goes through a role update. I'm not sure that any accounts like this exist but it seems safer to include the check for it. There is a unit test to make sure no permissions are updated and the sponsor service isn't called now.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: This is a really good point. I'll add a JIRA to track that.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: That makes sense. Updated to loop through existing permissions and delete where there is a match.

src/main/java/org/sagebionetworks/bridge/services/PermissionService.java
jnelsonjava: Updated.

DwayneJengSage (approved):

PR #612, opened 2022-08-16
Add support for demographics

imranmaj:
[https://sagebionetworks.jira.com/browse/BRIDGE-2964](https://sagebionetworks.jira.com/browse/BRIDGE-2964)


src/main/java/org/sagebionetworks/bridge/json/DemographicUserAssessmentDeserializer.java
```diff
@@ -0,0 +1,96 @@
...
+                }
+                if (resultStep.getValue() != null) {
+                    // remove null
+                    for (ListIterator<DemographicValue> iter = resultStep.getValue().listIterator(); iter.hasNext();) {
+                        DemographicValue next = iter.next();
+                        if (next == null || next.getValue() == null) {
+                            iter.remove();
+                        }
+                    }
+                }
```
DwayneJengSage: We should add test cases for the following:

* null step
* null DemographicValue
* null value in DemographicValue (is this actually possible?_

src/main/java/org/sagebionetworks/bridge/models/studies/Demographic.java
```diff
@@ -0,0 +1,129 @@
...
+    public void setValues(List<DemographicValue> values) {
+        if (values != null) {
+            // remove null
+            for (ListIterator<DemographicValue> iter = values.listIterator(); iter.hasNext();) {
+                DemographicValue next = iter.next();
+                if (next == null || next.getValue() == null) {
+                    iter.remove();
+                }
+            }
+        }
```
DwayneJengSage, line 112: Should add tests for

* null List
* null DemographicValue
* null value in DemographicValue

src/main/java/org/sagebionetworks/bridge/models/studies/DemographicUser.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @JsonProperty("demographics")
+    public void setDemographics(Map<String, Demographic> demographics) {
+        if (demographics != null) {
+            for (Map.Entry<String, Demographic> entry : demographics.entrySet()) {
+                if (entry.getValue() != null) {
+                    entry.getValue().setCategoryName(entry.getKey());
+                }
+            }
+        }
```
DwayneJengSage, line 104: Add test cases for:

* null map
* null value in map
* value has different category name than key

src/test/java/org/sagebionetworks/bridge/json/DemographicUserSerializerDeserializerTest.java
```diff
@@ -0,0 +1,166 @@
...
+import com.google.common.collect.ImmutableList;
+
+public class DemographicUserSerializerDeserializerTest {
+    /**
+     * Tests whether a DemographicUser is serialized correctly.
+     */
+    @Test
+    public void serialize() throws JsonProcessingException {
+        DemographicUser demographicUser = new DemographicUser("id1", "appid1", null, "userid1",
+                new HashMap<>());
```
DwayneJengSage: Add studyId to verify that it is ignored by the serializer.

src/test/java/org/sagebionetworks/bridge/services/DemographicServiceTest.java
```diff
@@ -0,0 +1,263 @@
...
+        when(demographicDao.saveDemographicUser(any(), any(), any(), any()))
+                .thenAnswer((invocation) -> invocation.getArgument(0));
+
+        DemographicUser returnedDemographicUser = demographicService.saveDemographicUser(demographicUser);
+
+        verify(demographicDao).saveDemographicUser(demographicUser, TEST_APP_ID, TEST_STUDY_ID, TEST_USER_ID);
+        assertEquals(returnedDemographicUser.getId(), "0");
+        Iterator<Demographic> iter = returnedDemographicUser.getDemographics().values().iterator();
+        for (int i = 1; iter.hasNext(); i++) {
+            assertEquals(Integer.toString(i), iter.next().getId());
```
DwayneJengSage: Also verify the demographics.getDemographicsUser()

src/test/java/org/sagebionetworks/bridge/services/DemographicServiceTest.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+        assertSame(returnedDemographicUser, demographicUser);
+    }
+
+    /**
+     * Tests overwriting a DemographicUser.
+     */
+    @Test
+    public void saveDemographicUserOverwrite() {
```
DwayneJengSage: This test appears identical to the previous one. Do we need them both?

src/main/java/org/sagebionetworks/bridge/spring/controllers/DemographicController.java
```diff
@@ -0,0 +1,353 @@
...
+     * @throws ConsentRequiredException    if called by the user themself OR at an
+     *                                     app level and not by the user themself
+     *                                     but the caller is not consented.
+     * @throws UnsupportedVersionException if the caller's app version is not
+     *                                     supported.
+     */
+    @PostMapping({ "/v5/studies/{studyId}/participants/{userId}/demographics",
+            "/v5/studies/{studyId}/participants/self/demographics",
+            "/v1/apps/self/participants/{userId}/demographics",
+            "/v1/apps/self/participants/self/demographics" })
```
DwayneJengSage: We should change these to "/v3/participants/{userId}/demographics" and "/v3/participants/self/demographics" to be consistent with our other participant APIs.

Similarly with the other app/self APIs in this controller.

src/main/java/org/sagebionetworks/bridge/spring/controllers/DemographicController.java
```diff
@@ -0,0 +1,353 @@
...
+
+        UserSession session;
+        String userIdUnwrapped;
+        if (userId.isPresent() && studyId.isPresent()) {
+            // posted on the user's behalf at a study level
+            session = getAuthenticatedSession(Roles.RESEARCHER, Roles.STUDY_COORDINATOR);
+            userIdUnwrapped = userId.get();
+        } else if (userId.isPresent() && !studyId.isPresent()) {
+            // posted on the user's behalf at an app level
+            session = getAdministrativeSession();
```
DwayneJengSage: getAdministrativeSession() actually means any account with a role. I think you actually want getAuthenticatedSession(Roles.ADMIN)

Similarly in the other methods in this controller.

DwayneJengSage (changes requested):
Mostly looks good. A few comments about tests and a few comments about Controller methods.

src/main/java/org/sagebionetworks/bridge/models/studies/Demographic.java
imranmaj: > null value in DemographicValue

For this format I don't have a custom deserializer; Jackson constructs this type automatically, and if a value is null then it should make a null DemographicValue (which I added a test for) instead of a DemographicValue containing null. Since it isn't constructed explicitly anywhere and I don't think there's any possible input that can result in this, I couldn't think of a way to test this case directly, so I added the test on this Demographic.setValues() method itself.

src/main/java/org/sagebionetworks/bridge/json/DemographicUserAssessmentDeserializer.java
imranmaj: > null value in DemographicValue (is this actually possible?)

I don't think so but I added a test for it anyway.

src/main/java/org/sagebionetworks/bridge/models/studies/DemographicUser.java
imranmaj: > value has different category name than key

I added a check to make sure this is updating correctly, but just to clarify I believe at the point you've highlighted the category name is always null, so it will always be different before this method runs.

src/main/java/org/sagebionetworks/bridge/spring/controllers/DemographicController.java
imranmaj: How about for fetching all app-level user demographics within an app, (currently "/v1/apps/self/participants/demographics"), does "/v3/participants/demographics" work?

DwayneJengSage (approved):

DwayneJengSage (approved):

PR #624, opened 2022-09-23
BRIDGE-3327 deleting apps and accounts should delete all associated data

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3327

These changes are needed to delete study data to reduce the amount of manual work required.

There's not really a way to integ test these. I did, however, run the integration tests to make sure nothing breaks.

PR #625, opened 2022-09-27
[BUILD BREAK] Give Org Admin permissions to delete participant reports

DwayneJengSage:
This fixes a build break.

Apparently, org admins have permission to delete participants, but not participant reports. Since we added Participant Reports to be deleted when accounts are deleted (see https://github.com/Sage-Bionetworks/BridgeServer2/pull/624), this caused a build break in Dev.

Currently, only Study Coordinators and Study Designers can read/write/delete Participant Reports. Since Org Admins can grant the Study Coordinator and Study Designer roles (including to themselves), this doesn't add anything Org Admins can't already do.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
+++ b/src/main/java/org/sagebionetworks/bridge/AuthUtils.java
@@ -139,12 +139,10 @@ public class AuthUtils {
     public static final AuthEvaluator CAN_DOWNLOAD_PARTICIPANT_ROSTER = new AuthEvaluator()
             .canAccessStudy().hasAnyRole(STUDY_COORDINATOR).or()
             .hasAnyRole(RESEARCHER, ADMIN);
-    
-    /**
-     * Can the caller read participant reports? 
-     */
+
+    /** Can the caller read/write/delete participant reports? */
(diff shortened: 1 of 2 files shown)


DwayneJengSage:
Merging to unblock build break.

imranmaj:
Would it be a good idea to add study coordinator and study designer to org admin's `PASSES_AS_ROLE` in `Roles`? It sounds like it would be allowed since org admins can grant themselves study coordinator and study designer anyway. Then I think this might work without explicitly stating `ORG_ADMIN` here because `hasAnyRole` internally checks `PASSES_AS_ROLE`.

DwayneJengSage:
> Would it be a good idea to add study coordinator and study designer to org admin's `PASSES_AS_ROLE` in `Roles`? It sounds like it would be allowed since org admins can grant themselves study coordinator and study designer anyway. Then I think this might work without explicitly stating `ORG_ADMIN` here because `hasAnyRole` internally checks `PASSES_AS_ROLE`.

It makes sense to do it, since Org Admin can grant Study Coordinator and Study Designer roles. However, that would affect all Study Coordinator and Study Designer APIs, and since this is being fast-tracked into Prod, I wanted to make the change with the bare minimum amount of impact.

PR #626, opened 2022-09-27
Add support for exporting demographics with participant versions

imranmaj:
https://sagebionetworks.jira.com/browse/BRIDGE-3338

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
@@ -15,9 +15,8 @@ public final class Exporter3Configuration {
 
     /** Helper method that returns true if all configuration attributes are specified. */
     public boolean isConfigured() {
-        return dataAccessTeamId != null && participantVersionTableId != null
-                && participantVersionDemographicsTableId != null && participantVersionDemographicsViewId != null
-                && projectId != null && rawDataFolderId != null && storageLocationId != null;
+        return dataAccessTeamId != null && participantVersionTableId != null && projectId != null &&
+                rawDataFolderId != null && storageLocationId != null;
     }
 
     /** SNS topic to publish to when a study is initialized in this app. (This is not used for study configs.) */
--- a/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
@@ -8,6 +8,7 @@ public final class Exporter3Configuration {
     private Long dataAccessTeamId;
     private String participantVersionTableId;
     private String participantVersionDemographicsTableId;
+    private String participantVersionDemographicsViewId;
     private String projectId;
     private String rawDataFolderId;
     private Long storageLocationId;
@@ -48,13 +49,22 @@ public final class Exporter3Configuration {
         this.participantVersionTableId = participantVersionTableId;
     }
 
-    /** The Synapse table where Participant Version demographics information is stored. */
+    /** The Synapse table to where we exporter Participant Version Demographics. */
     public String getParticipantVersionDemographicsTableId() {
         return participantVersionDemographicsTableId;
     }
 
     public void setParticipantVersionDemographicsTableId(String participantVersionDemographicsTableId) {
         this.participantVersionDemographicsTableId = participantVersionDemographicsTableId;
+    }
+
+    /** The Synapse materialized view which joins the Participant Versions table and the Demographics table. */
+    public String getParticipantVersionDemographicsViewId() {
+        return participantVersionDemographicsViewId;
+    }
+
+    public void setParticipantVersionDemographicsViewId(String participantVersionDemographicsViewId) {
+        this.participantVersionDemographicsViewId = participantVersionDemographicsViewId;
     }
 
     /**
--- a/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
@@ -1,7 +1,5 @@
 package org.sagebionetworks.bridge.models.apps;
 
-import java.util.Objects;
-
 /** This class holds Exporter 3.0 configuration for a given app. */
 public final class Exporter3Configuration {
     private String createStudyNotificationTopicArn;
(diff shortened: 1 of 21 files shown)


src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
```diff
@@ -1,20 +1,21 @@
...
     private String rawDataFolderId;
     private Long storageLocationId;
 
     /** Helper method that returns true if all configuration attributes are specified. */
     public boolean isConfigured() {
-        return dataAccessTeamId != null && participantVersionTableId != null && projectId != null &&
-                rawDataFolderId != null && storageLocationId != null;
+        return dataAccessTeamId != null && participantVersionTableId != null
+                && participantVersionDemographicsTableId != null && participantVersionDemographicsViewId != null
+                && projectId != null && rawDataFolderId != null && storageLocationId != null;
```
DwayneJengSage: Thinking about this some more, now that we have several studies using Exporter 3.0, this change would disable Exporter for all those studies.

A better change would be to leave this alone, but add a check in the Worker so that if these attributes don't exist, we simply don't export Demographics.

src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
```diff
@@ -80,33 +99,82 @@ public void setStorageLocationId(Long storageLocationId) {
...
+        result = prime * result + ((dataAccessTeamId == null) ? 0 : dataAccessTeamId.hashCode());
+        result = prime * result + ((participantVersionDemographicsTableId == null) ? 0
+                : participantVersionDemographicsTableId.hashCode());
+        result = prime * result + ((participantVersionDemographicsViewId == null) ? 0
+                : participantVersionDemographicsViewId.hashCode());
+        result = prime * result + ((participantVersionTableId == null) ? 0 : participantVersionTableId.hashCode());
+        result = prime * result + ((projectId == null) ? 0 : projectId.hashCode());
+        result = prime * result + ((rawDataFolderId == null) ? 0 : rawDataFolderId.hashCode());
+        result = prime * result + ((storageLocationId == null) ? 0 : storageLocationId.hashCode());
+        return result;
```
DwayneJengSage: By convention, we use "java.util.Objects.equals and hashCode (java 7+)" format to generate equals() and hashCode(). The end result is functionally the same, but staying consistent means that we have less code churn when we regenerate the equals() and hashCode() functions.

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
```diff
@@ -167,6 +187,52 @@ public class Exporter3Service {
...
+        ColumnModel participantVersionColumn = new ColumnModel();
+        participantVersionColumn.setName("participantVersion");
+        participantVersionColumn.setColumnType(ColumnType.INTEGER);
+        listBuilder.add(participantVersionColumn);
+
+        ColumnModel appIdColumn = new ColumnModel();
+        appIdColumn.setName("appId");
+        appIdColumn.setColumnType(ColumnType.STRING);
+        healthCodeColumn.setMaximumSize(255L);
+        listBuilder.add(appIdColumn);
```
DwayneJengSage: This column is not necessary, since participants will never be in more than one app, nor will any Synapse project have data from more than one app.

src/main/java/org/sagebionetworks/bridge/services/Exporter3Service.java
```diff
@@ -167,6 +187,52 @@ public class Exporter3Service {
...
+        ColumnModel studyIdColumn = new ColumnModel();
+        studyIdColumn.setName("studyId");
+        studyIdColumn.setColumnType(ColumnType.STRING);
+        healthCodeColumn.setMaximumSize(255L);
+        listBuilder.add(studyIdColumn);
+
+        ColumnModel categoryNameColumn = new ColumnModel();
+        categoryNameColumn.setName("demographicCategoryName");
+        categoryNameColumn.setColumnType(ColumnType.STRING);
+        healthCodeColumn.setMaximumSize(768L);
```
DwayneJengSage: Category name had to be limited to 255 characters in the database, due to MySQL 5.6 limitations, so this column will never be longer than 255 characters.

This is important, because table width in Synapse is a valuable resource that needs to be conserved.

src/test/java/org/sagebionetworks/bridge/services/Exporter3ServiceTest.java
```diff
@@ -524,7 +567,7 @@ private void verifyEx3ConfigAndSynapse(Exporter3Configuration ex3Config, String
 
         // Verify created project. Note that we call this method again later, which is why we verify it twice now.
         ArgumentCaptor<Entity> entitiesToCreateCaptor = ArgumentCaptor.forClass(Project.class);
-        verify(mockSynapseHelper, times(2)).createEntityWithRetry(entitiesToCreateCaptor
+        verify(mockSynapseHelper, times(3)).createEntityWithRetry(entitiesToCreateCaptor
```
DwayneJengSage, line 572: Further up in verifyEx3ConfigAndSynapse(), you'll want to verify that ex3Config was populated with the new attributes.

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
```diff
@@ -163,7 +192,7 @@ void createParticipantVersion(ParticipantVersion participantVersion) {
     // Compares non-key attributes for participant versions. Returns true if they are the same, false if they are
     // different.
     // Package-scoped for unit tests.
-    static boolean isIdenticalParticipantVersion(ParticipantVersion oldVersion, ParticipantVersion newVersion) {
+    boolean isIdenticalParticipantVersion(ParticipantVersion oldVersion, ParticipantVersion newVersion) {
```
DwayneJengSage: Is there a reason for this becoming non-static?

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/services/ParticipantVersionService.java
imranmaj: I had some changes before which I deleted, thanks for the catch

src/main/java/org/sagebionetworks/bridge/models/apps/Exporter3Configuration.java
imranmaj: Ah I see the issue. What if we re-run `initExporter3` from the worker in the case where Exporter3 is enabled but `Exporter3Configuration` is not configured? I think `initExporter3` was designed with this case in mind:

```
     * (b) If in the future, we add something new (like a notification queue, or a default view), we can re-run this
     * API to create the new stuff without affecting the old stuff.
```

That would also avoid the case where existing studies using Exporter3 expect their demographics to be exported, but they would not be because those Synapse resources would never be created, so the check to see if they exist would always be false, and therefore demographics wouldn't be exported.

DwayneJengSage: I'm not sure that's a good idea. As the Exporter scales up, there will be more requests coming in at the same time, for the same app, for possibly overlapping studies. Plus, there's the weird circular dependency between the Server, the Exporter, and the Server again.

It would be simpler to just not export Demographics unless it's been configured.

DwayneJengSage (changes requested):
Looks good. Just the one comment about isConfigured.

DwayneJengSage (approved):

PR #629, opened 2022-10-08
Add ImageResource to Assessments

imranmaj:
https://sagebionetworks.jira.com/browse/BMC-384

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/hibernate/LabelConverter.java
+++ b/src/main/java/org/sagebionetworks/bridge/hibernate/LabelConverter.java
@@ -1,0 +1,24 @@
+package org.sagebionetworks.bridge.hibernate;
+
+import javax.persistence.Converter;
+
+import org.sagebionetworks.bridge.models.Label;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+
+@Converter
+public class LabelConverter extends BaseJsonAttributeConverter<Label> {
+
+    private static final TypeReference<Label> TYPE_REF = new TypeReference<Label>() {
+    };
+
+    @Override
+    public String convertToDatabaseColumn(Label list) {
+        return super.serialize(list);
+    }
+
+    @Override
+    public Label convertToEntityAttribute(String string) {
+        return super.deserialize(string, TYPE_REF);
+    }
+}
--- a/src/main/java/org/sagebionetworks/bridge/models/assessments/ImageResource.java
+++ b/src/main/java/org/sagebionetworks/bridge/models/assessments/ImageResource.java
@@ -1,9 +1,11 @@
 package org.sagebionetworks.bridge.models.assessments;
 
+import java.util.List;
+
 import javax.persistence.Convert;
 import javax.persistence.Embeddable;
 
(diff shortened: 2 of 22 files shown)


src/main/java/org/sagebionetworks/bridge/models/assessments/ImageResource.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.persistence.Embeddable;
+import javax.persistence.Table;
+
+import org.sagebionetworks.bridge.hibernate.LabelConverter;
+import org.sagebionetworks.bridge.json.BridgeTypeName;
+import org.sagebionetworks.bridge.models.BridgeEntity;
+import org.sagebionetworks.bridge.models.Label;
+
+@Embeddable
+@Table(name = "ImageResources")
```
imranmaj: remove

syoung-smallwisdom:
@imranmaj I messed up and the label needs to be an array with a default value in English so that we can support other languages with the same icon. In the client apps, the label is used for accessibility.

src/main/java/org/sagebionetworks/bridge/hibernate/LabelConverter.java
```diff
@@ -0,0 +1,24 @@
+package org.sagebionetworks.bridge.hibernate;
+
+import javax.persistence.Converter;
+
+import org.sagebionetworks.bridge.models.Label;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+
+@Converter
+public class LabelConverter extends BaseJsonAttributeConverter<Label> {
```
jnelsonjava, line 10: This class might not be necessary now that you're using the LabelListConverter.

imranmaj: Yeah I'm no longer using it but since it's already written and tested I thought maybe it could save someone time in the future, do you think I should just delete it anyway?

src/test/java/org/sagebionetworks/bridge/models/assessments/ImageResourceTest.java
```diff
@@ -0,0 +1,48 @@
...
+    @Test
+    public void canSerialize() throws JsonMappingException, JsonProcessingException {
+        ImageResource imageResource = new ImageResource();
+        imageResource.setName("default");
+        imageResource.setModule("sage_survey");
+        imageResource.setLabels(LABELS);
+
+        JsonNode node = BridgeObjectMapper.get().valueToTree(imageResource);
+
+        assertEquals(node.get("name").textValue(), "default");
```
jnelsonjava, line 25: Optional addition: assert the node size. It would cause the test to fail if someone adds another field but forgets to update the unit tests.

src/main/java/org/sagebionetworks/bridge/hibernate/LabelConverter.java
jnelsonjava: That's a good point. I checked with Shannon and I think you can keep it in there. Seems like it could get some use when more localization support is added in the future.

jnelsonjava (approved):
Looks good.

DwayneJengSage (approved):

PR #642, opened 2023-01-26
Add alerts

imranmaj:
See:

- https://sagebionetworks.jira.com/wiki/spaces/MTB/pages/2624421937/Alerts+Requierments
- https://sagebionetworks.jira.com/browse/MTB-651
- https://sagebionetworks.jira.com/browse/BRIDGE-3316


src/main/resources/db/changelog/changelog.sql
```diff
@@ -1040,3 +1040,19 @@ ADD INDEX `TimelineMetadata-SessionInstanceGuid` (sessionInstanceGuid);
...
+CREATE TABLE IF NOT EXISTS `Alerts` (
+    `id` varchar(60) NOT NULL,
+    `createdOn` bigint(20) NOT NULL,
+    `studyId` varchar(60) NOT NULL,
+    `appId` varchar(60) NOT NULL,
+    `userId` varchar(255) NOT NULL,
+    `category` varchar(255) NOT NULL,
+    `data` varchar(2048) NOT NULL,
+    PRIMARY KEY (`id`),
+    UNIQUE KEY (`studyId`, `appId`, `userId`, `category`),
```
DwayneJengSage: Does this create an index on app and study? If not, we need to add this index.

Same with index on app+userId and index on app+study+userId.

src/main/java/org/sagebionetworks/bridge/models/activities/ActivityEventObjectType.java
```diff
@@ -157,7 +157,7 @@ public String getEventId(String objectId, ActivityEventType eventType, String an
      */
     INSTALL_LINK_SENT(FUTURE_ONLY) {
         public String getEventId(String objectId, ActivityEventType eventType, String answerValue) {
-            return this.name().toLowerCase();
+            return TIMELINE_RETRIEVED_ID;
```
DwayneJengSage: This needs to be moved to the correct enum value.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -84,13 +85,16 @@
 public class AdherenceService {
     private static final Logger LOG = LoggerFactory.getLogger(AdherenceService.class);
     
+    private static final int WEEKLY_ADHERENCE_PERCENT_ALERT_THRESHOLD = 60;
```
DwayneJengSage: We should use the study configured value, not the hard-coded 60%.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -410,7 +419,14 @@ public WeeklyAdherenceReport getWeeklyAdherenceReport(String appId, String study
...
         
         WeeklyAdherenceReport weeklyReport = deriveWeeklyAdherenceFromStudyReportWeek(studyId, account, report);
-        
+
+        // trigger alert for low weekly adherence
+        if (weeklyReport.getWeeklyAdherencePercent() != null
+                && weeklyReport.getWeeklyAdherencePercent() <= WEEKLY_ADHERENCE_PERCENT_ALERT_THRESHOLD) {
+            alertService.createAlert(
+                    Alert.lowAdherence(studyId, appId, account.getId(), WEEKLY_ADHERENCE_PERCENT_ALERT_THRESHOLD));
+        }
```
DwayneJengSage: As per discussion, there is already a hook in the Adherence Report Worker that calculates adherence and checks against the adherence threshold. Do we want to use that?

Also, in the worker code, it uses less than, not less than or equal to. Can we confirm requirements?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AlertController.java
```diff
@@ -0,0 +1,84 @@
...
+     * @throws NotAuthenticatedException if the caller is not authenticated.
+     * @throws UnauthorizedException     if the caller is not a researcher or study
+     *                                   coordinator.
+     * @throws BadRequestException       if offsetBy or pageSize is invalid.
+     */
+    @GetMapping("/v5/studies/{studyId}/alerts")
+    public PagedResourceList<Alert> getAlerts(@PathVariable String studyId,
+            @RequestParam(required = false) String offsetBy, @RequestParam(required = false) String pageSize)
+            throws NotAuthenticatedException, UnauthorizedException, BadRequestException {
+        UserSession session = getAuthenticatedSession(Roles.RESEARCHER, Roles.STUDY_COORDINATOR);
```
DwayneJengSage, line 61: We should also check using AuthUtils.CAN_EDIT_STUDY_PARTICIPANTS (which is used for both reading and editing participants). For an example of this, see StudyParticipantController.

Same for deleteAlerts().

src/main/java/org/sagebionetworks/bridge/spring/controllers/AlertController.java
```diff
@@ -0,0 +1,84 @@
...
+     *                                   coordinator.
+     * @throws EntityNotFoundException   if the alerts to delete do not exist or are
+     *                                   not from this study.
+     */
+    @PostMapping("/v5/studies/{studyId}/deleteAlerts")
+    public StatusMessage deleteAlerts(@PathVariable String studyId)
+            throws NotAuthenticatedException, UnauthorizedException, EntityNotFoundException {
+        UserSession session = getAuthenticatedSession(Roles.RESEARCHER, Roles.STUDY_COORDINATOR);
+        List<String> alertIds = parseJson(new TypeReference<List<String>>() {
+        });
```
DwayneJengSage: To be consistent with all our other APIs, we should create a wrapper object that contains a list of alert IDs.

DwayneJengSage (changes requested):
A few comments. Will provide additional review when unit tests are ready.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
imranmaj: > As per discussion, there is already a hook in the Adherence Report Worker that calculates adherence and checks against the adherence threshold. Do we want to use that?

I changed the alert trigger to only occur when the worker calls the API, but I still kept it in the server code because it was easier for me to implement and test with what I already had, and I also didn't see a disadvantage to doing so (functionally I think both approaches are the same). If it would be better to move it to the worker I can do that though.

> Also, in the worker code, it uses less than, not less than or equal to. Can we confirm requirements?

I'll check on that

DwayneJengSage (approved):

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
imranmaj: Confirmed that less than or equal to is correct

DwayneJengSage (approved):

PR #644, opened 2023-02-07
[FIX BUILD BREAK] Remove alert from EnrollmentService.addEnrollment()

DwayneJengSage:
ParticipantService.createParticipant() will enroll the participant *before* creating the account. This means that EnrollmentService.addEnrollment() will fail to create the alert because the participant doesn't exist yet. This results in createParticipant() failing, which causes a lot of issues.

See https://sagebionetworks.jira.com/browse/MTB-651


DwayneJengSage:
I sent this to Jack and Imran for review, but I'm going ahead and merging this to un-break the build.

PR #652, opened 2023-03-29
DHP-256 Update adherence uploadedOn when upload completes

jnelsonjava:
[https://sagebionetworks.jira.com/browse/DHP-256](https://sagebionetworks.jira.com/browse/DHP-256)

Adding the earliest successful uploadedOn date for uploads related to adherence record. Existing record is updated when an upload completes or a new adherence record is created. The uploadId is also noted on the adherence record.

Functionality only exists if the upload metadata includes valid instanceGuid, eventTimestamp, and startedOn referencing the participant's timeline. Works whether time window is persistent or not.


src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceRecord.java
```diff
@@ -148,4 +154,16 @@ public String getAssessmentGuid() {
...
+    }
+    public void setUploadIds(Set<String> uploadIds) {
+        this.uploadIds = uploadIds;
+    }
+    public void addUploadId(String uploadId) {
+        if (uploadIds == null) {
+            uploadIds = new HashSet<>();
+        }
+        uploadIds.add(uploadId);
+    }
```
DwayneJengSage: If it's easy to do, we should create upload IDs as a separate table so that we can query on these easily. (There's a way to do this in Hibernate, so you don't need to create a new class, but I forget the details off the top of my head.)

If that's a big amount of work, we could serialize this as a comma-delimited list, and put a SQL index on it, which is a bit hacky, but we should still be able to query on it.

DwayneJengSage: See for example, the annotations on HibernateSharedModuleMetadata.getTags()

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -646,6 +661,118 @@ public UploadViewEx3 getUploadViewForExporter3(String appId, String studyId, Str
...
+            
+            // Must include instanceGuid, eventTimestamp, and startedOn
+            if (instanceGuidNode != null && instanceGuidNode.textValue() != null &&
+                    eventTimestampNode != null && eventTimestampNode.textValue() != null &&
+                    startedOnNode != null && startedOnNode.textValue() != null) {
+                
+                String instanceGuid = instanceGuidNode.textValue();
+                DateTime eventTimestamp;
+                try {
+                    eventTimestamp = DateTime.parse(eventTimestampNode.textValue());
```
jnelsonjava, line 683: Confirm the format that these fields are sent in with the mobile devs.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -493,6 +505,9 @@ public void uploadComplete(String appId, UploadCompletionClient completedBy, Upl
             // and duplicate records.
             return;
         }
+        
+        // Save uploadedOn date and uploadId to related adherence records.
+        updateAdherenceWithUploadInfo(appId, upload);
```
jnelsonjava: Move to end of method.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -646,6 +661,118 @@ public UploadViewEx3 getUploadViewForExporter3(String appId, String studyId, Str
...
+                        PagedResourceList<AdherenceRecord> recordList = adherenceService.getAdherenceRecords(appId,
+                                search.build());
+                        
+                        List<AdherenceRecord> recordsToUpdate = new ArrayList<>();
+                        boolean foundExistingRecord = false;
+                        
+                        for (AdherenceRecord record : recordList.getItems()) {
+                            if (timelineMetadata.isTimeWindowPersistent() && !record.getStartedOn().isEqual(startedOn)) {
+                                // If the window is persistent then the records would have to share startedOn values
+                                // to be considered the same.
```
jnelsonjava, line 730: Add a log here.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
jnelsonjava: Done

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
jnelsonjava: Done

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/AdherenceRecord.java
jnelsonjava: Thank you for the example. It's updated and working. The compound ID for adherence records is 5 fields, so the new table is unfortunately quite large for what it is.

Another option would be to add a single generated unique ID to the adherence record that's not exposed through the API. And just use it for internal joins. Let me know if that sounds better to you and I could give it a shot in the morning.

DwayneJengSage (approved):

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
jnelsonjava: Confirmed the eventTimestamp and startedOn are sent in ISO-8601 in UTC.

PR #653, opened 2023-03-29
BRIDGE-3390 ClientInfo Fixes

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-3390

This pull request contains 2 changes:

1. There is a bug where even though the ClientInfo is being captured in the uploadRequest API, we're not actually saving it to DynamoDB, so it is being overridden in the uploadComplete API. This means that if the app is updated between Request and Complete, the ClientInfo will be wrong. This is also more likely to happen in the case of redrives.

2. We only provide the parsed value of ClientInfo. We should also provide the raw user-agent header value. This way, if for whatever reason the user-agent value can’t be parsed, we still provide something rather than nothing.


jnelsonjava (approved):
Looks good.

PR #655, opened 2023-04-14
BRIDGE-3357 - Detailed adherence report for single participant

jnelsonjava:
https://sagebionetworks.jira.com/browse/BRIDGE-3357

- Adding API call for assessment level adherence reporting for individual participants.
- Returns session reports with nested assessment reports.
- Matches sort order of the participant's timeline.


src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -549,4 +559,223 @@ public AdherenceStatistics getAdherenceStatistics(String appId, String studyId,
...
+        Schedule2 schedule = scheduleService.getScheduleForStudy(appId, studyId)
+                .orElseThrow(() -> new EntityNotFoundException(Schedule2.class));
+        List<TimelineMetadata> metadata = scheduleService.getScheduleMetadata(study.getScheduleGuid());
+        
+        Timeline timeline = scheduleService.getTimelineForSchedule(appId, schedule.getGuid());
+        
+        List<StudyActivityEvent> events = studyActivityEventService.getRecentStudyActivityEvents(
+                appId, studyId, account.getId()).getItems();
+        
+        List<AdherenceRecord> adherenceRecords = getAdherenceRecords(appId, new AdherenceRecordsSearch.Builder()
```
DwayneJengSage: Note that this is a paginated API, and the default page size is 250. (The max page size is 500.) Are we simply not supporting more than 250 records at initial release? Or will we need to paginate on this API?

How many adherence records can we expect a participant to have, and how do we make sure we don't end up spinning for a very long time trying to get all the adherence records?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/detailed/DetailedAdherenceReportAssessmentRecord.java
```diff
@@ -0,0 +1,94 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.joda.time.DateTime;
+import org.sagebionetworks.bridge.json.DateTimeSerializer;
+
+public class DetailedAdherenceReportAssessmentRecord {
+    private String assessmentName;
+    private String assessmentId;
+    private String assessmentGuid;
+    private String assessmentInstanceGuid;
+    private String assessmentStatus;
```
DwayneJengSage: Should this be an enum?

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -549,4 +559,223 @@ public AdherenceStatistics getAdherenceStatistics(String appId, String studyId,
...
+        String zoneId = studyService.getZoneId(appId, studyId, account.getClientTimeZone());
+        
+        AdherenceState.Builder builder = new AdherenceState.Builder();
+        builder.withNow(createdOn);
+        builder.withClientTimeZone(zoneId);
+        
+        Study study = studyService.getStudy(appId, studyId, true);
+        if (study.getScheduleGuid() == null) {
+            throw new EntityNotFoundException(Schedule2.class);
+        }
```
DwayneJengSage: This check is already part of scheduleService. We probably don't need to include it here. (We should also update the unit test to use scheduleService instead of studyService.)

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
```diff
@@ -1383,6 +1396,340 @@ public void getAdherenceStatistics_adherenceTooHigh() {
...
+        
+        
+        DetailedAdherenceReport report = service.getDetailedAdherenceReportForParticipant(
+                TEST_APP_ID, TEST_STUDY_ID, account);
+        
+        assertEquals(report.getParticipant().getIdentifier(), TEST_USER_ID);
+        assertEquals(report.getParticipant().getExternalId(), TEST_EXTERNAL_ID);
+        assertTrue(report.isTestAccount());
+        assertEquals(report.getJoinedDate().withZone(DateTimeZone.forID("America/Los_Angeles")).toString(),
+                ENROLLMENT.toString());
```
DwayneJengSage: I think the withZone() should be on enrollment, not joinedDate. This allows us to test that joinedDate was set to the correct timezone.

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
```diff
@@ -1461,4 +1812,121 @@ private AdherenceRecordList mockRecordUpdate(AdherenceRecord rec1, AdherenceReco
         
         return new AdherenceRecordList(records);
     }
+    
+    private Account mockDetailedAdherenceTimeline(List<AdherenceRecord> testAdherenceRecords) {
+        when(mockStudyService.getZoneId(TEST_APP_ID, TEST_STUDY_ID, TEST_CLIENT_TIME_ZONE))
+                .thenReturn(TEST_CLIENT_TIME_ZONE);
```
DwayneJengSage: About half of our test timestamps are in UTC and the other half are in Pacific Time. I recommend setting the timezone to something different entirely, such as America/New_York, so that we can verify that the code sets the timezone correctly.

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
```diff
@@ -1383,6 +1396,340 @@ public void getAdherenceStatistics_adherenceTooHigh() {
...
+                TEST_APP_ID, TEST_STUDY_ID, account);
+        
+        assertEquals(report.getParticipant().getIdentifier(), TEST_USER_ID);
+        assertEquals(report.getParticipant().getExternalId(), TEST_EXTERNAL_ID);
+        assertTrue(report.isTestAccount());
+        assertEquals(report.getJoinedDate().withZone(DateTimeZone.forID("America/Los_Angeles")).toString(),
+                ENROLLMENT.toString());
+        
+        List<DetailedAdherenceReportSessionRecord> sessionRecords = report.getSessionRecords();
+        assertEquals(sessionRecords.size(), 2);
```
DwayneJengSage, line 1426: Should we also check the sort priority of the two session records and their assessment records?

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -549,4 +559,223 @@ public AdherenceStatistics getAdherenceStatistics(String appId, String studyId,
...
+                
+                Map<String, DetailedAdherenceReportAssessmentRecord> assessmentRecords = sessionRecords
+                        .get(scheduledSession.getInstanceGuid()).getAssessmentRecordMap();
+                assessmentRecords.put(assessmentRecord.getAssessmentInstanceGuid(), assessmentRecord);
+                
+            } else if (scheduledSessionMap.containsKey(instanceGuid)) {
+                ScheduledSession scheduledSession = scheduledSessionMap.get(instanceGuid);
+                
+                fillSessionRecord(sessionRecords, scheduledSession, state, sortReferenceMap, record);
+            }
```
DwayneJengSage, line 732: It looks like this is only possible if we have scheduled sessions that have no session assessments. Is that possible? Should we have a unit test for this scenario?

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
```diff
@@ -244,4 +245,19 @@ public StatusMessage deleteAdherenceRecord(
         service.deleteAdherenceRecord(record);
         return DELETED_MSG;
     }
+    
+    @GetMapping("/v5/studies/{studyId}/participants/{userId}/adherence/detail")
+    public DetailedAdherenceReport getDetailedParticipantAdherenceReport(@PathVariable String studyId,
+                                                                         @PathVariable String userId) {
+        UserSession session = getAuthenticatedSession(DEVELOPER, RESEARCHER, STUDY_DESIGNER, STUDY_COORDINATOR);
+        
+        CAN_READ_STUDIES.checkAndThrow(STUDY_ID, studyId);
```
DwayneJengSage: I think for Adherence, we usually use CAN_ACCESS_ADHERENCE_DATA. This takes in both studyId and userId (although since we never get adherence for self in this API, the userId won't actually matter).

Unrelated, it appears that the other adherence report APIs in this Controller don't check that the caller has access to the study.

DwayneJengSage (approved):
Overall looks good. Some comments. Can merge on Tuesday morning if comments are resolved. Thanks!

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
jnelsonjava: This could get over that limit, especially because it includes assessment adherence records. The other reports use only sessions so they haven't hit a need to go over a page. I updated it to collect using pagination.

I used the max page size constant though, which is set to 100. Do you think I should increase it to 250 or 500?

src/main/java/org/sagebionetworks/bridge/models/schedules2/adherence/detailed/DetailedAdherenceReportAssessmentRecord.java
jnelsonjava: That would make sense. I changed it to an enum similar to the SessionCompletionState enum.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
jnelsonjava: Updated.

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
jnelsonjava: Good catch. It wasn't failing only because I was using the America/Los_Angeles zone.

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
jnelsonjava: Updated.

src/test/java/org/sagebionetworks/bridge/services/AdherenceServiceTest.java
jnelsonjava: I added another assessment to get an assessment order check in. The sessions and bursts are both just ScheduledSessions in the timeline so their order is getting checked as well.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
jnelsonjava: I actually don't know if it's possible to have a session without assessments. I haven't seen that.

This section is comparing the instanceGuid from the adherence record to the timeline. And the adherence record could reference either an assessment instance or a session instance. This block takes the session adherence records and uses them to fill out startedOn/finishedOn information in the session report records.
The tricky part is that the adherence records can come in any order, so I couldn't guarantee that the session record would exist before its child assessments are created. So a placeholder session record is created whenever its first child assessment shows up. But that record won't have its own adherence data updated until its own related session adherence record comes through.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AdherenceController.java
jnelsonjava: Updated. It looks like the other reports are doing their auth checks in the service instead of the controller.

PR #662, opened 2023-06-12
DHP-915 Adherence Post-Processing Attributes

DwayneJengSage:
https://sagebionetworks.jira.com/browse/DHP-915

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
+++ b/src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
@@ -396,7 +396,7 @@ public class AdherenceService {
         );
 
         // Get the adherence record that we need to update.
-        AdherenceRecordsSearch search = new AdherenceRecordsSearch.Builder()
+        AdherenceRecordsSearch.Builder searchBuilder = new AdherenceRecordsSearch.Builder()
                 .withStudyId(studyId)
                 .withUserId(userId)
                 .withInstanceGuids(ImmutableSet.of(instanceGuid))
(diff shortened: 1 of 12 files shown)


src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -376,7 +378,70 @@ public void deleteAdherenceRecord(AdherenceRecord record) {
...
+        checkNotNull(attributes);
+
+        // Check permissions.
+        CAN_ACCESS_ADHERENCE_DATA.checkAndThrow(
+                AuthEvaluatorField.STUDY_ID, studyId,
+                AuthEvaluatorField.USER_ID, userId
+        );
+
+        // Get the adherence record that we need to update.
+        AdherenceRecordsSearch search = new AdherenceRecordsSearch.Builder()
```
DwayneJengSage: If we have startedOn, add it here, so we can handle Persistent tasks.

src/main/java/org/sagebionetworks/bridge/services/AdherenceService.java
```diff
@@ -376,7 +378,70 @@ public void deleteAdherenceRecord(AdherenceRecord record) {
...
+            record.setEventTimestamp(eventTimestamp);
+        } else {
+            // Otherwise, update the existing record. There should only be one record.
+            record = adherenceRecordList.get(0);
+        }
+
+        // Update the record.
+        record.setPostProcessingAttributes(attributes.getPostProcessingAttributes());
+        record.setPostProcessingCompletedOn(attributes.getPostProcessingCompletedOn());
+        record.setPostProcessingStatus(attributes.getPostProcessingStatus());
```
DwayneJengSage, line 438: Only write these if they are non-null. Merge old attributes with new attributes.

DwayneJengSage: Put this logic in service.updateAdherenceRecords() so that no one can accidentally clobber these attributes.

jnelsonjava (approved):

PR #672, opened 2023-10-11
DHP-1071 CSV Storage API

DwayneJengSage:
https://sagebionetworks.jira.com/browse/DHP-1071

API and object model for storing table rows. This is written by the Exporter3Worker to store table rows. This is read by the (currently non-existent) CSV Worker to assemble the rows into a CSV.

TODO: unit tests

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadTableController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadTableController.java
@@ -1,0 +1,96 @@
+package org.sagebionetworks.bridge.spring.controllers;
+
+import static org.sagebionetworks.bridge.Roles.SUPERADMIN;
+import static org.sagebionetworks.bridge.Roles.WORKER;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.CrossOrigin;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import org.sagebionetworks.bridge.exceptions.EntityNotFoundException;
+import org.sagebionetworks.bridge.models.PagedResourceList;
+import org.sagebionetworks.bridge.models.StatusMessage;
+import org.sagebionetworks.bridge.services.StudyService;
+import org.sagebionetworks.bridge.services.UploadTableService;
+import org.sagebionetworks.bridge.upload.UploadTableRow;
+import org.sagebionetworks.bridge.upload.UploadTableRowQuery;
+
+/** Controller for upload table rows. */
+@CrossOrigin
+@RestController
+public class UploadTableController extends BaseController {
+    static final StatusMessage CREATED_MSG = new StatusMessage("Upload table row created.");
+    static final StatusMessage DELETED_MSG = new StatusMessage("Upload table row deleted.");
+
+    private StudyService studyService;
+    private UploadTableService uploadTableService;
+
+    @Autowired
+    public final void setStudyService(StudyService studyService) {
+        this.studyService = studyService;
+    }
+
+    @Autowired
+    public final void setUploadTableService(UploadTableService uploadTableService) {
+        this.uploadTableService = uploadTableService;
+    }
+
+    /** Delete a single upload table row. This is currently only used by integration tests (so only superadmin). */
+    @DeleteMapping("/v1/apps/{appId}/studies/{studyId}/uploadtable/{recordId}")
+    public StatusMessage deleteUploadTableRowForSuperadmin(@PathVariable String appId, @PathVariable String studyId,
+            @PathVariable String recordId) {
+        getAuthenticatedSession(SUPERADMIN);
+
+        // Verify study exists.
+        studyService.getStudy(appId, studyId, true);
+
+        uploadTableService.deleteUploadTableRow(appId, studyId, recordId);
+        return DELETED_MSG;
+    }
+
+    /** Get a single upload table row. This is currently only used by integration tests (so only superadmin). */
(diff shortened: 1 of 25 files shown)


src/main/resources/db/changelog/changelog.sql
```diff
@@ -1102,3 +1102,25 @@ ADD COLUMN `postProcessingStatus` varchar(255) DEFAULT NULL;
...
 ADD COLUMN `phase` enum('DRAFT', 'REVIEW', 'PUBLISHED') DEFAULT 'DRAFT';
+
+-- changeset bridge:82
+
+CREATE TABLE IF NOT EXISTS `UploadTableRows` (
+    `appId` varchar(60) NOT NULL,
+    `studyId` varchar(60) NOT NULL,
+    `recordId` char(24) NOT NULL,
+    `assessmentId` varchar(255) NOT NULL,
+    `assessmentRevision` int(10) unsigned NOT NULL,
```
DwayneJengSage: Change to assessmentGuid

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadTableController.java
```diff
@@ -0,0 +1,96 @@
...
+        this.uploadTableService = uploadTableService;
+    }
+
+    /** Delete a single upload table row. This is currently only used by integration tests (so only superadmin). */
+    @DeleteMapping("/v1/apps/{appId}/studies/{studyId}/uploadtable/{recordId}")
+    public StatusMessage deleteUploadTableRowForSuperadmin(@PathVariable String appId, @PathVariable String studyId,
+            @PathVariable String recordId) {
+        getAuthenticatedSession(SUPERADMIN);
+
+        // Verify study exists.
```
DwayneJengSage: Add comment throwsException=true

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadTableController.java
```diff
@@ -0,0 +1,96 @@
...
+        studyService.getStudy(appId, studyId, true);
```
DwayneJengSage: Move this to the service.

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadTableController.java
```diff
@@ -0,0 +1,96 @@
...
+    @GetMapping("/v1/apps/{appId}/studies/{studyId}/uploadtable/{recordId}")
+    public UploadTableRow getUploadTableRowForSuperadmin(@PathVariable String appId, @PathVariable String studyId,
+            @PathVariable String recordId) {
+        getAuthenticatedSession(SUPERADMIN);
+
+        // Verify study exists.
+        studyService.getStudy(appId, studyId, true);
+
+        return uploadTableService.getUploadTableRow(appId, studyId, recordId).orElseThrow(
+                () -> new EntityNotFoundException(UploadTableRow.class));
```
DwayneJengSage: Move orElseThrow to the service.

src/main/resources/db/changelog/changelog.sql
syoung-smallwisdom: I believe the assessment guid includes the assessment revision.

syoung-smallwisdom (approved):
Looks good.

I think you are going to need a second table to store the json type and description for each key-value pair in both the `data` and `metadata`. That's the first thing every researcher always asks for (which is how we ended up with the requirement of including json schemas).

DwayneJengSage:
@syoung-smallwisdom 

> I think you are going to need a second table to store the json type and description for each key-value pair in both the `data` and `metadata`. That's the first thing every researcher always asks for (which is how we ended up with the requirement of including json schemas).

This is documention the documentation that I intended to store it in AssessmentResources as part of https://sagebionetworks.jira.com/browse/DHP-1070

Last I checked, we decided not to do that, and I don't think anyone's proposed an alternative yet 🤷

syoung-smallwisdom:
> @syoung-smallwisdom
> 
> > I think you are going to need a second table to store the json type and description for each key-value pair in both the `data` and `metadata`. That's the first thing every researcher always asks for (which is how we ended up with the requirement of including json schemas).
> 
> This is documention the documentation that I intended to store it in AssessmentResources as part of https://sagebionetworks.jira.com/browse/DHP-1070
> 
> Last I checked, we decided not to do that, and I don't think anyone's proposed an alternative yet 🤷

I'm pointing it out as a warning. Storing all JSON values as strings loses type information that is needed to restore the value. Since strings aren't inherently typed for int/double but most serializers are, this leads to errors. Also, we inevitably get asked "what do the values _mean_?" where what they want to know is "what was the question asked?".

In the past, we've handled this by linking to schemas defined using [json-schema.org](https://json-schema.org) schemas. I would recommend a JIRA to add a table for storing the mapping of `[key : JsonSchemaProperty]` for each assessment.

While in actuality, [JsonSchemaProperty](https://github.com/BridgeDigitalHealth/JsonModel-Swift/blob/main/Sources/JsonModel/JsonSchema.swift) is somewhat obfuscated and difficult to describe in Java or Kotlin, it could be simplified to:

```
enum class JsonType { string, integer, number, boolean, array, object }

data class JsonSchemaProperty(
    type: JsonType, 
    description: String,

    // The schema for the items in an array.
    items: ArrayItems? = null,
)

data class ArrayItems(type: JsonType, enum: List<String>? = null)
``` 

This is about the limit of what most researchers can seem to grok. Attempts to give researchers more than this have led to political intrigue, but not giving them this has lead to "This is useless and doesn't tell me what I need to know".

Further reading: [Json Schema](https://cswr.github.io/JsonSchema/)

Note: I'm describing this in Kotlin b/c that's what we're planning to use for surveys and for those, this would need to be derived from the config, but I'm using Json-schema syntax b/c that's more widely used.

DwayneJengSage:
@syoung-smallwisdom Yes, I agree that this needs to be done, but the original plan was to solve it with Assessment Resources, and it’s currently tracked by https://sagebionetworks.jira.com/browse/DHP-1070

But we decided not to do that, and I don’t know why, so I don’t know what the plan is, and I don’t know how to move forward with this 🤷

nategbrown9 (approved):
Looks good.

PR #673, opened 2023-10-24
Add Worker APIs to support DHP-1073 CSV Worker

DwayneJengSage:
https://sagebionetworks.jira.com/browse/DHP-1073

See https://github.com/Sage-Bionetworks/BridgeWorkerPlatform/pull/177

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
+++ b/src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
@@ -7,6 +7,7 @@ import static org.sagebionetworks.bridge.BridgeConstants.SHARED_APP_ID;
 import static org.sagebionetworks.bridge.Roles.ADMIN;
 import static org.sagebionetworks.bridge.Roles.DEVELOPER;
 import static org.sagebionetworks.bridge.Roles.STUDY_DESIGNER;
+import static org.sagebionetworks.bridge.Roles.WORKER;
 
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
```diff
@@ -95,7 +96,16 @@ public Assessment getAssessmentByGuid(@PathVariable String guid) {
         }
         return service.getAssessmentByGuid(appId, ownerId, guid);
     }
-    
+
+    @GetMapping("/v1/apps/{appId}/assessments/{guid}")
+    public Assessment getAssessmentByGuidForWorker(@PathVariable String appId, @PathVariable String guid) {
+        getAuthenticatedSession(WORKER);
+        if (SHARED_APP_ID.equals(appId)) {
+            throw new UnauthorizedException(SHARED_ASSESSMENTS_ERROR);
```
syoung-smallwisdom, line 108: So... the way things currently work, only shared assessments can be used by all organizations so the ARC assessments are all "shared". Won't those still need to be scored and have a CVS file created by the worker?

syoung-smallwisdom (commented):
Confused about not allowing getting a shared assessment by appId.

src/main/java/org/sagebionetworks/bridge/spring/controllers/AssessmentController.java
DwayneJengSage: The API doesn't allow shared assessments to be consistent with all the other APIs for assessments and shared assessments.

As per stand-up meeting today, the short term hack is to add fallback logic to the Worker to check for local assessments, then fall back to shared assessments. In 2024, we'll need to rework schedules and shared assessments to better fit how we actually use them.

DwayneJengSage (commented):
Add getAssessmentConfigForWorker to unblock Nathaniel's stuff.

DwayneJengSage:
@nategbrown9 This is updated with the assessment config API and unit tests.

nategbrown9 (approved):
Looks good.

syoung-smallwisdom (approved):
looks good

PR #674, opened 2023-10-25
Workaround for DIAN-749: Android inv-arc app has wrong public key

nategbrown9:


Erin-Mounts (approved):
looks "good" to me

DwayneJengSage (approved):
Looks good. As long as the Integ Test for UploadTest works, this should be fine to push to prod.

syoung-smallwisdom (approved):
It was a hack, hack, hack,
All dressed in black, black, black...

PR #675, opened 2023-10-27
Workaround for DIAN-749: Android inv-arc app has wrong public key, take 2

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/DIAN-749

Nathaniel realized that his original fix (https://github.com/Sage-Bionetworks/BridgeServer2/pull/674) didn't work because the InputStream has already been read, so the recursive call can't read the same InputStream again.

I wrote UploadArchiveServiceInvArcTest.testInvArc() to repro this bug with Nathaniel's original change. Then, I did a small refactoring so that the try-catch is outside the InputStream read instead of inside of it. This required some small changes to DecryptHandler and associated tests.

Unit tests succeed. UploadTest in IntegTest also succeeds.


Erin-Mounts (approved):
looks good

syoung-smallwisdom (approved):
looks good

nategbrown9 (approved):
Looks good.

PR #676, opened 2023-11-01
Initial commit for redrive upload API - implementation & unit tests

chuguoxipo:


src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -527,6 +566,89 @@ public void uploadComplete(String appId, UploadCompletionClient completedBy, Upl
...
+        String currentYear = String.valueOf(DateUtils.getCurrentDateTime().getYear());
+        String currentMonth = String.valueOf(DateUtils.getCurrentDateTime().getMonthOfYear());
+        String s3Key = REDRIVE_UPLOAD_S3_KEY_PREFIX + currentYear + "-" + currentMonth;
+
+        // Upload file to S3 bucket
+        InputStream inputStream = new ByteArrayInputStream(fileBytes);
+        ObjectMetadata metadata = new ObjectMetadata();
+        metadata.setContentType(URLConnection.guessContentTypeFromStream(inputStream));
+
+        PutObjectRequest request = new PutObjectRequest(CONFIG_KEY_BACKFILL_BUCKET, s3Key, inputStream, metadata);
```
DwayneJengSage: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: When objects are uploaded to Amazon S3 by using streams (either through an AmazonS3 client or TransferManager), you might encounter network connectivity or timeout issues.

**Fix**: The most reliable way to avoid a ResetException is to provide data by using a File or FileInputStream, which the AWS SDK for Java can handle without being constrained by mark and reset limits. If the input stream isn&#39;t a FileInputStream but does support mark and reset, you can set the mark limit by using the [setReadLimit](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/RequestClientOptions.html#setReadLimit-int-) method of [RequestClientOptions.](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/RequestClientOptions.html) Its default value is 128 KB. Setting the read limit value to one byte greater than the size of stream will reliably avoid a ResetException.

[Learn more](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/best-practices.html)

DwayneJengSage: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: This code is uploading an object of type stream to S3 without setting the content length of the object. Failure to specify a content length will cause the entire contents of the input stream to be buffered locally in memory so that the content length can be calculated, which can result in negative performance problems.

**Fix**: Specify the content length for the data stream in the object metadata parameter.

[Learn more](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/PutObjectRequest.html#PutObjectRequest-java.lang.String-java.lang.String-java.io.InputStream-com.amazonaws.services.s3.model.ObjectMetadata-)

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -527,6 +568,97 @@ public void uploadComplete(String appId, UploadCompletionClient completedBy, Upl
...
+        String s3Key = REDRIVE_UPLOAD_S3_KEY_PREFIX + currentYear + "-" + currentMonth;
+
+        // Upload file to S3 bucket
+        File tempFile = File.createTempFile("temp", ".tmp");
+        try (FileOutputStream fos = new FileOutputStream(tempFile)) {
+            fos.write(fileBytes);
+        }
+        FileInputStream fis = new FileInputStream(tempFile);
+        fis.close();
+        tempFile.delete();
```
DwayneJengSage: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The result of deleting a file on this line should be checked and failure should be logged. java.io.File.delete() will simply return false if it fails to delete the file, for example, when the file does not exist. It will not throw unless there is a security exception.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -82,15 +103,18 @@
 @Component
 public class UploadService {
 
-    private static final Logger logger = LoggerFactory.getLogger(UploadService.class);
+    private static Logger logger = LoggerFactory.getLogger(UploadService.class);
```
DwayneJengSage: This should be marked as final because it's a constant and should never be changed.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -82,15 +103,18 @@
...
 
     static final long EXPIRATION = 24 * 60 * 60 * 1000; // 24 hours
     
     // package-scoped to be available in unit tests
     static final String CONFIG_KEY_UPLOAD_BUCKET = "upload.bucket";
+    static final String CONFIG_KEY_BACKFILL_BUCKET = "backfill.bucket";
```
DwayneJengSage, line 103: This bucket isn't defined anywhere. You'll need to add this config to src/main/resources/BridgeServer2.conf. The values are the same as the ones defined in BridgeWorkerPlatform.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -82,15 +103,18 @@
...
+    static final String REDRIVE_UPLOAD_S3_KEY_PREFIX = "redrive-upload-id-";
     static final String METADATA_KEY_EVENT_TIMESTAMP = "eventTimestamp";
     static final String METADATA_KEY_INSTANCE_GUID = "instanceGuid";
     static final String METADATA_KEY_STARTED_ON = "startedOn";
+    static final String WORKER_NAME_UPLOAD_REDRIVE = "uploadRedriveWorker";
```
DwayneJengSage: This is case-sensitive, and the U needs to be uppercase.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -137,6 +165,9 @@ public final void setExporter3Service(Exporter3Service exporter3Service) {
     @Autowired
     final void setConfig(BridgeConfig config) {
         uploadBucket = config.getProperty(CONFIG_KEY_UPLOAD_BUCKET);
+        redriveUploadBucket = config.getProperty(CONFIG_KEY_BACKFILL_BUCKET);
+        workerQueueUrl = config.getProperty(BridgeConstants.CONFIG_KEY_WORKER_SQS_URL);
```
DwayneJengSage: Note that SqsInitializer runs after Spring, so this won't be available when Spring calls setConfig(). You'll need to load the workerQueueUrl on the fly, just before you call sqsClient. See Exporter3Service.exportUpload() for an example.

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadController.java
```diff
@@ -167,6 +171,15 @@ public String uploadComplete(@PathVariable String uploadId,
...
     }
+
+    @PostMapping("/v3/uploads/redrive")
+    public String redriveUploads(@RequestBody byte[] fileBytes) throws IOException {
+        if (fileBytes != null && fileBytes.length != 0) {
+            uploadService.redriveUpload(fileBytes);
+            return "Redrive uploads attempted.";
+        }
+        return "Please provide a non-empty file for upload redrive.";
+    }
```
DwayneJengSage: We don't want to read raw bytes here. We should create a Java class that contains a list of upload IDs, and parse that from JSON. See for example upload() in this class.

Also, you need to call getAuthenticatedSession(Roles.SUPERADMIN) at the top to make sure only people with the correct permissions can call this API.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -527,6 +568,97 @@ public void uploadComplete(String appId, UploadCompletionClient completedBy, Upl
...
+            String appId = upload.getAppId();
+            if (appId == null) {
+                appId = healthCodeDao.getAppId(upload.getHealthCode());
+            }
+            uploadComplete(appId, UploadCompletionClient.REDRIVE, upload, true);
+            UploadValidationStatus validationStatus = pollUploadValidationStatusUntilComplete(uploadId);
+            if (validationStatus.getStatus() != UploadStatus.SUCCEEDED) {
+                logger.error("Redrive failed for uploadId=" + uploadId + ": " + COMMA_SPACE_JOINER.join(
+                        validationStatus.getMessageList()));
+            }
```
DwayneJengSage: Split this into 2 loops. First loop calls uploadComplete(). Second loop calls pollUploadValidationStatusUntilComplete(). This way, we end up completing all the uploads in parallel instead of in sequence.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
DwayneJengSage: If you use s3Helper.writeLinesToS3(), none of this is necessary. There's an example of this in Exporter3Service.initExporter3Internal().

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -805,4 +937,8 @@ protected void updateAdherenceWithUploadInfo(String appId, Upload upload) {
     private static interface UploadSupplier {
         ForwardCursorPagedResourceList<Upload> get(DateTime startTime, DateTime endTime);
     }
+
+    public void setLogger(Logger logger) {
+        this.logger = logger;
+    }
```
DwayneJengSage: We shouldn't set the logger. If you need to verify that the logger was called, make a helper method. There is a spy on svc in UploadServiceTest, so you can just verify() that that method was called. There's an example in UploadServiceTest.uploadComplete().

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
```diff
@@ -527,6 +568,97 @@ public void uploadComplete(String appId, UploadCompletionClient completedBy, Upl
...
+                        validationStatus.getMessageList()));
+            }
+        }
+    }
+
+    private void redriveLargeAmountOfUploads(byte[] fileBytes) throws IOException {
+        // set up s3 Key.
+        String currentYear = String.valueOf(DateUtils.getCurrentDateTime().getYear());
+        String currentMonth = String.valueOf(DateUtils.getCurrentDateTime().getMonthOfYear());
+        String s3Key = REDRIVE_UPLOAD_S3_KEY_PREFIX + currentYear + "-" + currentMonth;
```
DwayneJengSage: If two redrives happen at the same time for whatever reason, they will conflict with each other. It would be better to use DateUtils.getCurrentDateTime().getMillis().

If you need to test it, UploadServiceTest already uses Joda to mock the current time, so the time in the test is always TIMESTAMP.getMillis().

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/spring/controllers/UploadController.java
```diff
@@ -167,6 +171,15 @@ public String uploadComplete(@PathVariable String uploadId,
         // Upload validation status may contain the health data record. Use the filter to filter out health code.
         return HealthDataRecord.PUBLIC_RECORD_WRITER.writeValueAsString(validationStatus);
     }
+
+    @PostMapping("/v3/uploads/redrive")
+    public String redriveUploads(@RequestBody byte[] fileBytes) throws IOException {
```
DwayneJengSage: Also, this should return a StatusMessage, not a String.

src/main/java/org/sagebionetworks/bridge/services/UploadService.java
chuguoxipo: Not relevant since change of code

DwayneJengSage (approved):

DwayneJengSage:
Looks good. Let me know when unit tests and integration tests are ready for review.

DwayneJengSage (approved):

PR #685, opened 2024-08-13
Merge the 'develop' branch into the 'uat' branch.

ConsoleCatzirl:
The dev environment is healthy again after the certificate update, release the changes from develop to UAT.

* Add new RDS CA certificates to truststore
* Additional logging added by Dwayne
* Use a recent version of synapseJavaClient


.travis.yml
```diff
@@ -29,6 +29,6 @@ deploy:
     develop: org-sagebridge-bridgeserver2-deployment-devdevelop
     uat: org-sagebridge-bridgeserver2-deployment-devstaging
     prod: org-sagebridge-bridgeserver2-deployment-prod
-  zip-file: target/BridgeServer2-2.0.war
+  zip-file: target/BridgeServer2-2.1-SNAPSHOT.war
```
zaro0508, line 32: are you sure you want a snapshot?  typically we want to set to a released version

pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>BridgeServer2</artifactId>
-    <version>2.0</version>
+    <version>2.1-SNAPSHOT</version>
```
zaro0508, line 8: probably not a good idea to set to snapshot

src/main/java/org/sagebionetworks/bridge/upload/UploadValidationTask.java
```diff
@@ -88,6 +88,8 @@ public void run() {
         // Create temp dir.
         File tempDir = fileHelper.createTempDir();
         context.setTempDir(tempDir);
+        String tempDirPath = tempDir.getAbsolutePath();
+        logger.info("Created temp dir " + tempDirPath);
```
zaro0508, line 92: shouldn't this right under the line where the dir is created,  `File tempDir = fileHelper.createTempDir()`?

zaro0508 (changes requested):

src/main/java/org/sagebionetworks/bridge/upload/UploadValidationTask.java
ConsoleCatzirl: I'm not sure that matters if the only other place the variable is referenced is the try/catch block below. Also, this is a PR merged by Dwayne that was never promoted from dev to uat and I'd rather not change it.

.travis.yml
ConsoleCatzirl: This is the first time the version is being incremented, and it's not used anywhere else. I want to update itto mark our team taking ownership, but it doesn't matter for the deployment. this travis job will upload the zip file with a hash name (e.g. `travis-f594a6bd734f571c395616256187ab3c1d91c56a-1723507589.zip`) and the version that shows in elastic beanstalk matches the hash (e.g. version `travis-f594a6bd734f571c395616256187ab3c1d91c56a-1723507595`)

zaro0508 (approved):

PR #686, opened 2024-08-27
[IT-3777] Add a WAF to the ALB

ConsoleCatzirl:
Extend the Application Load Balancer created by Beanstalk with a Web Application Firewall similar to this post [1], modified to use baseline rulesets managed by AWS [2].

[1] https://repost.aws/knowledge-center/elastic-beanstalk-host-attacks
[2] https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-baseline.html

### DIFF
--- a/ebextensions/waf.config
+++ b/ebextensions/waf.config
@@ -1,0 +1,58 @@
+Resources:
+  WafAcl:
+    Type: 'AWS::WAFv2::WebACL'
+    Properties:
+      Name: WafAcl
+      Scope: REGIONAL
+      DefaultAction:
+        Allow: {}
+      VisibilityConfig:
+        SampledRequestsEnabled: true
+        CloudWatchMetricsEnabled: true
+        MetricName: WafAclMetric
+      Rules:
+        - Name: AwsCommonRule
+          Priority: 0
+          OverrideAction:
+            None: {}
+          VisibilityConfig:
+            SampledRequestsEnabled: true
+            CloudWatchMetricsEnabled: true
+            MetricName: AwsCommonRuleMetric
+          Statement:
+            ManagedRuleGroupStatement:
+              VendorName: AWS
+              Name: AWSManagedRulesCommonRuleSet
+        - Name: AwsKnownBadRule
+          Priority: 1
+          OverrideAction:
+            None: {}
+          VisibilityConfig:
+            SampledRequestsEnabled: true
+            CloudWatchMetricsEnabled: true
+            MetricName: AwsKnownBadRuleMetric
+          Statement:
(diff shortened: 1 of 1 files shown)


zaro0508:
IT-377 doesn't seem like the right issue for this PR

ConsoleCatzirl:
You're right, that should be IT-3777

ebextensions/waf.config
```diff
@@ -0,0 +1,55 @@
...
+            MetricName: AwsAdminRuleMetric
+          Statement:
+            ManagedRuleGroupStatement:
+              VendorName: AWS
+              Name: AWSManagedRulesAdminProtectionRuleSet
+
+  WafAclAssociation:
+    Type: 'AWS::WAFv2::WebACLAssociation'
+    Properties:
+      ResourceArn: !Ref AWSEBV2LoadBalancer
```
zaro0508, line 57: i don't think this is gonna work since `AWSEBV2LoadBalancer` is not in here.  

you might have better luck if you move this template to the [BridgeServer2-infra repo](https://github.com/Sage-Bionetworks/BridgeServer2-infra).  you'll probably want to setup something like [bridgeserver2-https-redirect.yaml](https://github.com/Sage-Bionetworks/BridgeServer2-infra/blob/develop/config/develop/bridgeserver2-https-redirect.yaml)

zaro0508 (changes requested):

ebextensions/waf.config
ConsoleCatzirl: Beanstalk let's us [reference the resources created by beanstalk for the environment](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/customize-containers-format-resources-eb.html) in these templates. since our beanstalk environment is web-server tier with load-balancing, beanstalk will create `AWSEBV2LoadBalancer` for us.

zaro0508: ohh cool, wondering if this is something new or whether it has always been there and i just didn't know about it.  anyways, wish i knew about this earlier.

Do you know whether it's also possible to magically reference `AWSEBV2LoadBalancer` like this in a cloudformation template when deployed from BridgeServer2-infra repo?

zaro0508 (approved):

ebextensions/waf.config
ConsoleCatzirl: No, I don't think you can directly reference `AWSEBV2LoadBalancer` from another template, I think beanstalk might be magically combining this with a cloudformation template it makes behind the scenes. 

but it is possible to export a stack output and reference that from the stack in BridgeServer2-infra.

PR #701, opened 2026-07-15
[IT-5233] Publish catalina.out to S3 and read both Tomcat log paths

ConsoleCatzirl:
After the AL2 -> AL2023 upgrade the workers/exporter stopped logging to Sumo Logic. Two things combined: AL2023 runs Tomcat 9 (native log dir /var/log/tomcat9), and the platform's default S3 log-publication set no longer includes catalina.out, so the application log (written by our rsyslog redirect to /var/log/tomcat) stopped reaching the Sumo Logic collectors, which ingest from the EB S3 publish bucket. CloudWatch was unaffected because it read the file directly.

Approach: read catalina.out from both locations, so logs are captured no matter which location holds them, without duplicating entries.

- Add ebextensions/publishlogs.config to explicitly publish catalina.out (and rotations) from both paths to S3, restoring the Sumo Logic feed.
- tomcatlogs.config: rsyslog keeps writing the single canonical /var/log/tomcat/catalina.out; the CloudWatch agent reads both /var/log/tomcat and /var/log/tomcat9, and BOTH stream into the same version-less /var/log/tomcat log group (the reusable group the *-infra stack provisions and attaches its metric-filter alarms to).
- logrotate-tomcat.config: rotate both paths.

The version-less /var/log/tomcat is intentionally the canonical path (*-infra dropped the version "for future reusability"); /var/log/tomcat9 is read as a fallback for the platform's native Tomcat 9 dir.

### DIFF
--- a/ebextensions/logrotate-tomcat.config
+++ b/ebextensions/logrotate-tomcat.config
@@ -4,6 +4,7 @@ files:
     owner: root
     group: root
     content: |
+      # Rotates the Tomcat catalina.out written by our rsyslog redirect (see tomcatlogs.config).
       /var/log/tomcat/catalina.out {
         daily
         rotate 7
--- a/ebextensions/publishlogs.config
+++ b/ebextensions/publishlogs.config
@@ -1,0 +1,22 @@
+###################################################################################################
+#### Adds Tomcat's catalina.out to Elastic Beanstalk's "publish logs to S3" task.
+####
+#### EB rotates a platform-defined set of logs to the environment's S3 bucket under
+#### resources/environments/logs/publish/<env-id>/<instance-id>/, and the Sumo Logic collectors
+#### ingest from that bucket. The AL2023 Tomcat 9 platform's default publish set does NOT include
+#### the application's catalina.out (only Tomcat's JULI catalina.<date>.log files), so application
+#### logs stopped reaching Sumo Logic after the AL2 -> AL2023 upgrade. This restores them by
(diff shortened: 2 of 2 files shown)


ebextensions/logrotate-tomcat.config
```diff
@@ -4,7 +4,13 @@ files:
...
     group: root
     content: |
-      /var/log/tomcat/catalina.out {
+      # Rotates the Tomcat catalina.out written by our rsyslog redirect (see tomcatlogs.config).
+      #
+      # rsyslog writes to /var/log/tomcat (canonical; matches the reusable CloudWatch log group).
+      # /var/log/tomcat9 is the platform's native Tomcat 9 dir, listed so rotation still applies
+      # if the log ever lands there instead. `missingok` tolerates whichever path is absent.
+      /var/log/tomcat/catalina.out
+      /var/log/tomcat9/catalina.out {
```
zaro0508: does this need to be updated when tomcat is updated to a different version?  if so is there an approach that would make it so we don't need to update anymore?

ConsoleCatzirl: I'm trying to future-proof, but yeah, we'll need to update this if tomcat changes the directory it writes the logs to. I try to get everything in `/var/log/tomcat` so that we don't need to update anything, but the transition from AL2 -> AL2023 caused tomcat to start writing to `/var/log/tomcat9`, so this change just updates everything to support reading from both locations.

zaro0508 (approved):

